PR #58, opened 2020-06-23
Merging linear algebra and differential equations code

JonathanWenger:


JonathanWenger:
We need to remove the code on noisy linear solvers from this pull request as it is not yet working and tests are not running through. This can be readded via pull request as a later date.

src/probnum/linalg/linearsolvers/solutionbased.py
```diff
@@ -26,9 +29,49 @@ class SolutionBasedSolver(ProbabilisticLinearSolver):
...
+        if iter >= maxiter:
+            warnings.warn(message="Iteration terminated. Solver reached the maximum number of iterations.")
+            return True, "maxiter"
+        # residual below error tolerance
+        elif np.linalg.norm(resid) <= atol:
+            return True, "resid_atol"
+        elif np.linalg.norm(resid) <= rtol * np.linalg.norm(self.b):
+            return True, "resid_rtol"
+        # uncertainty-based
+        # todo: based on posterior contraction
```
pnkraemer, line 72: is this included as a "todo" on purpose?

src/probnum/prob/distributions/distribution.py
```diff
@@ -72,21 +74,53 @@ def __init__(self, parameters=None, pdf=None, logpdf=None, cdf=None, logcdf=None
...
+        except NotImplementedError:
+            # Set shape based on a sample
+            if np.isscalar(self.sample(size=1)):
+                shape_sample = ()
+            else:
+                shape_sample = self.sample(size=1).shape
+            if shape is None or shape_sample == shape:
+                self._shape = shape_sample
+            else:
+                raise ValueError("Shape of distribution mean and given shape do not match.")
```
pnkraemer, line 109: shape of distribution sample?

src/probnum/prob/distributions/distribution.py
```diff
@@ -72,21 +74,53 @@ def __init__(self, parameters=None, pdf=None, logpdf=None, cdf=None, logcdf=None
...
+        try:
+            # Set shape based on mean
+            if np.isscalar(self.mean()):
+                shape_mean = ()
+            else:
+                shape_mean = self.mean().shape
+            if shape is None or shape_mean == shape:
+                self._shape = shape_mean
+            else:
+                raise ValueError("Shape of distribution mean and given shape do not match.")
```
pnkraemer, line 99: whenever I see code blocks with a leading comment, I personally think it would be more readable to extract this code into a function, i.e. `self. _set_shape_based_on_mean(...)` and so on which makes each function nice and short (and the code self-documenting). I understand that this is just personal preference.

src/probnum/prob/distributions/distribution.py
```diff
@@ -229,7 +263,7 @@ def sample(self, size=()):
             return self._sample(size=size)
         else:
             raise NotImplementedError(
-                'The function \'sample\' is not implemented for object of class {}'.format(type(self).__name__))
+                'The function \'sample\' is not implemented for object of class {}.'.format(type(self).__name__))
```
pnkraemer, line 266: maybe extract these kinds of error into an `errormsg` variable? That makes the line automatically shorter and more readable.

src/probnum/prob/distributions/normal.py
```diff
@@ -398,7 +402,7 @@ def sample(self, size=()):
                                                    cov=self.cov(), size=size,
                                                    random_state=self.random_state)
 
-    def reshape(self, shape):
+    def reshape(self, newshape):
         raise NotImplementedError
 
     # Arithmetic Operations ###############################
```
pnkraemer, line 408: block comments seem inconsistent between different code pieces? Shall we use `# bla #########` or just `# bla`

src/probnum/prob/distributions/normal.py
```diff
@@ -468,7 +472,7 @@ def pdf(self, x):
                                                            mean=self.mean().ravel(),
                                                            cov=self.cov())
         # TODO: this reshape is incorrect, write test for multiple matrices
```
pnkraemer, line 474: what about this?

tests/test_linalg/test_linops/test_linearoperators.py
```diff
@@ -81,6 +81,15 @@ def test_matvec(self):
...
                                     msg="Matrix-vector multiplication with (n,1) vector failed.")
 
+    def test_broadcasting(self):
+        """Test broadcasting rules for linear operators and compare to np.array broadcasting."""
+        # for alpha, A in list(itertools.product(self.scalars, self.ops)):
+        #     with self.subTest():
+        #         self.assertAllClose((alpha * A).todense(), alpha * A.todense())
+        #         self.assertAllClose((alpha + A).todense(), alpha + A.todense())
+        #         self.assertAllClose((A - alpha).todense(), A.todense() - alpha)
+        pass
```
pnkraemer, line 91: is this a todo or outdated?

src/probnum/prob/distributions/normal.py
JonathanWenger: I vote for just `# ` or a standard based on n`#`s left and right of the comment.

pnkraemer: Yeah what happened in ` # Arithmetic Operations ############################### ` (and the respective other places) is one `#` on the left and as many on the right as necessary to get to 79 lines (to serve as visual separation between blocks of methods but still not longer than the default column width). I don't think this decision is any kind of bottleneck though so I don't really care too much.

pnkraemer:
Github shows a conflict that needs resolving? Other than that everything is approved :)

JonathanWenger:
> Github shows a conflict that needs resolving? Other than that everything is approved :)

Still working on cleaning this up. Should be done soon!

pnkraemer:
No pressure, I was just wondering.

src/probnum/prob/distributions/distribution.py
JonathanWenger: Following `scipy` convention here which is: Custom error type for often repeated errors of a specific form, otherwise (potentially repeated) standard error messages.

src/probnum/prob/distributions/normal.py
JonathanWenger: This will be eventually resolved by work from Aljoscha. Removed Todos and created an issue instead.

JonathanWenger:
Test coverage goes down since the noisy linear solver is only a mockup so far and the draft state has been moved to a local fork.

PR #59, opened 2020-06-23
Adding differential equations code to the main package

JonathanWenger:


pnkraemer:
On the diffeq/filtsmooth side, all todos seem to be done.

JonathanWenger:
This should be ready to merge now after the linalg updates are added.

PR #87, opened 2020-07-14
Joint efforts with Nathanael

pnkraemer:
Simplified filtering and smoothing, adapted ODE solver correspondingly plus some minor changes like docstrings or variable renaming.


pnkraemer:
Will fix it now. 
Stay tuned.

pnkraemer:
There we go. Let's see what happens now.

PR #93, opened 2020-07-15
Remove `probnum.optim` and `probnum.prob.sampling` completely

nathanaelbosch:
Completely remove these parts: Package code, tests, and references in the documentation.

The removed files can still be recovered in the future:
- It can be recovered through the git history of this repository
- I have a branch in my personal fork of ProbNum that is set to the state before this removal: https://github.com/nathanaelbosch/probnum/tree/optim%2Bsampling
- @pnkraemer also has a local copy of this code


nathanaelbosch:
"Draft" because it might very well be that Travis fails since I did not run tests and build the documentation locally.

nathanaelbosch:
Not really sure what it means to have "codecov/project" fail. Maybe just because the percentual coverage got smaller?

pnkraemer:
Are the inits and the docs adapted? Maybe this is why it fails...

nathanaelbosch:
The "fail" is just in codcov/project, not in travis-ci, so there is no actual test failing (otherwise Travis-CI would complain)

pnkraemer:
Weird. I am tempted to merge anyway but still hesitant... what shall we do?

nathanaelbosch:
Personally I would also be tempted to just merge. I strongly believe it is just about having a negative delta in code-coverage. I did not find any decisive information in the codecov docs however.

pnkraemer:
yeah it was probably because the samplers especially were excessively tested and now the overall coverage is down. If in the next days there is an unforeseen problem we know the problem was here.

Issue #60, opened 2020-06-23
Jupyter notebooks illustrating ODE solvers


JonathanWenger:
The examples section of the documentation should contain jupyter notebooks illustrating the usage of `probnum.diffeq`.

pnkraemer:
@nathanaelbosch something like the plot in your presentation yesterday would be awesome. Can you do this? Further, I/we can do some convergence rate plots like in the papers (https://arxiv.org/pdf/1807.09737.pdf and https://arxiv.org/abs/2004.00623). Some scaffolding is already there, but I am not sure whether this refers to up-to-date versions...

PR #106, opened 2020-07-17
Restore diffeq notebooks

nathanaelbosch:
Addresses  #60 in a minimal way: I just fixed the notebooks that @pnkraemer already wrote!

This way we have some examples for the ODE solver on the homepage :)


JonathanWenger:
Thanks for the quick fix. This will be a good temporary addition. I suggest to move the legends in the `adaptive_steps_odefilter.ipynb` to one of the corners instead of the center. Once that's done this can be merged.

pnkraemer:
Great! I was actually doing the same in the process of writing the stylefile thing, but I guess I was too slow with a PR ;)

PR #111, opened 2020-07-20
Use flake8 for better code-style enforcement

### LINKED ISSUE #108: Use flake8 to better verify PEP 8 compliance
This is complementary to using *Black* (#102). There is also overlap with #66. 

### About flake8
[flake8](https://flake8.pycqa.org/en/latest/index.html) checks code compliance to [PEP 8](https://www.python.org/dev/peps/pep-0008/). This is not a duplication to our usage of *Black*: *Black* formats code automatically, but there are code issues that *Black* can not automatically solve, e.g. [wildcard imports](https://docs.quantifiedcode.com/python-anti-patterns/maintainability/from_module_import_all_used.html). flake8 captures these issues.

Sidenote: Trying out flake8 locally seems to raise mostly the following issues:
- line-length (~75%)
- star imports (~10%)
- unused imports (~10%)

There is also a flake8 plugin which I encountered together with flake8 (in the [*Black* repo](https://github.com/psf/black/blob/master/.flake8) actually): [flake8-bugbear](https://github.com/PyCQA/flake8-bugbear). The provided [list of warnings](https://github.com/PyCQA/flake8-bugbear#list-of-warnings) gives me the impression that it overall provides useful feedback.

#### My personal opinion
Cons:
- Yet another tool
- Additional set up requirements for developers
- Another larger code migration

Pros:
- Better code quality through better PEP 8 compliance
- Less work for third-party / beginner PR reviews since feedback on these code quality aspects is given automatically

### Concrete Proposal
- using both flake8 and flake8-bugbear
- adding flake8 to tox, and thus to TravisCI
  (alternatively we could probably also use GitHub Actions)
- adding a pre-commit hook to support that workflow, similar to what I did for `black`

If we chose to do this, then again we will need some form of code *migration* through a larger commit.
I believe that this should best be done right after merging (#102).

nathanaelbosch:
Closes #108. The linked issue provides more in-depth explanation.


nathanaelbosch:
Alright! The current state in this PR:
- Check for flake8 compliance through GitHub Actions
  You can see how it looks above. We get separate feedback for Black, flake8 (through GH-actions) and testing / documentation (through Travis)
- I fixed many flake8 issues, but most of them quite mechanically. What I mean by this is that I did e.g. not remove TODOs or comments, even though we want them to be gone. I did not re-write docstrings when they did not use the correct grammar. I think that it is far less painful to improve such parts as you work with the code. Personally, I will try pay extra attention to that for any future work on the diffeq parts.
- I excluded some errors for now:
  - C901: Code complexity
    This affects  'SymmetricMatrixBasedSolver._get_prior_params' and 'NoisySymmetricMatrixBasedSolver._get_prior_params' . I'm not familiar with that code (I guess @JonathanWenger is). I'll let you judge if that is something you want to improve, or if we should just increase the code complexity setting to something >21.
  - Line-length in `./tests`
    Resolving these issues should be possible. But I have the impression that `unittests` adds quite some additional indentation, so it might also be an option to just allow for longer lines in `./tests`. I'm not sure about this.
  - Again in `./tests`: F841 "Local variable assigned but never used"
    This is not a big deal. I just did not want to delete the respective lines, since they might actually be there for a reason (ie. computing a variable is somewhat already a test; it could fail and raise an error). Should be easy to fix, or we just keep this exception.

JonathanWenger:
>     * I excluded some errors for now:
>       
>       * C901: Code complexity
>         This affects  'SymmetricMatrixBasedSolver._get_prior_params' and 'NoisySymmetricMatrixBasedSolver._get_prior_params' . I'm not familiar with that code (I guess @JonathanWenger is). I'll let you judge if that is something you want to improve, or if we should just increase the code complexity setting to something >21.

This should definitely improve (there is also some code duplication). So do not increase the complexity setting because of this.

nathanaelbosch:
We should also remove star imports from `__init__.py` files.
Example: We currently have `probnum.prob.scipy` available

JonathanWenger:
@nathanaelbosch can you change the target branch to `master`? Deletion of `development` (#129 ) is blocked by this PR.

PR #113, opened 2020-07-22
Quick tox improvements

nathanaelbosch:
Following our lunch discussion.

Instead of specifying `py36` for the default `tox` run, I relaxed it to `py3`. This way, people that do not have Python 3.6 available locally (because they use another Python 3.x version) do not run into a tox error, but `tox` just uses the `python3` that is available. The drawback is that the local *default* test run and the run on Travis now differ a bit more, but I don't think that this should lead to significant errors. I would suspect that there would be more issues because people don't know how to work with tox, than because of locally passing tests that fail on Travis-CI. We could also further reduce the chances for the latter issues by adding more Python versions to Travis, e.g. adding 3.7 and 3.8.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py36,docs
+envlist = py3,docs
 
 [travis]
 python =
@@ -35,11 +35,12 @@ commands =
 
 [testenv:black]
 description = Code linting with Black
+basepython = python3
 deps = black
 commands = black --check --diff .
 
 [testenv:format]
 description = Code formatting with Black (and possibly other tools in the future)
+basepython = python3
 deps = black
 commands = black .
-


tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py36,docs
+envlist = py3,docs
 
 [travis]
 python =
```
JonathanWenger, line 10: I am in favor of adding Python 3.7 and 3.8 here to reduce the possibility of differences between a local run of tox and a Travis build.

nathanaelbosch: I added testing for 3.7 and 3.8 to Travis. There could still be differences between local runs and travis builds: If 3.6 passes, which I use locally, but 3.7 fails, then I only realize that after building with Travis. 

But before the change we could have encountered problems the other way round: Local tests with 3.7 fail, while 3.6 passes on Travis. This way at least we would realize when we don't support a Python version.

I guess there will always be a compromise on some level, between "thorough local testing that mimics Travis" and "easy simple local testing via `tox`".

Issue #76, opened 2020-07-03
Register the package with the python package index PyPi


JonathanWenger:
This allows the package to be installed with `pip` and reserves the name `probnum`. Ideally we would want to set up a Github action (or similar) which whenever a new version is released here on github automatically updates the one on PyPi. A tutorial on how to do this can be found here:

- https://packaging.python.org/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

JonathanWenger:
There is also already a predefined workflow via github actions available: 

https://github.com/probabilistic-numerics/probnum/new/master?filename=.github%2Fworkflows%2Fpython-publish.yml&workflow_template=python-publish

JonathanWenger:
Closed by #121 and #118

PR #118, opened 2020-07-23
Publish package on pypi

JonathanWenger:
This PR enables automatic upload to PyPI by setting the correct format of the long description. Upload to PyPI was tested on testPyPI and can be viewed under this link: https://test.pypi.org/project/probnum/0.0.1/

Once this PR is merged into master any release created in GitHub will trigger an upload to PyPI via a github action.


JonathanWenger:
This resolves #76.

README.md
```diff
@@ -14,7 +14,7 @@ measures over the output of a numerical routine, which then can be propagated al
 This repository aims to implement methods from PN in Python 3 and to provide a common interface for them. This is
 currently a work in progress, therefore interfaces are subject to change.
 
-## Installation and Documentation
+## Installation
```
nathanaelbosch, line 17: If this PR enables people to install ProbNum through `pip install probnum`, we should add such a line to the README (and possibly the documentation).

README.md
```diff
@@ -24,13 +24,10 @@ Alternatively you can clone this repository with
 git clone https://github.com/probabilistic-numerics/probnum
```
nathanaelbosch, line 24: Not this line, but the one above between both installation.
I would change it to something like 
"Alternatively, you can install this package from source with"

PR #120, opened 2020-07-23
Opinionated refactoring: Changed/removed docstrings, functions, variable names, with the goal on making the code more explicit

nathanaelbosch:
Open TODO: Write a `_get_steprule` function to get that complexity into it's separate function.

### DIFF
--- a/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
+++ b/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
@@ -60,22 +60,22 @@ class GaussianIVPFilter(odesolver.ODESolver):
         times, means, covars = [self.ivp.t0], [current_rv.mean()], [current_rv.cov()]
         while times[-1] < self.ivp.tmax:
             intermediate_step = float(step / nsteps)
-            tm = times[-1]
+            t = times[-1]
 
             pred_rv = current_rv
             interms, intercs, interts = [], [], []
             for idx in range(nsteps):
-                newtm = tm + intermediate_step
-                pred_rv, __ = self.gfilt.predict(tm, newtm, pred_rv, **kwargs)
+                t_new = t + intermediate_step
+                pred_rv, __ = self.gfilt.predict(t, t_new, pred_rv, **kwargs)
                 interms.append(pred_rv.mean().copy())
                 intercs.append(pred_rv.cov().copy())
-                interts.append(newtm)
-                tm = newtm
+                interts.append(t_new)
+                t = t_new
 
-            new_time = tm
+            t_new = t
             zero_data = 0.0
             filt_rv, covest, ccest, mnest = self.gfilt.update(
-                new_time, pred_rv, zero_data, **kwargs
+                t_new, pred_rv, zero_data, **kwargs
             )
 
             interms[-1] = filt_rv.mean().copy()
--- a/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
+++ b/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
@@ -87,8 +87,13 @@ class GaussianIVPFilter(odesolver.ODESolver):
             else:
                 current = RandomVariable(distribution=Normal(means[-1], covars[-1]))
             step = self._suggest_step(step, errorest)
+
         means, covars = self.undo_preconditioning(means, covars)
-        return np.array(means), ssqest * np.array(covars), np.array(times)
+
+        means, covars, times = np.array(means), np.array(covars), np.array(times)
+        covars *= ssqest
+
+        return means, covars, times
 
     def odesmooth(self, means, covs, times, **kwargs):
         """
--- a/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
+++ b/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
@@ -61,6 +61,7 @@ class GaussianIVPFilter(odesolver.ODESolver):
         while times[-1] < self.ivp.tmax:
             intermediate_step = float(step / nsteps)
             tm = times[-1]
+
             interms, intercs, interts = [], [], []
             for idx in range(nsteps):
                 newtm = tm + intermediate_step
@@ -70,6 +71,7 @@ class GaussianIVPFilter(odesolver.ODESolver):
                 interts.append(newtm)
                 tm = newtm
             predicted = current
+
             new_time = tm
             zero_data = 0.0
             current, covest, ccest, mnest = self.gfilt.update(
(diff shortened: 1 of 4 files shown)


src/probnum/diffeq/steprule.py
```diff
@@ -1,28 +1,18 @@
...
-        e_n at step h_n.
-        """
+        """Suggest a new step h_{n+1} given error estimate e_n at step h_n."""
         raise NotImplementedError
 
     @abstractmethod
     def is_accepted(self, proposedstep, errorest, **kwargs):
-        """
-        Checks if proposed step should be accepted or not.
+        """Check if the proposed step should be accepted or not.
```
pnkraemer, line 14: Shouldn't it be either `"""Check if ..."""` or `"""\nCheck if ...\n"""`?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -248,16 +248,16 @@ def _create_solver_object(
...
     _prior = _string2prior(ivp, which_prior, precond_step, **kwargs)
     if tol is not None:
-        stprl = _step2steprule_adap(tol, _prior)
+        stprl = steprule.AdaptiveSteps(tol, _prior.ordint + 1, **kwargs)
         if firststep is None:
             firststep = ivp.tmax - ivp.t0
     else:
-        stprl = _step2steprule_const(step)
+        stprl = steprule.ConstantSteps(step)
         firststep = step
```
pnkraemer, line 261: I remember you saying that you don't see an easy way of extracting this into a function. Why again was that the case? Because there would be too many inputs for this function?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -56,45 +54,57 @@ def solve(self, firststep, nsteps=1, **kwargs):
...
-            predicted = current
-            new_time = tm
+                t_new = t + intermediate_step
+                pred_rv, __ = self.gfilt.predict(t, t_new, pred_rv, **kwargs)
+                interms.append(pred_rv.mean().copy())
+                intercs.append(pred_rv.cov().copy())
+                interts.append(t_new)
+                t = t_new
+
+            t_new = t
```
pnkraemer, line 75: this is just intermediate, no? If I understand correctly, this block will be gone with the next PR anyway?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -56,45 +54,57 @@ def solve(self, firststep, nsteps=1, **kwargs):
 
         ####### This function surely can use some code cleanup. #######
 
-        current = self.gfilt.initialdistribution
+        current_rv = self.gfilt.initialrandomvariable
         step = firststep
-        ssqest, ct = 0.0, 0
-        times, means, covars = [self.ivp.t0], [current.mean()], [current.cov()]
+        ssqest, n_steps = 0.0, 0
```
pnkraemer, line 59: How strong is your opinion on 'n_steps'? ;) I agree that it is better than `ct` (count), maybe something even better would be `num_steps`? Not a big issue though :)

src/probnum/diffeq/steprule.py
nathanaelbosch: Does not have to: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings "The summary line may be on the same line as the opening quotes or on the next line."

But I'm fine with changing that of course :)

src/probnum/diffeq/steprule.py
```diff
@@ -75,16 +60,19 @@ class AdaptiveSteps(StepRule):
...
+        **kwargs
     ):
         self.tol_per_step = float(tol_per_step)
         self.safetyscale = float(safetyscale)
         self.localconvrate = float(localconvrate + 1)
         self.limitchange = limitchange
 
     def suggest(self, laststep, errorest, **kwargs):
-        """
-        """
```
pnkraemer, line 87: While you're at it, would it be okay if you put a single sentence into this docstring instead of deleting the empty one? This is not criticism (!) but it would be cool to have something here that describes what ''suggest'' does :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Yep, this is gone with the dense-output stuff. But I still use `t` and `t_new`, and `pred_rv` as variable names there.

src/probnum/diffeq/steprule.py
pnkraemer: I agree! My intention behind this comment was rather that the rest of the repository does it in the style I proposed here, so it was more of a consistency thing instead of a right/wrong thing :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: Fair enough. I don't see any problem here (for now), was just wondering.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Not strong! I can definitely change that to `num_steps`. I'd also be fine with `count`, `n`, `n_iter`, or some other proposals.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
nathanaelbosch: The function would take `step`, `_prior` (or at least `localconvrate = _prior.ordint + 1`), `tol`, `firststep`, and something related to `ivp.tmax` and `ivp.tmin`, and it would output `steprule, firststep`.

I have the impression that that's a lot of function arguments, and it grows with any new steprule. I think it would be worth it to improve that aspect, but it seems like more than just moving a few lines around, so I thought I't leave it at the old behavior for now.

If you have a clear idea on how to improve that code let me know! :)

src/probnum/diffeq/steprule.py
nathanaelbosch: `suggest` of the abstract base class is already documented and states the general purpose of the `suggest` function. Do you mean to describe more precisely what it does for `AdaptiveSteps`?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
pnkraemer: Okay that is clearly something very subjective and I have no problems with the current solution. I, personally wouldn't mind a function ```steprule, firststep = _step2steprule(step, tol, firststep, _prior, ivp)``` but I understand why you find that odd :)

src/probnum/diffeq/steprule.py
pnkraemer: wasn't aware that it is already documented in the superclass. In this case please ignore my comment.

PR #123, opened 2020-07-27
PN colorscheme in docs/tutorials

pnkraemer:
Using `plt.style.use("probnum.mplstyle")` one can get the following plotting style (compared to the default matplotlib style):
[styledefault.pdf](https://github.com/probabilistic-numerics/probnum/files/4980681/styledefault.pdf)
[styleprobnum.mplstyle.pdf](https://github.com/probabilistic-numerics/probnum/files/4980682/styleprobnum.mplstyle.pdf)
[first_order_linearisation.pdf](https://github.com/probabilistic-numerics/probnum/files/4980709/first_order_linearisation.pdf)

Essentially, anything that was defined through `plt.rcParams` before is now contained in this file.
The major changes compared to the default are: figure size, top and right axes removed, pn-orange-green as default colors, fileformat pdf plus some linewidth changes. 

Difference to the previous rcParams: `usetex` is set to `False` (it takes ages to render, the axes ticks look super weird and basic sub/superscripts as well as blackboard bold work without)

Not done yet: adapt all notebooks, so far only the ODE related tutorials use this file.


JonathanWenger:
Personally, I don't agree with not using `tex`. We may want to use LaTeX packages for more involved plot annotations in the future. Some linear algebra notebooks currently use `bm` for example. I haven't had any issues with longer render times and I prefer the LaTeX font.

pnkraemer:
Haha there we go, formatting plots is just too subjective to have a single "correct" way of doing it :D 
I have already seen that there is some maths-boldfacing in the linear algebra notebooks and this is precisely why I haven't touched those yet.

I think there are three potential solutions which would each avoid unnecessarily long arguments about matplotlib-fonts (which I would like to avoid at any costs):
1) We ignore this PR, use the matplotlib default and each of us does what he likes in the notebooks.
2) We use the proposed style and accept that the current style is **just** a default and wherever fancy labelling, etc. are required this can obviously be used (there will probably never be a one size fits all solution). I guess that would mean that the linear algebra notebooks potentially have different fonts (latex) than the ODE notebooks but so be it.
2) We see what the other people think (@nathanaelbosch?). If I am the only one that doesn't like latex fonts I will happily include a `usetex=True` in the style file.

nathanaelbosch:
I like how the provided .pdf files look so I'm fine with the defaults you propose. The default colors only contain orange and green, right? How well does it work when plotting a third function?

I don't have strong feelings about using tex. There are definitely cases in which we want to use it, but I personally don't care if it's a default setting or not. I don't think we can get rid of all plot customization forever, there will always be something that you might want to adjust in order to make a plot look better. So I guess I'm not really helpful to resolve that argument :D

JonathanWenger:
Let's discuss in the PN Software Interest Group meeting today. I think we should definitely have your style file in the repo either with `usetex=True` or `False`. It's really nice and looks much cleaner to have the same style across notebooks!

JonathanWenger:
Decision from PN Software Interest Group: `usetex=True` should be the default in the style file.

pnkraemer:
Update: `usetex=True` (if you're wondering what maths looks like: see the integral stuff in the title of:
[first_order_linearisation.pdf](https://github.com/probabilistic-numerics/probnum/files/4989882/first_order_linearisation.pdf)
)
the notebooks are updated accordingly (in the preamble---@JonathanWenger I did not know which titles you needed in sans-serif and which ones in serif, but sans serif is default and serif can perhaps be done by something like `\\textit{\\textrm{A}}^{-1}`? The rest essentially stayed the same.

JonathanWenger:
@pnkraemer can you change the target branch to `master`? Deletion of `development` (#129 ) is blocked by this PR.

pnkraemer:
The merge conflict was too annoying to fix it in place. I close this PR in favour of #177.

PR #133, opened 2020-08-03
Vectorise ah qh

pnkraemer:
For the IBM, the computation of A(h) and Q(h) is now fully vectorised. The corresponding unittests go down from 8s to roughly 5-6s (which is the only performance tests I have tried). Notably, for A(h) looping over the diagonal elements (diagonal->first offdiagonal->second offdiagonal etc.) is faster than a fully vectorised, numpy based implementation (5.5s vs 7s). I suppose this is because scipy.special.factorial---albeit being able to handle matrices as inputs---is much slower than np.math.factorial (see https://stackoverflow.com/questions/21753841/factorial-in-numpy-and-scipy).


pnkraemer:
Oh, another difference is that I put the formulas for A and Q into the IBM-chapmankolmogorov docstring.

Issue #130, opened 2020-07-28
Only use np.float internally


JonathanWenger:
Set a standard (maybe in styleguide) to only use numpy floats internally. This allows numpy operations (like slicing) etc., which then return `np.arrays`. See also https://numpy.org/doc/stable/user/basics.types.html for the difference between python floats and numpy floats.

pnkraemer:
Just out of curiousity: Which direction are you thinking about here? 
```python
if np.isscalar(scalar) and scalar.dtype not in [np.float32, np.float64, np.integer]:
    scalar = np.asarray(scalar)
```
or
```python
if np.isscalar(scalar) and scalar.dtype not in [np.float32, np.float64, np.integer]:
    raise TypeError
```

marvinpfoertner:
> Just out of curiousity: Which direction are you thinking about here?
> 
> ```python
> if np.isscalar(scalar) and scalar.dtype not in [np.float32, np.float64, np.integer]:
>     scalar = np.asarray(scalar)
> ```
> 
> or
> 
> ```python
> if np.isscalar(scalar) and scalar.dtype not in [np.float32, np.float64, np.integer]:
>     raise TypeError
> ```

I would say the first one

marvinpfoertner:
> Just out of curiousity: Which direction are you thinking about here?
> 
> ```python
> if np.isscalar(scalar) and scalar.dtype not in [np.float32, np.float64, np.integer]:
>     scalar = np.asarray(scalar)
> ```
> 
> or
> 
> ```python
> if np.isscalar(scalar) and scalar.dtype not in [np.float32, np.float64, np.integer]:
>     raise TypeError
> ```

I implemented a proposal for this code in my PR #135. What do you guys think? @JonathanWenger @pnkraemer @nathanaelbosch

marvinpfoertner:
I just learned that numpy's array scalars don't always behave exactly like arrays. See also https://numpy.org/doc/stable/reference/arrays.scalars.html.
IMHO we would avoid boilerplate code if scalar quantities behaved exactly the same as arrays.
Maybe we should use arrays of shape () internally?

PR #134, opened 2020-08-11
(Advanced) Indexing, Slicing, Masking, Reshaping and Transposition for Random Variables and Distributions

### LINKED ISSUE #11: Implement indexing and reshaping for random variables
Indexing, masking and reshaping as is possible for ``numpy`` arrays should be possible for random variables in a similar manner.

marvinpfoertner:
This PR enables full numpy-style indexing for random variables and distributions (currently Normal and Dirac) as in

https://numpy.org/doc/stable/reference/arrays.indexing.html

Closes #11.

### DIFF
--- a/src/probnum/prob/distributions/normal.py
+++ b/src/probnum/prob/distributions/normal.py
@@ -140,26 +140,35 @@ class Normal(Distribution):
         # Select submatrix from covariance matrix
         cov = self.cov().reshape(self.mean().shape + self.mean().shape)
         cov = cov[key][tuple([slice(None)] * mean.ndim) + key]
-        cov = cov.reshape(mean.size, mean.size)
+
+        if cov.ndim > 2:
+            cov = cov.reshape(mean.size, mean.size)
 
         return Normal(mean=mean, cov=cov, random_state=self.random_state)
 
     def reshape(self, newshape):
         try:
-            return Normal(
-                mean=self.mean().reshape(newshape),
-                cov=self.cov(),
-                random_state=self.random_state,
-            )
+            reshaped_mean = self.mean().reshape(newshape)
         except ValueError:
             raise ValueError(
-                "Cannot reshape this normal distribution to the given shape: {}".format(
-                    str(newshape)
-                )
+                f"Cannot reshape this normal distribution to the given shape: "
+                f"{newshape}"
             )
+
+        reshaped_cov = self.cov()
+
+        if reshaped_mean.ndim > 0 and reshaped_cov.ndim == 0:
+            reshaped_cov = reshaped_cov.reshape(1, 1)
+
+        return Normal(
+            mean=reshaped_mean, cov=reshaped_cov, random_state=self.random_state,
+        )
 
     def _reshape_inplace(self, newshape):
         self.mean.shape = newshape
+
+        if self.mean().ndim > 0 and self.cov().ndim == 0:
+            self.cov().shape = (1, 1)
 
     def transpose(self, *axes):
         if len(axes) == 1 and isinstance(axes[0], tuple):
@@ -439,16 +448,14 @@ class _UnivariateNormal(Normal):
             loc=self.mean(), scale=self.std(), size=size, random_state=self.random_state
         )
 
-    def __getitem__(self, key):
-        raise NotImplementedError
-
-    def reshape(self, newshape):
-        raise NotImplementedError
+    def transpose(self, *axes):
+        return Normal(
+            mean=self.mean().copy(),
+            cov=self.cov().copy(),
+            random_state=self.random_state,
+        )
 
     def _reshape_inplace(self, newshape):
-        raise NotImplementedError
-
-    def transpose(self, *axes):
         raise NotImplementedError
 
 
--- a/src/probnum/prob/distributions/normal.py
+++ b/src/probnum/prob/distributions/normal.py
@@ -129,7 +129,7 @@ class Normal(Distribution):
         ----------
         key : int or slice or ndarray or tuple of None, int, slice, or ndarray
             Indices, slice objects and/or boolean masks specifying which entries to keep
-            while marinalizing over all other entries.
+            while marginalizing over all other entries.
         """
         if not isinstance(key, tuple):
             key = (key,)
--- a/src/probnum/prob/distributions/normal.py
+++ b/src/probnum/prob/distributions/normal.py
@@ -144,6 +144,16 @@ class Normal(Distribution):
 
         return Normal(mean=mean, cov=cov, random_state=self.random_state,)
 
+    def reshape(self, newshape):
+        try:
+            return Normal(mean=self.mean().reshape(newshape), cov=self.cov())
+        except ValueError:
+            raise ValueError(
+                "Cannot reshape this normal distribution to the given shape: {}".format(
+                    str(newshape)
+                )
+            )
+
     # Binary arithmetic
 
     def __add__(self, other):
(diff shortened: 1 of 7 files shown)


marvinpfoertner:
I noticed that the `.reshape()` call on a random variable changes the shape of the random variable in-place.

https://github.com/probabilistic-numerics/probnum/blob/1ddf55a62e2f77e1d709d42e0b158f715034261d/src/probnum/prob/randomvariable.py#L193-208

To be consistent with the numpy API, I would propose we change this to return a new random variable with the desired shape.

Any thoughts @JonathanWenger, @pnkraemer, @nathanaelbosch?

JonathanWenger:
> I noticed that the `.reshape()` call on a random variable changes the shape of the random variable in-place.
> 
> https://github.com/probabilistic-numerics/probnum/blob/1ddf55a62e2f77e1d709d42e0b158f715034261d/src/probnum/prob/randomvariable.py#L193-L208
> 
> To be consistent with the numpy API, I would propose we change this to return a new random variable with the desired shape.
> 
> Any thoughts @JonathanWenger, @pnkraemer, @nathanaelbosch?

In general we should be consistent with the numpy API, so I agree that `.reshape()` should not perform the operation in place, but return a new random variable. However `rv.shape = (a,b)` should perform the operation in place to be consistent with NumPy. 

It seems to be a bit more complicated (see also: https://stackoverflow.com/questions/46574568/does-numpy-reshape-create-a-copy). While `some_array.reshape(size)` creates a new object (a so-called view) it still shares the same data buffer, which is not the same as `reshape` creating a new array.

marvinpfoertner:
> > I noticed that the `.reshape()` call on a random variable changes the shape of the random variable in-place.
> > https://github.com/probabilistic-numerics/probnum/blob/1ddf55a62e2f77e1d709d42e0b158f715034261d/src/probnum/prob/randomvariable.py#L193-L208
> > 
> > To be consistent with the numpy API, I would propose we change this to return a new random variable with the desired shape.
> > Any thoughts @JonathanWenger, @pnkraemer, @nathanaelbosch?
> 
> In general we should be consistent with the numpy API, so I agree that `.reshape()` should not perform the operation in place, but return a new random variable. However `rv.shape = (a,b)` should perform the operation in place to be consistent with NumPy.
> 
> It seems to be a bit more complicated (see also: https://stackoverflow.com/questions/46574568/does-numpy-reshape-create-a-copy). While `some_array.reshape(size)` creates a new object (a so-called view) it still shares the same data buffer, which is not the same as `reshape` creating a new array.

I agree but since we'll call numpy's resize internally, our random variables inherit the same data buffer. Actually I'm pretty sure that a view is essentially a new array which just shares the pointer to the contiguous piece of memory of the old array.

JonathanWenger:
> > > I noticed that the `.reshape()` call on a random variable changes the shape of the random variable in-place.
> > > https://github.com/probabilistic-numerics/probnum/blob/1ddf55a62e2f77e1d709d42e0b158f715034261d/src/probnum/prob/randomvariable.py#L193-L208
> > > 
> > > To be consistent with the numpy API, I would propose we change this to return a new random variable with the desired shape.
> > > Any thoughts @JonathanWenger, @pnkraemer, @nathanaelbosch?
> > 
> > 
> > In general we should be consistent with the numpy API, so I agree that `.reshape()` should not perform the operation in place, but return a new random variable. However `rv.shape = (a,b)` should perform the operation in place to be consistent with NumPy.
> > It seems to be a bit more complicated (see also: https://stackoverflow.com/questions/46574568/does-numpy-reshape-create-a-copy). While `some_array.reshape(size)` creates a new object (a so-called view) it still shares the same data buffer, which is not the same as `reshape` creating a new array.
> 
> I agree but since we'll call numpy's resize internally, our random variables inherit the same data buffer. Actually I'm pretty sure that a view is essentially a new array which just shares the pointer to the contiguous piece of memory of the old array.

That's also my understanding from a quick scan of the topic.

marvinpfoertner:
> > > > I noticed that the `.reshape()` call on a random variable changes the shape of the random variable in-place.
> > > > https://github.com/probabilistic-numerics/probnum/blob/1ddf55a62e2f77e1d709d42e0b158f715034261d/src/probnum/prob/randomvariable.py#L193-L208
> > > > 
> > > > To be consistent with the numpy API, I would propose we change this to return a new random variable with the desired shape.
> > > > Any thoughts @JonathanWenger, @pnkraemer, @nathanaelbosch?
> > > 
> > > 
> > > In general we should be consistent with the numpy API, so I agree that `.reshape()` should not perform the operation in place, but return a new random variable. However `rv.shape = (a,b)` should perform the operation in place to be consistent with NumPy.
> > > It seems to be a bit more complicated (see also: https://stackoverflow.com/questions/46574568/does-numpy-reshape-create-a-copy). While `some_array.reshape(size)` creates a new object (a so-called view) it still shares the same data buffer, which is not the same as `reshape` creating a new array.
> > 
> > 
> > I agree but since we'll call numpy's resize internally, our random variables inherit the same data buffer. Actually I'm pretty sure that a view is essentially a new array which just shares the pointer to the contiguous piece of memory of the old array.
> 
> That's also my understanding from a quick scan of the topic.

Should I make these changes in this PR?

JonathanWenger:
Fine with me.

marvinpfoertner:
> Fine with me.

Done

src/probnum/prob/distributions/normal.py
```diff
@@ -375,14 +448,15 @@ def sample(self, size=()):
...
-        self._shape = newshape
+    def transpose(self, *axes):
+        return Normal(
+            mean=self.mean().copy(),
+            cov=self.cov().copy(),
+            random_state=self.random_state,
+        )
+
+    def _reshape_inplace(self, newshape):
+        raise NotImplementedError
```
marvinpfoertner, line 459: Inplace reshaping for univariate normals is actually impossible due to the fact that the attributes of numpy's array scalars are immutable. See also https://numpy.org/doc/stable/reference/arrays.scalars.html. We should overthink #130. It would be best if scalar quantities behave exactly the same as arrays. Maybe we should use arrays of shape () internally?

marvinpfoertner: @JonathanWenger

tests/test_prob/test_distributions/test_normal.py
```diff
@@ -180,6 +225,302 @@ def test_symmetric_samples(self):
...
+
+        # Test covariance
+        for ii, ij in itertools.product(range(2), range(2)):
+            for ji, jj in itertools.product(range(2), range(2)):
+                idx = (2 * ii + ij, 2 * ji + jj)
+                idx_t = (2 * ij + ii, 2 * jj + ji)
+
+                self.assertEqual(dist_t.cov()[idx_t], dist.cov()[idx])
+
+        # Sadly, sampling is not stable w.r.t. permutations of variables
```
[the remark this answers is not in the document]
marvinpfoertner: Discussed in person. Opened #146

marvinpfoertner:
@JonathanWenger I resolved all your comments. The PR would be ready to merge. Any objections?

src/probnum/prob/distributions/normal.py
JonathanWenger: Yes i think representing everything internally as arrays makes a lot of sense. This removes quite a few annoying conditionals in the code as well.

marvinpfoertner: Alright, I'll rephrase #130 then?

PR #135, opened 2020-08-11
Univariate normal should have numpy scalars as parameters

### LINKED ISSUE #94: _UnivariateGaussian should only be constructed for scalars
IMHO a mean (and cov) with shape (1, 1) should construct a _MatrixVariateNormal and a mean with shape (1,) should construct a multivariate normal to be consistent with numpy. This would also make slicing much more intuitive.

https://github.com/probabilistic-numerics/probnum/blob/f13f26cf441e4182339faa2bc2b8347da77960c8/src/probnum/prob/distributions/normal.py#L271-L280

### LINKED ISSUE #130: Only use np.float internally

marvinpfoertner:
Closes #130 (for the univariate normal)
Closes #94.


marvinpfoertner:
@nathanaelbosch As you predicted, several tests are failing in diffeq, how should I go about this?

Issue #112, opened 2020-07-22
Benchmarks should be test run by Travis


JonathanWenger:
If somebody changes the API or adds a new benchmark, these may potentially break. Therefore Travis via tox should run the benchmarks once in dev / fast mode.

JonathanWenger:
Implemented by #143.

Issue #127, opened 2020-07-28
TravisCI does not seem to run tests with Python 3.6


nathanaelbosch:
Found this when looking into this build: https://travis-ci.org/github/probabilistic-numerics/probnum/builds/712479064

The Python 3.6 run seems only to run the `docs` tox env, while the 3.7 and 3.8 runs run tests. Not a huge issue, since we're still testing, but this is not the desired behaviour.

JonathanWenger:
Is it possible that this is due to an extra space in the config? From looking at the default `envlist` 
```ini
[tox]
envlist = py3,docs,benchmarks
```
which is executed when I run `tox` locally, should the `travis` environment in `tox.ini` instead of 
```ini
[travis]
python =
  3.6: py36, docs
  3.7: py37
  3.8: py38
```
maybe look more like this
```
[travis]
python =
  3.6: py36,docs,benchmarks
  3.7: py37,docs,benchmarks
  3.8: py38,docs,benchmarks
```
This is also related to #112 and therefore #137 .

nathanaelbosch:
I would not expect it, but yes it could be that the whitespace influences that, I could check that tomorrow. But do we actually want to run docs and benchmarks on all python versions? I would have thought that this is only really relevant for the test suite.

JonathanWenger:
I could imagine cases where the Python version breaks 'asv' functionality or one of the sphinx addons used to build the documentation. But i'm not sure what the best practice is here. Since the builds run in parallel on Travis this shouldn't increase the development loop times much.

JonathanWenger:
Whitespace is not the problem. From a quick google search this seems to be a problem with the `tox.ini` setup for Travis. If i make the suggested change above the docs and benchmarks are run but not the tests for any environment.

nathanaelbosch:
I found something: Adding `py36,py37,py38` to the `envlist` in `tox.ini` resolves this issue (see https://github.com/nathanaelbosch/probnum/commit/e1477eb44924659625845efbd0cc8d3c37f7ffd0 and the corresponding Travis run https://travis-ci.org/github/nathanaelbosch/probnum).

I see two fixes here:
1. Add `py36,py37,py38` back into the envlist again. This however will lead to errors if people just run `tox` to quickly check if things seem fine, but don't have all python versions installed, which is why we moved to just having `py3` in the envlist.
[...]

PR #137, opened 2020-08-11
Dry run benchmarks using tox to discover potential bugs

JonathanWenger:
This PR adds a `benchmarks` environment to `tox` which makes a quick dry run of the benchmarks in the current environment defined by tox. This checks whether benchmarks execute properly or not (see also https://asv.readthedocs.io/en/stable/writing_benchmarks.html#running-benchmarks-during-development).


JonathanWenger:
This closes #112.

JonathanWenger:
It's probably a good idea early on where interfaces might be changed or people add new benchmarks. This is somewhat different from the build, docs and tests which are effected by nearly all code changes. Long-term this will likely only extend tox runtime with little benefit.

nathanaelbosch:
True, I also think we should add the `benchmarks` env to the CI workflow!

tox.ini
```diff
@@ -4,13 +4,13 @@
...
-envlist = py3,docs
+envlist = py3, docs, benchmarks
 
 [travis]
 python =
-  3.6: py36, docs
-  3.7: py37
-  3.8: py38
+  3.6: py36, docs, benchmarks
+  3.7: py37, docs, benchmarks
```
nathanaelbosch, line 12: To not mix too many things regarding the ongoing discussion in #127, I would propose to stick to the previous behaviour to merge this PR, so adding `benchmarks` just to the `envlist` and 3.6, but not to 3.7 and 3.8. Depending on the conlcusion in #127 this could then be fixed alter in a different PR.

JonathanWenger: Good point. Reverted back to the previous behavior.

nathanaelbosch:
Sorry if my previous statement was confusing. In order to actually run the benchmarks in Travis, could you add `benchmarks` to `3.6` in line 11 again?

Issue #95, opened 2020-07-16
Automatically check whether docstrings are complete and correctly formatted


JonathanWenger:
~Use pytest, e.g. `pytest --doctest-glob='*.rst'. ` during the Travis build.~

Also look into `pydocstyle` and `darglint`. These check whether docstrings are complete and correctly formatted. Output looks like this:

```
backpack/core/derivatives/basederivatives.py:jac_t_mat_prod:63: DAR101: Missing parameter(s) in Docstring: - module
backpack/core/derivatives/basederivatives.py:jac_t_mat_prod:63: DAR101: Missing parameter(s) in Docstring: - g_inp
backpack/core/derivatives/basederivatives.py:jac_t_mat_prod:63: DAR101: Missing parameter(s) in Docstring: - g_out
backpack/core/derivatives/basederivatives.py:jac_t_mat_prod:63: DAR101: Missing parameter(s) in Docstring: - mat
backpack/core/derivatives/basederivatives.py:jac_t_mat_prod:63: DAR201: Missing "Returns" in Docstring: - return
backpack/core/derivatives/basederivatives.py:_jac_t_mat_prod:82: DAR101: Missing parameter(s) in Docstring: - module
backpack/core/derivatives/basederivatives.py:_jac_t_mat_prod:82: DAR101: Missing parameter(s) in Docstring: - g_inp
backpack/core/derivatives/basederivatives.py:_jac_t_mat_prod:82: DAR101: Missing parameter(s) in Docstring: - g_out
backpack/core/derivatives/basederivatives.py:_jac_t_mat_prod:82: DAR101: Missing parameter(s) in Docstring: - mat
backpack/core/derivatives/basederivatives.py:_jac_t_mat_prod:82: DAR401: Missing exception(s) in Raises section: -r NotImplementedError
```

_Originally posted by @JonathanWenger in https://github.com/probabilistic-numerics/probnum/issues/14#issuecomment-614483553_

nathanaelbosch:
It should be easy to include running `pydocstyle` (or whatever tool we end up using) into tox, and thereby into Travis :) 

Less easy: Fixing all the errors that `pydocstyle` raises when I run it on the `development` branch :fearful:

nathanaelbosch:
With the `pylint` PR merged (#150), I see the following points for improving the docstring situation:
1.  Write missing docstrings: `pyling src --enable="missing-module-docstring,missing-function-docstring"` should pass. 
[...]

Issue #69, opened 2020-06-30
Remove comments and todos in development branch


JonathanWenger:

nathanaelbosch:
The linked PR is merged. Is there still anything actionable here? If yes it might be good to explicitly write what should be done, if not we should close this.

JonathanWenger:
While the previously linked PR #59  is merged there are still tons of TODOs in `master`e.g. in `probnum.linalg.linearsolvers.matrixbased`.

nathanaelbosch:
The occurrence of TODOs can now be checked more easily, with 
```
pylint src --enable="fixme"
```
So the part about "removing todos" is well defined, as having this command pass without errors. Once done, the `./pyproject.toml` should also be adjusted to include this message into future pylint runs.

@JonathanWenger Do you have a similar definition in mind for "removing comments"? I'm not sure that all comments should actually be removed, as there might be some valid cases. See eg. the [google style guide](https://google.github.io/styleguide/pyguide.html#385-block-and-inline-comments) (which also has a nice part about [todo comments](https://google.github.io/styleguide/pyguide.html#312-todo-comments) btw)

nathanaelbosch:
Update: I added the "fixme" pylint exception only to specific modules. Resolving these is therefore also a part of #158, #159, #160.

JonathanWenger:
There seem to be some legitimate situations that warrant TODO comments in the code in my opinion. For example for some functions of random variables there is in principle a way to compute them and even a paper reference, but no capacity to implement them at this point. I'd suggest we stick to the Google style guide wherever possible, so the actionable part of this issue is to add names to all remaining TODO comments in the google style: 
```python
# TODO(kl@gmail.com): Use a "*" here for string repetition.
# TODO(Zeke) Change this to use relations.
```

Issue #131, opened 2020-07-28
Remove Distribution class and only use RandomVariable with subclasses


JonathanWenger:
In the current usage of `RandomVariable` and `Distribution` there is no reason for the existence of two separate classes beyond semantic separation. The current implementation also forces a lot of duplicate implementation of the same functionality in both classes.

Therefore remove `Distribution` and subclass `RandomVariable` with classes like `NormalRandomVariable` instead. This is also closer to an envisioned implementation of `GaussianProcess` inheriting from `RandomProcess`.

JonathanWenger:
The original separation was in part inspired by TensorFlow Probability, where `RandomVariable` is the analogue to a tensor and also a `Distribution` class exists.

https://www.tensorflow.org/probability/api_docs/python/tfp/edward2/RandomVariable
https://www.tensorflow.org/probability/api_docs/python/tfp/edward2/Distribution

nathanaelbosch:
One pylint message that only relates to this part of the code, and that might therefore be fixed as part of this refactoring: `broad-except`.
Check with `pylint src --enable="broad-except"`.
In case you resolve it, please also the exception from the `./pyproject.toml` :)

nathanaelbosch:
Instead of my previous comment: When working on this refactor, it might be useful to look at #159 from time to time, since this issue might resolve some of the current pylint exceptions :)

JonathanWenger:
> Instead of my previous comment: When working on this refactor, it might be useful to look at #159 from time to time, since this issue might resolve some of the current pylint exceptions :)

@marvinpfoertner I assume you have done so, but just to make sure can you do a check in #164 for this?

Issue #83, opened 2020-07-14
Use compositionality instead of subclassing when defining Probabilistic numerical methods


JonathanWenger:
Rewrite development template for a probabilistic numerical method by passing callables instead of subclassing.

PR #150, opened 2020-08-18
Code Linting with Pylint

### LINKED ISSUE #108: Use flake8 to better verify PEP 8 compliance

nathanaelbosch:
Closes #108 (and relates to #111).

__Current state of the PR__
I created configs, added a github action, and started fixing many errors. However I did not fix *all* issues, and often chose to add exceptions. I think that the best way would be to discuss these exceptions, fix those that should *absolutely* go into this PR, and then open specific issues about individual points.

__Commentary for each of the exceptions:__
- `invalid-name`: We will probably want to keep this, since we often want to use *mathy* notation
- `no-else-return`, `no-else-raise`: Not so clear to me what the benefit of fixing this is
- `fixme`: Can be resolved through the separate issue #69
- `too-many-arguments`: Appears accross modules. Good code style improvement, but not crucial. We could also raise the number of arguments from  to 6 or 7 to resolve most warnigns, and to put the focus towards the cases with 12 (`distribution.py`) and 10 (`linearsolvers.py`) arguements.
- `missing-module-docstring`, `missing-function-docstring`: Could be considered as part of #95, or we can also create a new issue. Although I have the impression that noone will do this at once.
- `abstract-method`: Needs to be fixed in `probnum.linalg` and `probnum.prob`
- `unused-argument`: Many of these are cases where `**kwargs` gets passed around. I had the impression that this is often the "standard" in our code so I did not go ahead and fix it.
- `too-many-branches`: Only affects `linearsolves.py` and `matrixbased.py`. I'd just open an issue for this specific point.
- `arguments-differ`: Multiple cases in `linalg`, but also appearsx in `prob`, `quad` and `diffeq`
- `redefined-builtin`: In `quad` and `linalg`; Fixing this will change some function signatures
- `too-few-public-methods`: Mostly `quad`, but also `diffeq.ODESolver`; Suggests considering removing these classes
- `too-many-locals`: Seems like a good tool to improve code, but is also non-trivial to fix
- `no-self-use`: Suggests making functions instead of methods
- `too-many-lines`: Suggests refactoring `matrixbased`
- `broad-except`: Only appears in `normal.py`. Should be easier to fix for someone familiar with that code.
- `abstract-class-instantiated`: Only appears in `diffeq` due to Nico's code structure; I think it would be best if he has a look.
- `protected-access`: Something we probably want, but it's also not trivial or fast to fix
- `duplicate-code`: Only appears in `filtsmooth`
- `too-complex`: McCabe complexity checker; Complains for `matrixbased`
- `too-many-instance-attributes`: Indicated that `matrixbased.py` and `distribution` might need some refactoring
- `too-many-statements`: Again in `matrixbased`
- `attribute-defined-outside-init`: Mostly in `matrixbased`. One appearance in `bayesquadrature.py`, but that code seems empty anyways.

__More exceptions for tests:__
`duplicate-code`, `line-too-long`, `missing-class-docstring`, `unnecessary-pass`. I did not fix them becasue it did not feel too important for now, but especially the `line-too-long` and `unnecessary-pass` should be easy to fix quite mechanically. Also `unused-variable`, but in tests we often want to compute some value without actually using it afterwards, so that seems reasonable.

__Next steps__
I think it would be good to merge a first version as soon as possible, and just open issues for the parts that we want to have fixed.
But it would be good to know if you think that we should still resolve one of the current exceptions in this PR.
If we decide to merge, I'll open individual issues for the remaining points.

Also, if you have any suggestion on how to better include pylint into our CI let me know!


JonathanWenger:
In general, I think all these issues raised should be fixed (obviously) but none in this PR specifically. For some there are already issues (e.g. for `linalg`: #51) and some might be fixed by larger changes in progress at the moment (e.g. #83, #131). In summary I agree this PR should be merged and some of the points above moved into (existing) issues.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -88,10 +88,10 @@ def __call__(self, t, smoothed=True):
             else:
                 return pred_rv
 
-        elif t > self.locations[-1]:
-            if smoothed:
-                warn("`smoothed=True` is ignored for extrapolation.")
-            return self._predict_to_loc(t)
+        # else: t > self.locations[-1]:
+        if smoothed:
```
JonathanWenger, line 92: Should this be
```
if smoothed and t > self.locations[-1]:
```
to retain the same behaviour? Clearly no tests are failing so I'm not sure, but according to codecov this is also not covered.

src/probnum/filtsmooth/statespace/continuous/linearsdemodel.py
```diff
@@ -25,15 +25,15 @@ class LinearSDEModel(continuousmodel.ContinuousModel):
 
     Parameters
     ----------
-    driftmatrixfct : callable, signature=(t, \*\*kwargs)
+    driftmatrixfct : callable, signature=(t, **kwargs)
```
JonathanWenger, line 28: These changes cause warnings in Sphinx (see https://travis-ci.org/github/probabilistic-numerics/probnum/jobs/718931687):
```bash
/home/travis/build/probabilistic-numerics/probnum/src/probnum/filtsmooth/__init__.py:docstring of probnum.filtsmooth.LinearSDEModel:9: WARNING: Inline strong start-string without end-string.

/home/travis/build/probabilistic-numerics/probnum/src/probnum/filtsmooth/__init__.py:docstring of probnum.filtsmooth.LinearSDEModel:13: WARNING: Inline strong start-string without end-string.

/home/travis/build/probabilistic-numerics/probnum/src/probnum/filtsmooth/__init__.py:docstring of probnum.filtsmooth.LinearSDEModel:17: WARNING: Inline strong start-string without end-string.

/home/travis/build/probabilistic-numerics/probnum/src/probnum/filtsmooth/gaussfiltsmooth/unscentedtransform.py:docstring of probnum.filtsmooth.UnscentedTransform.propagate:8: WARNING: Inline strong start-string without end-string.
```

 and likely do not produce the docs output we are looking for.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: The "if" statements before check for `t < self.locations[0]` and `self.locations[0] < t < self.locations[-1]`, so the check is not actually necessary. I just wrote it that way as first to make the code more readable, and thought now that the comment achieves the same thing.

src/probnum/filtsmooth/statespace/continuous/linearsdemodel.py
nathanaelbosch: My bad, thought I had checked those things but apparently not for all of those string changes! I'll look into fixing that such that the builds look as they should, while pylint does not complain.

nathanaelbosch: Solved this in nathanaelbosch:probnum/ pylint @6e847a5

nathanaelbosch:
A few things that should be done as a part of this PR:
- perform the pylint check on a "per-module" basis. This allows us to check for things that are currently violated in one of the modules and make sure that we do not introduce new issues.
- Open refactoring issues for the individual modules. Fixing these errors on a per-module basis is easier. Also, we can raise the standards after fixing each individual error, for each module separately.

JonathanWenger (approved):

Issue #159, opened 2020-08-19
Refactor `probnum.randvars` to resolve the pylint messages


nathanaelbosch:
At the time of writing, `probnum.randvars` triggers the following `pylint` messages:
- [x] ~~`too-many-instance-attributes`~~
- [x] ~~`broad-except`~~
- [x] ~~`arguments-differ`~~
- [x] ~~`abstract-method`~~
- [x] ~~`too-many-arguments`~~
- [x] ~~`protected-access`~~
- [x] ~~`duplicate-code`~~
- [x] ~~`unused-argument`~~
- [x] ~~`missing-module-docstring`~~
- [ ] `missing-function-docstring`
- [x] ~~`raise-missing-from`~~

__Goal:__
Our long-term goal is to enable all of these checks for all parts of the project. Therefore, we should continuously work on fixing these. Currently, pylint is set up to have different checks for different modules, such that we can work on these problems individually and progress more quickly :)

__To work on this issue__ you can check for these messages individually by calling:
```
pylint src/probnum/prob --disable=all --enable=<message>
```
Then, after fixing all errors/warnings, you should remove the exception from the corresponding line in `./tox.ini`, such that this message will be tracked in the future.

marvinpfoertner:
Most of these points will be resolved by #164.

marvinpfoertner:
Superseded by #523

Issue #61, opened 2020-06-26
Type Hints


JonathanWenger:
All internal classes and methods should have type hints (https://docs.python.org/3/library/typing.html). This prevents pesky type bugs. 

Example method with a `str` type hint:
```python
def greeting(name: str) -> str:
    return 'Hello ' + name
```

pnkraemer:
Only the internal ones? Why not all visible ones? I would find it hard to distinguish purely internal functions from purely public functions but maybe I am not trying hard enough :)

JonathanWenger:
Closing as this is now standard practice when accepting PRs and no longer an active TODO.

Issue #57, opened 2020-05-07
Sampling from random variables returns inconsistent size


JonathanWenger:
Currently sampling from univariate, multivariate and matrix-variate normal distributions has inconsistent behavior for different `size` arguments. Ensure that all sampling from random variables follows the behavior that:
- () returns a single sample
- 1 returns an array of shape (1,) with a single sample
- n returns an array of shape (n,)
- (1,) returns an array of shape (1,) with a single sample 
- (1,1) returns an array of shape (1,1) with a single sample

JonathanWenger:
Further, always keep the sampling dimensions as the first dimensions, such that a statement like `np.shape(sample_)` has the value of `size` as its first components.

JonathanWenger:
- [ ] Write tests to test all of the above (already for distributions or just random variables?)

Issue #67, opened 2020-06-26
Useful error messages


JonathanWenger:
Whenever an error is thrown, useful output about the object causing the error should be provided. For example if a wrong shape is given then the error message must contain the shape of the actual object and the expected shape.

PR #164, opened 2020-08-20
Refactor `RandomVariable`

### LINKED ISSUE #131: Remove Distribution class and only use RandomVariable with subclasses

### LINKED ISSUE #145: Refactor Normal class to compositional pattern
Instead of having a class hierarchy create a single `Normal` class which in its constructor composes different versions of normal by passing different implementations of `sample`, `pdf`, ...

This ensures that
- if `x=Normal(mean, cov)` is called, the type of `x` is `Normal`
- the user does not need to care about different types of normal distributions

The different classes implementing different normal distributions currently only exist to implement efficient sampling, pdf computation, etc. The refactored class would have one large ```__init__``` but the rest of the class is simpler.

marvinpfoertner:
In essence, this PR is a substantial refactor of the `RandomVariable` and `Distribution` classes that:
* gets rid of `Distribution`s and integrates their functionality into the `RandomVariable` class (closes #131)
* creates a corresponding `RandomVariable` subclass for every `Distribution` subclass (closes #131)
* gets rid of the `probnum.prob` module and introduces the new `probnum.random_variables` module
* introduces a functional/compositional pattern in `RandomVariable` and `Normal` to allow for flexible and efficient recombination of sampling, mean/cov/var/entropy computation, and (log){cd,pf,pm}f evaluation (closes #145)
* introduces a flexible way of implementing binary arithmetic between random variables and provides generic fallback operations for unary and binary arithmetic and for reshaping, transposition, (advanced) indexing, slicing and masking of random variables
* introduces `cached_property` decorators for all random variable properties in order to save computation time in case the distribution of the random variable is implicity given, e.g. via an MCMC sampler
* introduces type hints for the new `RandomVariable` class, its subclasses and all related parts of the API. It also introduces typing infrastructure in the `probnum.typing` and `probnum.utils.typeutils` modules (addresses #61)
* fixes almost all pylint messages that had to be disabled in #159. The remaining disabled message will be fixed in a follow-up documentation PR

This should also fix the bug about the sampling shapes, i.e. #57, but this is not tested yet.

Many of the changes are not yet documented properly. This will be done in a follow-up PR by @JonathanWenger with my support.


src/probnum/utils/randomutils.py
```diff
@@ -0,0 +1,20 @@
+from typing import Union
+
+import numpy as np
+
+
+def derive_random_seed(*args: Union[np.random.RandomState, np.random.Generator]) -> int:
```
JonathanWenger, line 6: `combine_random_seeds` seems like a more descriptive name to me, similar to `boost::hash_combine`

src/probnum/prob/random_variable/_dirac.py
```diff
@@ -0,0 +1,154 @@
...
+                self._support,
+                shape=(
+                    (self._support.size, self._support.size)
+                    if self._support.ndim > 0
+                    else ()
+                ),
+            ),
+            var=lambda: np.zeros_like(self._support),
+            properties={
+                "mode": self._support,  # TODO: Is this correct? Diracs don't have a pdf
```
JonathanWenger, line 72: I think this is sensible _definition_ if one interprets the dirac delta function as the limit of a Gaussian pdf.

src/probnum/prob/_random_variable.py
```diff
@@ -0,0 +1,904 @@
...
+        logcdf: Optional[Callable[[ValueType], float]] = None,
+        quantile: Optional[Callable[[float], ValueType]] = None,
+        mode: Optional[Callable[[], ValueType]] = None,
+        median: Optional[Callable[[], ValueType]] = None,
+        mean: Optional[Callable[[], ValueType]] = None,
+        cov: Optional[Callable[[], ValueType]] = None,
+        var: Optional[Callable[[], ValueType]] = None,
+        std: Optional[Callable[[], ValueType]] = None,
+        entropy: Optional[Callable[[], float]] = None,
+        properties: Optional[Dict[str, Any]] = None,
```
JonathanWenger, line 89: Let's discuss again whether it is necessary to have this `properties` arguments and if yes about its naming. It can be somewhat confusing what is acutally meant (e.g. rv properties or Python properties) when talking about `RandomVariable`'s properties.

src/probnum/prob/_random_variable.py
```diff
@@ -0,0 +1,904 @@
...
+        if shape is None:
+            return None
+        elif isinstance(shape, tuple) and all(
+            isinstance(entry, int) for entry in shape
+        ):
+            return shape
+        elif isinstance(shape, int):
+            return (shape,)
+        else:
+            raise TypeError("The given shape is not an int or a tuple of ints.")
```
JonathanWenger, line 144: I'd suggest to return the given shape in the Error message to make debugging easier, also in the light of #67 , i.e.
```python
raise TypeError(f"The given shape {str(shape)} is not an int or a tuple of ints.")
```

JonathanWenger:
In order to merge this PR the following TODOs need to be resolved:

- [ ] Update benchmarks (in particular the folder named `distribution`) to a more appropriate naming scheme
- [ ] Ensure all notebooks work, have been re-run and use the way we intend users to import `Dirac`, `Normal`, etc.
- [ ] Update the quickstart guide and notebooks explaining the use of random variables
- [ ] Update the documentation thoroughly in all changed files

src/probnum/prob/_random_variable.py
nathanaelbosch: I'm not here to actually review but since I saw it while passing: Calling `str` explicitly should not be necessary for this

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -121,16 +121,16 @@ def test_predict_value(self):
...
-        self.assertApproxEqual(expectedmean, pred.mean())
-        self.assertApproxEqual(expectedcov, pred.cov())
+        expectedmean = np.squeeze(ah @ (self.initrv.mean * np.ones(1)))
+        expectedcov = np.squeeze(ah @ (self.initrv.cov * np.eye(1)) @ ah.T + qh)
+        self.assertApproxEqual(expectedmean, pred.mean)
+        self.assertApproxEqual(expectedcov, pred.cov)
 
     def test_update(self):
-        data = np.array([self.measmod.sample(0.0, self.initrv.mean() * np.ones(1))])
+        data = np.array([self.measmod.sample(0.0, self.initrv.mean * np.ones(1))])
```
nathanaelbosch, line 130: Fix für den test: Das äußere `np.array` macht `data` hier zu einer Matrix, und damit wird in `Kalman` auch irgendwas matrix-valued. Ich glaub das `np.array` kann einfach weg, ich seh jetzt keinen Grund wozu man das bräuchte.

Die Zeile wäre dann: 
```python
        data = self.measmod.sample(0.0, self.initrv.mean * np.ones(1))
```
Damit sollten die tests laufen :)

tox.ini
```diff
@@ -58,7 +58,7 @@ commands =
          pylint src/probnum/diffeq --disable="protected-access,abstract-class-instantiated,too-many-locals,too-few-public-methods,too-many-arguments,unused-argument,missing-module-docstring,missing-function-docstring"
          pylint src/probnum/filtsmooth --disable="duplicate-code,protected-access,no-self-use,too-many-locals,arguments-differ,too-many-arguments,unused-argument,missing-module-docstring,missing-function-docstring"
          pylint src/probnum/linalg --disable="attribute-defined-outside-init,too-many-statements,too-many-instance-attributes,too-complex,protected-access,too-many-lines,no-self-use,too-many-locals,redefined-builtin,arguments-differ,abstract-method,too-many-arguments,too-many-branches,duplicate-code,unused-argument,fixme,missing-module-docstring"
-         pylint src/probnum/prob --disable="too-many-instance-attributes,broad-except,arguments-differ,abstract-method,too-many-arguments,protected-access,duplicate-code,unused-argument,fixme,missing-module-docstring,missing-function-docstring"
+         pylint src/probnum/prob --disable="protected-access,fixme,missing-module-docstring,missing-function-docstring"
```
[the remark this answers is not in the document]
marvinpfoertner: The former. This is fixing a few points from #159.

src/probnum/prob/_random_variable.py
marvinpfoertner: As discussed in person, I removed the `properties`.

src/probnum/prob/random_variable/_dirac.py
marvinpfoertner: Now that `Dirac`s are `DiscreteRandomVariables`, we should not need to define this anymore, since this is also the mode of the `pmf`, right?

tests/test_random_variables/test_normal.py
```diff
@@ -62,22 +63,29 @@ def setUp(self):
         sparsemat = scipy.sparse.rand(m=m, n=n, density=0.1, random_state=1)
         self.normal_params = [
             # Univariate
-            (-1, 3),
+            (-1.0, 3.0),
             # Multivariate
```
JonathanWenger, line 67: Ich würde hier noch eine Parameterkombination mit `int` hinzufügen um zu testen ob man auch mit nicht `float`s `Normal` aufrufen kann.

tests/test_random_variables/test_normal.py
```diff
@@ -234,52 +239,47 @@ def test_symmetric_samples(self):
...
-            if isinstance(
-                dist,
-                (
-                    prob.distributions.normal._UnivariateNormal,
-                    prob.distributions.normal._OperatorvariateNormal,  # TODO: Implement slicing on linear operators
-                ),
+            if isinstance(rv.mean, linops.LinearOperator) or isinstance(
+                rv.cov, linops.LinearOperator
             ):
+                # TODO: Implement slicing on linear operators
```
JonathanWenger, line 247: Can you create an issue for this missing test and reference it in this TODO?

tests/test_random_variables/test_normal.py
```diff
@@ -234,52 +239,47 @@ def test_symmetric_samples(self):
...
```
JonathanWenger, line 282: See above.

src/probnum/random_variables/_dirac.py
```diff
@@ -0,0 +1,167 @@
...
+
+        super().__init__(
+            shape=self._support.shape,
+            dtype=self._support.dtype,
+            random_state=random_state,
+            parameters={"support": self._support},
+            sample=self._sample,
+            in_support=lambda x: np.all(x == self._support),
+            pmf=lambda x: np.float_(1.0 if np.all(x == self._support) else 0.0),
+            cdf=lambda x: np.float_(0.0 if np.any(x < self._support) else 0.0),
```
JonathanWenger, line 76: This should be:
```python
cdf=lambda x: np.float_(0.0 if np.any(x > self._support) else 1.0)
```
and probably also get a test to avoid this error in the future.

src/probnum/random_variables/_scipy_stats.py
```diff
@@ -0,0 +1,225 @@
...
+_ValueType = Union[np.generic, np.ndarray]
+
+
+class _SciPyRandomVariableMixin:
+    @property
+    def scipy_rv(self):
+        return self._scipy_rv
+
+
+class WrappedSciPyRandomVariable(
```
JonathanWenger, line 20: Currently this class is relevant if one uses an RV that has an implementation in SciPy but not in ProbNum. Long term these should converge to zero. Should this class then be removed? Do I understand correctly that the only advantage of having these wrappers over just returning a `RandomVariable` is the access to the Scipy object?

src/probnum/random_variables/_utils.py
```diff
@@ -0,0 +1,64 @@
+import numpy as np
+import scipy.sparse
+
+from . import _random_variable
+from . import _dirac
+from . import _scipy_stats
+
+
+def asrandvar(obj) -> _random_variable.RandomVariable:
```
JonathanWenger, line 9: I assume you are type hinting `_random_variable.RandomVariable` instead of `probnum.RandomVariable` to avoid circular dependencies. From a quick local test this might break links in the automatic documentation of types with Sphinx, i.e. you get the type `RandomVariable` but it does not actually link to the `RandomVariable` class. I'm not sure whether this is solvable with forward references.

benchmarks/random_variables.py
```diff
@@ -88,18 +87,18 @@ class Sampling:
     Benchmark sampling routines for various distributions.
     """
 
-    param_names = ["dist"]
-    params = [distribution_names]
+    param_names = ["rv_name"]
```
JonathanWenger, line 90: Here I'd suggest to use `"rv"` since this actually shows up in the benchmark interface as the identifier.

JonathanWenger: Alternatively "randvar"

src/probnum/random_variables/_dirac.py
marvinpfoertner: Pretty sure this is correct. Imagine a 2d Dirac mass at [1, 1]. Then I should get a cdf of 0 for e.g. [0.5, 2] since the "rectangle" given by the integration boundaries does not encompass the point mass. The Dirac mass is only contained in the integration domain if all `x` coordinates surpass the coordinates of the point mass.

src/probnum/random_variables/_utils.py
marvinpfoertner: My local test actually links this correctly, that's weird! How should we go about this?

src/probnum/random_variables/_scipy_stats.py
marvinpfoertner: Yeah, so one advatage is, as you say, that you can access the scipy object.
I'd also say that it is convenient to have specific types for wrapped scipy random variables because you can now filter them out by their types in interfaces, etc. This might come in handy at a later point.

I would say that we should keep these classes even if we implement all scipy random variables because afterall scipy might decide to add more RVs in future versions. Then we can still support these rvs even if we don't yet have a `probnum` counterpart.

marvinpfoertner: I would say that we should keep these classes even if we implement all scipy random variables because afterall scipy might decide to add more RVs in future versions. Then we can still support these rvs even if we don't yet have a `probnum` counterpart.

src/probnum/random_variables/_dirac.py
JonathanWenger: Of course, my bad! But the `0.0 if .. else 0.0` should still be:
```python
cdf=lambda x: np.float_(0.0 if np.any(x < self._support) else 1.0)
```
Note the `1.0`.

src/probnum/random_variables/_utils.py
JonathanWenger: hmm strange. Since this is a doc issue i can fix it when going through the docs.

src/probnum/random_variables/_normal.py
```diff
@@ -0,0 +1,582 @@
...
+        )
+
+    # Matrixvariate Gaussian with Kronecker covariance
+    def _kronecker_cov_cholesky(self) -> linops.Kronecker:
+        assert isinstance(self._cov, linops.Kronecker)
+
+        A = self._cov.A.todense()
+        B = self._cov.B.todense()
+
+        eps = 10 ** -12  # damping needed to avoid negative definite covariances
```
JonathanWenger, line 540: ideally this should be a global constant and longterm maybe even adjusted based on the precision used. For now at least a constant across normal distributions would be good.
```python
COV_CHOLESKY_DAMPING = 10 **-12
```

JonathanWenger (approved):

src/probnum/random_variables/_dirac.py
marvinpfoertner: Ah yeah sure, I'll fix that!

src/probnum/random_variables/_dirac.py
```diff
@@ -73,7 +73,7 @@ def __init__(
             sample=self._sample,
             in_support=lambda x: np.all(x == self._support),
             pmf=lambda x: np.float_(1.0 if np.all(x == self._support) else 0.0),
-            cdf=lambda x: np.float_(0.0 if np.any(x < self._support) else 0.0),
+            cdf=lambda x: np.float_(1.0 if np.all(x >= self._support) else 0.0),
```
JonathanWenger, line 76: Super petty point: your original formulation (with 1.0 corrected), i.e.
```python
cdf=lambda x: np.float_(0.0 if np.any(x < self._support) else 1.0)
```
might be faster on average since `np.any` can exit without checking all elements of the boolean input array while `np.all` has to check all elements. I understand that this way it is more readable. I'll leave it to you what you prefer.

JonathanWenger (approved):

JonathanWenger (approved):

tests/test_random_variables/test_normal.py
marvinpfoertner: I instead implemented (advanced) indexing, slicing, and masking for `Normal` random variables using `todense` and removed the comment altogether :)

tests/test_random_variables/test_normal.py
marvinpfoertner: See above.

tests/test_random_variables/test_normal.py
marvinpfoertner: I will still open an issue though

PR #168, opened 2020-08-24
Airspeed velocity benchmark for IVP solver

pnkraemer:
This is a starting point for IVP solving benchmarks via airspeed velocity. It measures time and peak memory for eks0, ekf0 with and without preconditioning on a lotka volterra system with small steps and high order integrator (q=4). The latter is there to capture the linear algebra overhead. Thoughts, @nathanaelbosch ? I think we can (and should) build on this soon, but for the time being I think it is good to have at least a mini benchmark.

One thing I am not sure about is how long a benchmark should take. The longer, the more accurately changes can be measured, but for obvious reasons, too long is a waste of time and energy. Currently, the step sizes etc. are chosen such that each of the 4 benchmarks results in 0.5-1 sec. See image below. I noticed the benchmarks for linear solvers are much faster. Thoughts, @JonathanWenger ? 

Screenshot from 2020-08-24 13-37-05


nathanaelbosch:
Nice initiative! I also think it would be good to have a benchmark for the ODE solvers. 

I have one question, but probably just due to the fact that I didn't try out airspeed velocity myself so far: Since we have both `peakmem_solve` and `time_solve` does this result in running the benchmark twice, to measure time and memory separately? Intuitively I would have assumed that we should only need to run the solver once to get both of these values.

pnkraemer:
> I have one question, but probably just due to the fact that I didn't try out airspeed velocity myself so far: Since we have both peakmem_solve and time_solve does this result in running the benchmark twice, to measure time and memory separately? Intuitively I would have assumed that we should only need to run the solver once to get both of these values.

I am not sure, but the documentation (https://asv.readthedocs.io/en/stable/using.html) says

> To improve reproducibility, each benchmark is run in its own process.

pnkraemer:
I two changes: 1) benchmarks are made smaller (step is larger, order is 3 instead of 4) and 2) different priors are now benchmarked. The former is because it was super tedious to run all the benchmarks on my local machine and the latter is because I intend to make IOUP and Matern faster, which would be a good thing to see in the benchmarks

nathanaelbosch (changes requested):
Thanks for your comment to describe the changes!

One more idea: Maybe we could remove the if-else on the `precond == "with"` condition, by instead setting `self.precond_step = self.stepsize if precond == "with" else 1.0`, and then use `self.precond_step`? This could remove some duplication and make `time_solve` and `peakmem_solve` more readable. But as for previous comments, if you think that the current implementation is more readable or has advantages, you can also leave it as it is :)

nathanaelbosch (approved):

Issue #157, opened 2020-08-19
Refactor `probnum.filtsmooth` to resolve the pylint messages


nathanaelbosch:
At the time of writing, `probnum.filtsmooth` triggers the following `pylint` messages:
- `duplicate-code`
- `protected-access`
- `no-self-use`
- `too-many-locals`
- `arguments-differ`
- `too-many-arguments`
- `unused-argument`

__Goal:__
Our long-term goal is to enable all of these checks for all parts of the project. Therefore, we should continuously work on fixing these. Currently, pylint is set up to have different checks for different modules, such that we can work on these problems individually and progress more quickly :)

__To work on this issue__ you can check for these messages individually by calling:
```
pylint src/probnum/filtsmooth --disable=all --enable=<message>
```
Then, after fixing all errors/warnings, you should remove the exception from the corresponding line in `./tox.ini`, such that this message will be tracked in the future.

pnkraemer:
Thanks for that detailled summary! Some of those I have fixed already, some others are much harder to fix or perhaps should be ignored. I will make a WIP PR and give details below.

pnkraemer:
How would you deal with the `unused-argument` for abstract base classes. My idea would be to disable this warning in each respective line, i.e. 
```python
    @abstractmethod
    def update(self, time, randvar, data, **kwargs):
        """
        Update step of the Bayesian filter.

        Must be implemented by subclasses.
        """
        raise NotImplementedError
``` 
becomes 
```python
    @abstractmethod
    def update(
            self,
            time,  # pylint: disable=unused-argument
            randvar,  # pylint: disable=unused-argument
            data,  # pylint: disable=unused-argument
            **kwargs  # pylint: disable=unused-argument
    ):
        """
        Update step of the Bayesian filter.

        Must be implemented by subclasses.
        """
        raise NotImplementedError
```

This is tedious, and potentially requires lots of C&P of `# pylint: disable=unused-argument`, but disabling this warning in principle seems too sledgehammerish to me.

nathanaelbosch:
> How would you deal with the `unused-argument` for abstract base classes. 
The issue I see here is rather that `BayesFiltSmooth.update` is currently *not* an abstract method, since if it had the `@abstractmethod` decorator it would not trigger the "unused-argument" warning.

pnkraemer:
[...]

PR #169, opened 2020-08-24
Pylint filtsmooth [WIP?]

pnkraemer:
Response to #157. These are the pylint error messages that seemed easy to fix. Most notable changes are
* there is a utils.py module which contains the checks for discrete-discrete vs. continuous-discrete
* a function linear_discrete_update is in gaussfiltsmooth which does the discrete kalman update. It is used for the kalman filter and the extended kalman filter, the latter passes the linearisation. I find it a bit cumbersome to fit the ukf update into this...


src/probnum/filtsmooth/gaussfiltsmooth/_utils.py
```diff
@@ -0,0 +1,24 @@
...
+    ContinuousModel,
+    DiscreteModel,
+    LinearSDEModel,
+    DiscreteGaussianLinearModel,
+)
+
+
+def is_cont_disc(dynamod, measmod):
+    """Checks whether the state space model is continuous-discrete."""
+    dyna_is_cont = issubclass(type(dynamod), ContinuousModel)
```
nathanaelbosch, line 15: `isinstance(dynamod, ContinuousModel)` should also work and might be more concise.

nathanaelbosch (changes requested):
Do these changes resolve a whole class or pylint messages? If yes, please update the corresponding line in `tox.ini` to track those in the future. Otherwise I'm good with these changes, thanks a lot for the effort! :)

src/probnum/filtsmooth/gaussfiltsmooth/_utils.py
pnkraemer: Ad hoc, I dont think it would work. LinearSDE models for instance are subclasses of ContinuousModel.

nathanaelbosch: I'm not sure I understand your point. Do you have an example of something that `issubclass(type(variable), Class)` does that `isinstance(variable, Class)` does not? To my understanding both are equivalent, but I am definitely not sure about this :sweat_smile:

pnkraemer: I was convinced that it is different, checked, and it turns out you are right.

pnkraemer:
travis fails with error 
```
ERROR:   benchmarks: InvocationError for command /home/travis/build/probabilistic-numerics/probnum/.tox/benchmarks/bin/python -m pip install --exists-action w .tox/.tmp/package/1/probnum-0.0.1.zip (exited with code 2)

The command "tox" exited with 1.
```
I don't know how to fix this. @nathanaelbosch @JonathanWenger @marvinpfoertner would one of you mind taking a look?

marvinpfoertner:
> travis fails with error
> 
> ```
> ERROR:   benchmarks: InvocationError for command /home/travis/build/probabilistic-numerics/probnum/.tox/benchmarks/bin/python -m pip install --exists-action w .tox/.tmp/package/1/probnum-0.0.1.zip (exited with code 2)
> 
> The command "tox" exited with 1.
> ```
> 
> I don't know how to fix this. @nathanaelbosch @JonathanWenger @marvinpfoertner would one of you mind taking a look?

Looks like a network error, maybe rerunning the pipeline will fix it.

pnkraemer:
Now the code coverage check fails, I have no idea why. I did not touch any coverage-related things as far as I know...

nathanaelbosch:
> Now the code coverage check fails, I have no idea why. I did not touch any coverage-related things as far as I know...

So far we often merged even withouth the codecov test check passing, so I think it should be fine.

pnkraemer:
okay, if there is nothing else that needs to be changed maybe give it your approval? :)

nathanaelbosch (approved):

PR #176, opened 2020-08-26
Global pylint config 

### LINKED ISSUE #173: Create a project-wide pylint configuration
Define the union of all pylint exceptions in pyproject.toml. Once an exception is fixed in all modules it can be removed from the tox configuration and the global pylint config.

nathanaelbosch:
Closes #173. 

I added disabled messages to the global pylint config in `./pyproject.toml` to make sure that one can use pylint from the command line or inside a text editor (which is strongly recommended!), without getting many many false positivs (ie. errors, that are not actually expected to get fixed). To do so, I added the union of all disabled messages from `./tox.ini` to the pylint config.

I also wrote a small, very *hacky* script to make sure that both of these files, `./tox.ini` and `./pyproject.toml`, stay synchronized, and I added the script to our CI workflow.


.github/workflows/pylint_check.py
```diff
@@ -0,0 +1,79 @@
...
+
+# Check correctness and raise errors
+try:
+    in_pyproject_butnot_tox = pyproject_disables.difference(tox_disables).difference(
+        GLOBAL_DISABLES
+    )
+    assert not in_pyproject_butnot_tox
+except AssertionError:
+    raise Exception(
+        f"""
```
nathanaelbosch, line 41: I think the most important point for the review could be how informative these error messages are. Please make sure that you understand them and that you have an idea of what you would do if you ever get this error.

marvinpfoertner (approved):
Looks good to me! I don't think the error messages could be any clearer

PR #178, opened 2020-08-26
Resolve pylint benchmarks

pnkraemer:
I did what I claimed I will never do and touch linear solver code... Now the benchmark directory is pylint-clean. Mostly, this amounted to adding some pylint-discard messages but I also changed some code. Nothing major IMO. Everything still benchmarks and I hope the changes are not too opinionated. If I misunderstood something, @JonathanWenger please judge.


benchmarks/benchmark_utils.py
```diff
@@ -0,0 +1,14 @@
+"""Reocurring constants and methods for benchmarking functions."""
+
+import numpy as np
+
+
+RANDOM_5x5_SPD_MATRIX = np.array(
```
JonathanWenger, line 6: Good idea to separate this out. We might even want to have a folder with matrices etc. similar as to the tests where we collect these benchmark / test problems. For now just a naming correction: `SPD_MATRIX_5x5` is more descriptive since it is not a random matrix.

benchmarks/distribution.py
```diff
@@ -7,8 +7,10 @@
 import probnum.prob as prob
 import probnum.linalg.linops as linops
 
+from benchmarks.benchmark_utils import RANDOM_5x5_SPD_MATRIX
+
 # Module level variables
-distribution_names = [
+DISTRIBUTION_NAMES = [
```
JonathanWenger, line 13: Once you merge the most recent state of ProbNum, this should already be fixed, but these are `RANDVAR_NAMES` now with the refactoring done in #164 .

benchmarks/distribution.py
```diff
@@ -60,25 +54,28 @@ class Functions:
...
 
-    param_names = ["dist", "property"]
-    params = [distribution_names, ["pdf", "logpdf", "cdf", "logcdf"]]
+    param_names = ["dist", "method"]
+    params = [DISTRIBUTION_NAMES, ["pdf", "logpdf", "cdf", "logcdf"]]
 
-    def setup(self, dist, property):
+    def setup(self, dist, method):
+        # pylint: disable=missing-function-docstring,attribute-defined-outside-init
+        # pylint: disable=unused-argument
```
JonathanWenger, line 62:
```python
# pylint: disable=unused-argument
# pylint: disable=missing-function-docstring
```
will appear in almost every benchmark and should if possible be ignored for all benchmarks.

JonathanWenger (changes requested):

benchmarks/benchmark_utils.py
pnkraemer: yes, I agree that putting those things into a separate folder is a good longterm goal. Regarding the naming: sounds sensible.

benchmarks/distribution.py
pnkraemer: The docstring warning makes sense to be ignored globally IMO (the docstrings seem to be super repetitive anyway: "Time method xyz", "Benchmark memory usage of xyz". The `unused-argument` I am more hesitant about...

pnkraemer: I put the ignore for unused-argument back into the tox file. Any idea where we can best document that this is not one of the "we have not resolved this yet" ignores but a "final" ignore? I put a comment into the tox ini file, but if you have a better idea please share.

benchmarks/random_variables.py
```diff
@@ -59,24 +53,28 @@ class Functions:
     Benchmark various functions of distributions.
```
JonathanWenger, line 53: Quick suggestion, while you're working on this. This docstring should say:
```python
Benchmark various functions of random variables.
```

JonathanWenger (approved):

benchmarks/distribution.py
JonathanWenger: Maybe @nathanaelbosch knows a way to ignore specific warnings in a certain folder, however judging from [StackOverflow](https://stackoverflow.com/questions/36182847/pylint-disable-specific-warnings-for-specific-folder) one can only achieve this by having a `.pylintrc` file in `benchmarks` and then having two `pylint` statements (which we already have anyway).

nathanaelbosch: @marvinpfoertner also looked into it and, if I'm not mistaken, came to the conclusion that it's unfortunately not possible to have configs on a per-directory basis.

In general I'm definitely open for ideas to move these exceptions out of the `tox.ini`, since the current way does not feel ideal.

Issue #175, opened 2020-08-25
PN method tutorial and prototypical Implementation


JonathanWenger:
Create a PN tutorial which step-by-step explains with an example of what a PN method is, what it consists of and the type of problem it solves. Secondarily, reference this method in a detailed sample explanation with type hints, observations/actions, policy, stopping criteria, class structure, convenience function signature extra.

**Detailed Example:**
- noisy quadratic optimization of a 1D example: (a+e_a)x^2+(b+e_b)x+c+e_c = f(x) = \hat{y}
- results in a 1D example of the PLS
- plot the quadratic and its quantiles in a plot to illustrate the problem
- show iterations of the solver (1D belief over x-axis with true quadratic and sample quadratic)

**Development Tutorial:**
- convenience function with type hinted signature and method selection, callbacks, etc.
- Actual class with implementation of PN Method in compositional pattern
- Step-by-step implementation of class components (maybe referenced from detailed example) which outlines policy, stopping criteria, iteration (__iter__?) etc.

pnkraemer:
Controversial take: wouldn't it perhaps be easier to 1) showcase and 2) understand if this prototypical implementation is explained with actual code examples, e.g. linear solver of ODE solver? I don't have a strong opinion on this, but it might be worth thinking about?

JonathanWenger:
It will feature code examples, but they are not production code - at least not in the examples. It's much harder to read code of the actual repository for the average user than a simple (potentially inefficient) example. The development tutorial will still consider the simple 1D example (which essentially is a version of the PLS), but will serve as a template for developers, hence feature production style code.

pnkraemer:
hmmm sounds reasonable. I guess I will wait for the result and then think about it :)

PR #181, opened 2020-08-26
Documentation: PN Method example added and development tutorial updated to compositional pattern

### LINKED ISSUE #83: Use compositionality instead of subclassing when defining Probabilistic numerical methods

### LINKED ISSUE #105: Handle markdown conversion to restructuredText via m2r2
Currently we enforce the documentation build with a Sphinx version <=2.4.4. This is only necessary since the package `m2r` which converts markdown to rst has not been updated to work with Sphinx >=3.0. However there is a fork called `m2r2` which fixes this.

By simply replacing `m2r` with `m2r2` in `conf.py` and `./docs/requirements.txt` we could allow the newest version of Sphinx.

### LINKED ISSUE #182: Add explanation to developer docs about type usage
Currently we define argtypes and internal types in two different locations in the package. Write a concise guide for developers on how these types are to be used. 

This also relates to #181.

JonathanWenger:
This PR updates the documentation in multiple ways, primarily focussed on examples and tutorials.

**Docs Build Changes**
- [x] Sphinx pinned to version >=3.0
- [x] plugin for .`md.` to `.rst` conversion changed to `m2r2` from `m2r` to enable Sphinx 3.0 compatibility
- [x] Plugin `sphinx-autodoc-typehints` added to automatically add typehints to docstrings

Closes #105 

**Tutorials**
- [x] Jupyter notebook explaining a very basic 1D PN method to a user unfamiliar with PN
- [x] Sphinx Galleries for tutorial notebooks (see also https://nbsphinx.readthedocs.io/en/0.7.1/subdir/gallery.html)
Screenshot from 2020-08-29 17-19-57

**Development Guides**
- [x] Development Guide implementing a simple PN method in the desired code pattern, which can serve as a template for new PN methods 
    - Explanation on API and argument types. Closes #182. 
- [x] Developer guides are now Jupyter notebooks allowing the use of magic functions (e.g. code pasting from `.py` files via `%load`) and code formatting with [jupyter black](https://github.com/drillan/jupyter-black)
- [x] Developer guide on writing unit tests
- [x] Updated developer guide on writing documentation (with notes on type hints and shapes of arguments)

Closes #83 and #175

**Other**
- [x] Adjusted default plotting style in `probnum.mplstyle` to be more suitable to the range of plots in all notebooks
   - changed default plot size
- [x] Adds an "Advanced Developer Setup" section to the contribution guide which explains how to set up pre-commit hooks
- [x] Added more detailed attribution with images to "ProbNum" contributors section in docs

Note: 
- _You can view the built documentation by going to the ["Details" page](https://probnum--181.org.readthedocs.build/en/181/) of the `docs/readthedocs.org:probnum` check, which builds a version of the documentation corresponding to this PR._
- _You can also review the changes to the notebooks at https://app.reviewnb.com/probabilistic-numerics/probnum/pull/181/files/, which provides file differences for notebooks._


JonathanWenger:
First of all, excuse the huge PR. I'll make sure to not do this again. I would appreciate feedback on the following points from you. Easiest for feedback is to just look at the [ReadTheDocs Build](https://readthedocs.org/projects/probnum/builds/11987722/).

**PN Method Tutorial** (@nathanaelbosch @pnkraemer @marvinpfoertner):
- Is this simple enough for somebody unfamiliar with PN?
- Anything I missed which would improve presentation?
- Are you "sold" on PN after this tutorial coming from ML / numerics / computer science?

**PN Method Development Guide**:
- Does this pattern fit with `probnum.diffeq` as well? @nathanaelbosch @pnkraemer 
- Am I using custom types as intended and consistently? @marvinpfoertner 
- Any code improvements you would make? @marvinpfoertner 
- Is this simple example extensive / general enough to extrapolate to a new PN Method (e.g. quadrature?) @alpiges 
- is it overkill to have a `callback` function and an iterator in `ProbabilisticQuadraticOptimizer`? They offer essentially the same functionality after all, except only `callback` is part of the interface. @marvinpfoertner 

Finally, if you have an idea on how to use Jupyter notebook's functionality (or magic functions) to load code from file to show it, but do this every time you execute a cell, as opposed to %load which pastes the code once and then executes it on repeated cell executions, please let me know.

Thanks for the feedback!

pnkraemer:
Sorry that the feedback is rather sparse, but either way, here are my 2 cents:

* I like this, thanks for the good work!
* Seems like half the notebooks is plotting code, but I dont know either how to get rid of it. 
* The pattern seems to fit ODEs and so on as well. One thing that might make it even easier to understand is to not use terms like 'policy' and 'belief update', which one might have to read twice to be able to understand what is going on, but easier words like 'action rule' and 'training/computing/...'. Also, it might be sensible to mention in the advantages of PN that some things are better understood than others. The re-interpretation is no issue, but things with similar models as priors for others is unclear (yet), at least in the ODE case (and as far as I know, for BQ and other problems, too). Hyperparameter optimisation is a nice drop though, even if this probably needs some more work (nathanael had some great results in the past!). Maybe this is a nice combo with the 'reinterpretation' thing, too, because suddenly you are indeed able to optimise hyperparameters for some old and well-known algos.
* I believe that we should at some point invest the time and teach each other how the respective other PN methods work. I believe it will help us all, especially for writing this code and I must admit that I do not know how the PLS works.

alpiges:
Here's my feedback on "Implementing a Probabilistic Numerical Method", finally!

It looks great, and really seems helpful as guidance for implementation!

The suggested structure seems general enough to extrapolate to new PNMs, including quadrature. After all, it leaves quite some flexibility for their implementation behind the user interface.

Here are a few comments about the tutorial itself:

- The type hints section appears too prominent to me. It's a bit distracting at that point and not essential to understanding what the structure of a PNM should look like.

- To me it would be more intuitive to see the class `ProbabilisticQuadraticOptimizer` before the individual components (policy, stopping criteria, etc.). I would prefer to first get an overview and then dive into the details on how to implement the required methods.

- That last point might also resolve the previous one by moving the "Type Hints" section after the section "Class `ProbabilisticQuadraticOptimizer`". That would help first explain the overall structure and then the implementational details.

About the custom types: Is it a thing to define all sorts of types to specify how methods should look like? Why is that not done via classes? To me that looks like artificial inheritance (like having a base class `StoppingCriterion` that other stopping criteria inherit from). I haven't seen this before and am mildly confused :)


Minor notes / typos:

- Intro: leave out all the "in depth", "in detail", that's pretty evident and does not carry extra information :)

- mock-up -> mock up (it's a verb here)

- "A specific PN method so solve this problem is built up from its components and explained in detail in the tutorial on PN methods." what does "its components" refer to?
How about: "This particular problem has been used to showcase _what_ a PN method _is_ in the tutorial on PN methods -- now we will turn towards _how_ to implement such a method in ProbNum.
also "so solve" -> "to solve"

- Why loading the probnum logo?

- "The interface automatically detects that its faced" -> it is

marvinpfoertner (approved):

PR #191, opened 2020-08-28
Unittest prior

pnkraemer:
This is the unit test I was talking about earlier. I would've liked to include the vectorisation thing for IOUP but as I explained earlier, I deemed it not worth it :( 

The added unit test is everything productive that is left from this effort :)


tests/test_diffeq/test_odefiltsmooth/test_prior.py
```diff
@@ -53,34 +53,68 @@
...
+        The result of chapmankolmogorov() should be identical to the matrix fraction decomposition technique
+        implemented in LinearSDE, just faster.
+        """
+        mean, cov = np.ones(self.prior.ndim), np.eye(self.prior.ndim)
+        initrv = RandomVariable(distribution=Normal(mean, cov))
+        cke_super, __ = super(
+            type(self.prior), self.prior
+        ).chapmankolmogorov(  # pylint: disable=bad-super-call
+            0.0, STEP, STEP, initrv
+        )
```
nathanaelbosch, line 78: I would have probably put the pylint ignore here, at the end of the full expression. Maybe that fixes it?
Alternatively, just write it into the line right before (74 then), I don't think it should be a problem if the disable is active for the whole funciton, and it might be more readable.

nathanaelbosch (commented):
There seems to be a conflict that needs to be resolved.

Apart from that it looks good for me :)

nathanaelbosch (approved):

PR #192, opened 2020-08-28
Adjusted tox config such that asv fails on benchmark error

JonathanWenger:
With the current CI setup while benchmarks were run in `dev` mode with airspeed velocity, asv did not actually return non-zero exit code on failure. This is only possible with the newest version on Github and its `--strict` flag. This PR now ensures that the tox environment for the benchmarks actually fails if there is a breaking code change in the benchmarks.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -40,11 +40,11 @@ commands = black .
 [testenv:benchmarks]
 description = Dry run the benchmarks to check for errors
 basepython = python3
-deps = asv
+deps = git+https://github.com/airspeed-velocity/asv.git@master#egg=asv
 changedir = benchmarks
 commands =
 		 asv machine --yes
-		 asv dev
+		 asv run -e --quick --strict --show-stderr --python=same
 
 
 
--- a/tox.ini
+++ b/tox.ini
@@ -40,11 +40,11 @@ commands = black .
 [testenv:benchmarks]
 description = Dry run the benchmarks to check for errors
 basepython = python3
-deps = git+https://github.com/airspeed-velocity/asv.git@master#egg=asv
+deps = git+https://github.com/airspeed-velocity/asv.git@a131ca742daf53c4521fbfefeb53507b5c140c84#egg=asv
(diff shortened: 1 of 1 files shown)


tox.ini
```diff
@@ -40,11 +40,11 @@ commands = black .
 [testenv:benchmarks]
 description = Dry run the benchmarks to check for errors
 basepython = python3
-deps = asv
+deps = git+https://github.com/airspeed-velocity/asv.git@master#egg=asv
```
nathanaelbosch, line 43: I could imagine that depening on master could lead to some errors at some later point. One alternative would be to specify the exact commit that we want to depend on: 
```
deps = git+https://github.com/airspeed-velocity/asv.git@a131ca742daf53c4521fbfefeb53507b5c140c84#egg=asv
```
Tested locally this seems to work. On the other hand, I also don't think it's reall big deal to depend on master, even if at some point this CI check might fail, we handled similar problems fine :)

tox.ini
```diff
@@ -40,11 +40,11 @@ commands = black .
...
 changedir = benchmarks
 commands =
 		 asv machine --yes
-		 asv dev
+		 asv run -e --quick --strict --show-stderr --python=same
```
nathanaelbosch, line 47: Is `run --quick` basically the new `dev`? So a "faster" test-run of the benchmarks?

tox.ini
JonathanWenger: Good call. Once the new release is out on PyPi we can pin this to a version.

tox.ini
JonathanWenger: `asv dev` is equivalent to `asv run -e --quick --show-stderr --python=same` by additionally adding `--dry-run` and `--strict` we can a) avoid saving any results to disk and b) ensure a non-zero return value is given on failure.

nathanaelbosch: Thanks, good to know :)

nathanaelbosch (approved):

PR #201, opened 2020-08-31
Renamed probnum.typing and bumped version

JonathanWenger:
In preparation for a new release this PR bumps the version of the package to 0.1.0. In accordance with [semantic versioning](https://semver.org/) breaking changes to the API can still be made at any time:
> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

This PR also renames the `probnum.typing` module to `probnum._typing` in accordance with my understanding of its current use. This also resolves potential name conflicts with `typing.py` as observed locally in PyCharm.


marvinpfoertner (changes requested):
IMHO probnum's type aliases should be public so that a user can reuse them. @JonathanWenger How about we rename `probnum.typing` to `probnum.types` instead?

marvinpfoertner (approved):
We should maybe update the PR title before merging though :)

JonathanWenger:
> We should maybe update the PR title before merging though :)

Way ahead of you ;)

PR #204, opened 2020-09-02
Resolve utils pylint

pnkraemer:
the `utils` module is pylint-clean now. The only nontrivial changes were:
* I renamed some variables that were not pep8 conform (`x` is now `seed`/`shape`, `rv` is now `randvar` according to the styleguide)
* I changed the no-else-return that pylint complains about. I know that this might be controversial but IMO, if we stick to pylint as a code checker we pick pylint's view in controversial questions ;) 

I believe that @marvinpfoertner is the author of many of these functions (at least of the `argutils` module)? If not, I will request another review.

### DIFF
--- a/src/probnum/utils/argutils.py
+++ b/src/probnum/utils/argutils.py
@@ -16,10 +16,15 @@ __all__ = ["as_shape", "as_random_state", "as_numpy_scalar"]
 
 
 def as_random_state(x: RandomStateArgType) -> RandomStateType:
+    """
+    Transform a variable or RandomStateArgType into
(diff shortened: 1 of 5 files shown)


pnkraemer:
Interestingly, resolving the `no-else-return` decreases the code coverage...

pnkraemer:
In a conversation with @nathanaelbosch, I just found out that some of the things that I fixed were not meant to be fixed (`invalid-name` and `no-else-return`). I seems like I missed that decision. If, in light of this, I broke something please complain.

PR #206, opened 2020-09-03
Correct computation of Kronecker determinant

### LINKED ISSUE #205: Determinant of Kronecker products is computed incorrectly
The determinant of a Kronecker product A(x)B  of matrices A (m x m) and B (n x n) is incorrectly computed as det(A (x) B) = det(A)^m * det(A)^n, whereas the correct form is det(A (x) B) = det(A)^n * det(A)^m (see e.g. https://en.wikipedia.org/wiki/Kronecker_product#Relations_to_other_matrix_operations). This requires corrections in the functions `Kronecker.det()` and `Kronecker.logabsdet()` and a test should be written comparing against a numerically computed determinant on the full matrix.

JonathanWenger:
This PR fixes the computation of the determinant and log-determinant of the Kronecker product and adds some tests.

It also renames the `probnum.types` to `probnum.type` to avoid naming conflicts with [types.py](https://docs.python.org/2/library/types.html) and adds a note about recreating virtual environments with tox, which can be a problem if one switches between different branches or modifies the build configuration.

Closes #205.


CONTRIBUTING.md
```diff
@@ -53,6 +53,9 @@ documentation locally by simply calling
 ```bash
 tox
 ```
+Note, to reduce runtime tox caches and reuses the virtual environment it creates the first time you run the command. If 
+you are frequently switching between branches or adjusting the build configuration make sure to force recreation of the 
+virtual environment via ``tox -r``, if you experience unexpected tox failures.
```
nathanaelbosch, line 58: Useful addition, thanks :)

nathanaelbosch (approved):
I'm not sure I completely understand the naming conflict with `types`, is the following the example you had in mind and that you want to avoid?
```python
import types
from probnum import types
```
Ignoring all potential conflicts I'd say that `types` is the better naming compared to `type`, but if you see potential conflicts then I support you change of course :) I did not `probnum.types` myself yet so I don't really have a good opinion on it.

From what I understood the Kronecker bugfix and the tests seem good :) I have to admit however that I don't have a 100% understanding of the changes (and their meaning) in `test_linearoperators.py` so let me know if there are some specific parts for which you are unsure and I could look into it.

Thanks for the PR! :)

JonathanWenger:
> I'm not sure I completely understand the naming conflict with `types`, is the following the example you had in mind and that you want to avoid?
> 
> ```python
> import types
> from probnum import types
> ```
> 
> Ignoring all potential conflicts I'd say that `types` is the better naming compared to `type`, but if you see potential conflicts then I support you change of course :) I did not `probnum.types` myself yet so I don't really have a good opinion on it.

I had some trouble with PyCharm, where PyCharm would not run tests since it tried to load `probnum.typing` / `probnum.types` instead of the native `typing` and `types`. This might admittedly be operator error or a problem with my config, but i think this could be a somewhat common error for new developers (using PyCharm).

nathanaelbosch:
> PyCharm would not run tests since it tried to load `probnum.typing` / `probnum.types` instead of the native `typing` and `types`

Thanks for the explanation! I also just checked for lines where "import" and "types" appears since I thought that it might be related to relative imports or the file structure, but actually it is always an explicit import of the type `from probnum.types import ...`, which seems like the right thing to do.

I agree that even if it might be related to the specific PyCharm setup it is something that other people could run into, so it makes sense to change it :)

PR #214, opened 2020-09-16
Travis Migration to travis-ci.com

JonathanWenger:
Travis is migrating all its repositories from https://travis-ci.org to https://travis-ci.com by Dec 31st, 2020 (see also: https://docs.travis-ci.com/user/migrate/open-source-repository-migration)

> As you’ll no doubt have noticed, Travis CI has, for some time, been split between travis-ci.com and travis-ci.org. This was initially planned to enable open-source and commercial/private projects to be distinctly separate, but back in May 2018, we announced that open source builders would have the opportunity to build and test their projects under travis-ci.com, meaning your public and private repositories could be managed under one domain with a consistent interface and build environment.

In order to facilitate this change I performed the migration and updated any links in the repository. Builds will from now on be available at https://travis-ci.com/probabilistic-numerics/probnum. Build history can still be found at the old address and should be migrated eventually by Travis.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -17,8 +17,8 @@ for detailed instructions.
 
 ## Getting Started
 
-Begin by forking the repository on GitHub and cloning your fork to a local machine. Next, create a new branch in your 
-forked repository describing the feature you would like to implement. You can now get started writing code in your new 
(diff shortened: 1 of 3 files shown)


pnkraemer:
As far as I see, nothing to complain about. Thanks for doing that!

pnkraemer (approved):

PR #217, opened 2020-09-18
Samples from an ODE Posterior

pnkraemer:
Now it is possible to sample from an `ODESolution` or `KalmanPosterior` object via `posterior.sample(size=123456789)` (pls dont do that many samples...). We can now create plots like this:

[samples.pdf](https://github.com/probabilistic-numerics/probnum/files/5245452/samples.pdf)

# Notable code changes:

* sample() routines in KalmanPosterior and in ODESolution. For now, it is only possible to compute joint samples over the grid points. Combination with __call__ is not implemented yet, but the scaffolding is there and there are already unit tests (which asserts that corresponding exceptions are thrown if this is attempted). Beware, though, because computing multiple samples does not compute them at once, but in a loop.
* This is mostly a smoothing routine, so I extracted the required parts from `.smooth()` into a `smooth_list` function.
* Decision whether to smooth or not is now made in the creation of a `KalmanPosterior`. By default, the object returned by `.filter()` has `with_smoothing=False` and the one returned by `.smooth()` has `True`. For ODE solutions, this information is stored in the solver (which passes it on nicely)

# Other changes:

* `_RandomVariableList` still used `mean()` and `cov()`, instead of the properties. This is changed. 
* I used something like a chi-squared test in the unit tests (not sure whether it is exactly the same; please correct me if I am wrong). I started a `statistics.py` module in `tests/testing`. We might want to add more of those at some point


pnkraemer:
I will try my best to allow sampling from arbitrary locations but this might turn out to be super tedious. Let's see... You can review this code already though, because it will likely remain untouched.

src/probnum/filtsmooth/filtsmoothposterior.py
```diff
@@ -32,3 +32,16 @@ def __len__(self):
...
         """Return the corresponding index/slice of the discrete-time solution"""
         raise NotImplementedError
+
+    def sample(self, locations=None, size=None):
+        """
+        Draw samples from the filtering/smoothing posterior.
+
+
+        If location is None, it returns N joint samples from the list of RVs;
+        if location is one of the locations of the posterior,
```
nathanaelbosch, line 42: There is some grammatical mistake in the sentence here, and I'm not 100% sure I understand what it tells me.

Maybe also add a `Parameters` section to describe `locations` and `size`.

src/probnum/filtsmooth/gaussfiltsmooth/gaussfiltsmooth.py
```diff
@@ -75,38 +77,51 @@ def filter(self, dataset, times, **kwargs):
...
+        return KalmanPosterior(
+            filter_posterior.locations, rv_list, self, with_smoothing=True
+        )
+
+    def smooth_list(self, rv_list, locations, final_rv=None):
+        """
+        Apply smoothing to a list of RVs with desired final random variable.
+
+        Specification of a final RV is useful to compute samples from a KalmanPosterior object,
+        because in this case, the final RV is a Dirac (over a sample from the final Normal RV)
```
nathanaelbosch, line 106: Oh so that's how sampling from a smoothed posterior works :D

src/probnum/filtsmooth/gaussfiltsmooth/gaussfiltsmooth.py
```diff
@@ -75,38 +77,51 @@ def filter(self, dataset, times, **kwargs):
...
+        and not a Normal RV.
```
nathanaelbosch, line 107: Oh and sorry to state that again but it could be a good idea to also format the docstring with `Parameters` and `Returns`, since the content matches that structure already anyways.

src/probnum/filtsmooth/gaussfiltsmooth/gaussfiltsmooth.py
```diff
@@ -75,38 +77,51 @@ def filter(self, dataset, times, **kwargs):
                 **kwargs
             )
             rvs.append(filtrv)
-        return KalmanPosterior(times, rvs, self)
+        return KalmanPosterior(times, rvs, self, with_smoothing=False)
 
-    def smooth(self, filter_posterior, **kwargs):
+    def smooth(self, filter_posterior):
```
nathanaelbosch, line 82: After seeing `smooth_list` below, what about naming this `smooth_posterior`? If you think it is clear enough that `smooth` operates on a posterior we can leave it as is of course :)

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -122,3 +121,31 @@ def __len__(self):
 
     def __getitem__(self, idx):
         return self.state_rvs[idx]
+
+    def sample(self, locations=None, size=()):
+        """
+        Draw samples from the filtering/smoothing posterior.
+
+
+        If location is None, it returns N joint samples from the list of RVs;
```
nathanaelbosch, line 130: If you end up changing something in the other docstring remember to also change it here.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -122,3 +121,31 @@ def __len__(self):
 
     def __getitem__(self, idx):
         return self.state_rvs[idx]
+
+    def sample(self, locations=None, size=()):
```
nathanaelbosch, line 125: Why is `size=()` the default, and not `size=None` or `size=1`?

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -122,3 +121,31 @@ def __len__(self):
...
+        """
+        Draw samples from the filtering/smoothing posterior.
+
+
+        If location is None, it returns N joint samples from the list of RVs;
+        if location is one of the locations of the posterior,
+        if not (i.e. if it is a scalar), it returns a sample from the specified location.
+
+        If size is empty, it is a single sample. If not, multiple samples at once.
```
nathanaelbosch, line 134: I think it could also be helpful to describe the output some more, especially the shape.

nathanaelbosch (commented):
Good work! That's a really nice functionality to have :) 

I commented on a few small things regarding documentation, but in general it looks good to me. Thanks also for fixing the RV-list methods to properties.

nathanaelbosch:
Oh and I didn't review the notebooks, so I'd just trust you on these. If you think I should have a look let me know and I'll do that.

pnkraemer:
Now it is possible to sample on arbitrary locations. To this end, just enter an array (or list or whatever) into `locations`, respectively `t`. 

Note though, that it will be only as fast as vectorisation inside `call` (currently, it loops over inputs).

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -45,30 +47,31 @@ def state_rvs(self):
...
-        smoothed : bool, optional
-            If ``True`` (default) perform smooth interpolation. If ``False`` perform a
-            prediction from the previous location, without smoothing.
 
         Returns
         -------
         :obj:`RandomVariable`
             Estimate of the states at time ``t``.
         """
+        if np.isscalar(t) is False:
```
nathanaelbosch, line 71: "pythonic" would be: `if not np.isscalar(t)` ;) I'm surprised pylint didn't compain (yet?)

src/probnum/filtsmooth/filtsmoothposterior.py
pnkraemer: Oh good point! I wrote those before I wrote the code, this is why it sounds more like a shopping list than a description. I will update them!

src/probnum/filtsmooth/gaussfiltsmooth/gaussfiltsmooth.py
pnkraemer: essentially, yes^^

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: True.

I guess we excluded so many pylint warnings globally that it doesnt complain about anything anymore... :D

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -122,3 +125,46 @@ def __len__(self):
...
+
+        If size is empty, it is a single sample. If not, multiple samples at once.
+        """
+        errormsg = "Sampling not implemented."
+
+        if locations is None:
+            locations = self.locations
+            random_vars = self.state_rvs
+        else:
+            random_vars = _RandomVariableList(
```
nathanaelbosch, line 146: There's no need to loop over the `locations` here I think, `__call__(iterable)` does essentially the same

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: because this is what we are doing in random variable inputs. Not sure why exactly, maybe `size=None` sends mixed signals (do you want to sample or not??).

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: True. (I wrote this before I decided to put the loop into __call__ )

src/probnum/filtsmooth/gaussfiltsmooth/gaussfiltsmooth.py
pnkraemer: To be honest, I am not sure... I like the `smooth()` without `_posterior` because I think it is a nice default. But I would be easily convinced, if you have a strong opinion...

nathanaelbosch: I don't have a strong opinion, just wanted to raise the question. Let's keep `smooth` the way it is :)

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: Ah I see, I didn't think of checking our other `sample` implementations. If it's done that way there, then it should be the same here, I agree :+1:

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: Well, I'd remove this one (so the one docstring from `FiltSmoothPosterior` is used). Alright?

nathanaelbosch: Yes, sounds good to me :+1:

pnkraemer:
Any idea why the docs build failed? :/

src/probnum/diffeq/odesolution.py
```diff
@@ -149,19 +154,22 @@ def __call__(self, t, smoothed=True):
...
-        out_rv = self._proj_normal_rv(out_rv, 0)
-        return out_rv
+        out_rv = self._kalman_posterior(t)
+        if np.isscalar(t):
+            out_rv = self._solver.undo_preconditioning(out_rv)
+            return self._proj_normal_rv(out_rv, 0)
+        out_rvs = _RandomVariableList(
+            [self._solver.undo_preconditioning(out_rv_) for out_rv_ in out_rv]
+        )
+        return _RandomVariableList(
```
nathanaelbosch, line 170: As I understand it you're creating a RVList, just to iterate over it again and to do `self._proj_normal_rv`. Maybe it would be better to do this in one single loop?
That is
```python
        out_rvs = _RandomVariableList(
            [self._proj_normal_rv(self._solver.undo_preconditioning(out_rv_), 0) 
             for out_rv_ in out_rv]
        )
        return out_rvs
```
I guess "better" is the wrong word, the readability can be criticised (but I guess it would also look different through proper code formatting with Black), but it also saves an additional loop.

Even if you decide against it, there is no need for the first RVList and you could just have a python list.

nathanaelbosch: An alternative to this that comes to mind would be that `undo_preconditioning` and `_proj_normal_rv` could operate on both randomvariables and RVLists. Though I'm also not sure if that makes code cleaner since it makes things less explicit.

src/probnum/diffeq/odesolution.py
```diff
@@ -181,3 +189,25 @@ def __getitem__(self, idx):
...
+        size = utils.as_shape(size)
+
+        # implement only single samples, rest via recursion
+        if size != ():
+            return np.array([self.sample(t=t, size=size[1:]) for _ in range(size[0])])
+
+        samples = self._kalman_posterior.sample(locations=t, size=size)
+        return self._undo_preconditioning_and_project_samples(samples)
+
+    def _undo_preconditioning_and_project_samples(self, samples):
```
nathanaelbosch, line 205: This kind of relates to my comment above, so just a reminder to adapt this part depending on what you end up changing above.

nathanaelbosch: Maybe you could also have a `_undo_preconditioning_and_project_rv`, and just call
```python
_undo_preconditioning_and_project_rv([asrandvar(s) for s in samples]).mean
```
Again, no idea what of all this would be best

src/probnum/filtsmooth/filtsmoothposterior.py
```diff
@@ -32,3 +32,31 @@ def __len__(self):
     def __getitem__(self, idx):
         """Return the corresponding index/slice of the discrete-time solution"""
         raise NotImplementedError
+
+    def sample(self, locations=None, size=()):
+        """
+        Draw samples from the filtering/smoothing posterior.
+
+        If nothing is specified, a single sample is drawn (supported on self.locations).
```
nathanaelbosch, line 40: Really nice docstring! Thanks for the clarification :)

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -122,3 +126,39 @@ def __len__(self):
...
+
+        if size == ():
+            return self._single_sample_path(
+                locations=locations, random_vars=random_vars
+            )
+
+        return np.array(
+            [self.sample(locations=locations, size=size[1:]) for _ in range(size[0])]
+        )
+        # if np.isscalar(size):
```
nathanaelbosch, line 148: Should this comment stay?

tests/test_diffeq/test_odesolution.py
```diff
@@ -107,3 +107,79 @@ class TestODESolutionAdaptive(TestODESolution):
...
+
+    def test_output_shape(self):
+        loc_inputs = [
+            None,
+            self.solution.t[[2, 3]],
+            np.arange(0.0, 0.5, 0.025),
+        ]
+        dim = (self.ivp.ndim,)
+        single_sample_shapes = [
+            (len(self.solution),) + dim,
```
nathanaelbosch, line 127: I find the following more readable
```python
single_sample_shapes = [
            (len(self.solution), self.ivp.ndim),
            (2, self.ivp.ndim),
            (len(loc_inputs[-1]), self.ivp.ndim),
        ]
```

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
```diff
@@ -76,4 +77,79 @@ def test_call_to_discrete(self):
...
+        self.posterior = self.method.filter(self.obs, self.tms)
+
+    def test_output_shape(self):
+        loc_inputs = [
+            None,
+            self.posterior.locations[[2, 3]],
+            np.arange(0.0, 0.5, 0.025),
+        ]
+        dim = (self.method.dynamod.ndim,)
+        single_sample_shapes = [
```
nathanaelbosch, line 96: Same as above. Ignore if we keep things the same there.

src/probnum/diffeq/odesolution.py
pnkraemer: You're right.

nathanaelbosch (approved):
Looks good to me :)

PR #221, opened 2020-10-01
Added isort to tox

### LINKED ISSUE #140: Use isort to sort the order of import statements
While discussing #108 (flake8/pylint) with @marvinpfoertner we also talked about using [isort](https://github.com/timothycrosley/isort) to sort the order of import statements. It seems easy to set up in most popular editors (e.g. VS Code and PyCharm), or it can be used with pre-commit. It makes code more readable. And even if someone is completely unaware of the tool, PR errors could be fixed by installing and running isort (`pip install isort`, and `isort`), or through tox (`tox -e isort`).

It should also be quick to add to probnum.

@JonathanWenger any thoughts?

oriash93:
Similar to Black, running `tox -e isort` will check for sorting imports and will show required changes.
Additionally, running `tox -e format` will apply those changes.

Finally, I sorted all the files in the repo.

resolves #140

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ jobs:
 install:
   - sudo apt-get install pandoc # pandoc for jupyter notebooks
   - sudo apt install graphviz # graphviz for class inheritance diagrams in docs
-  - pip install isort # isort for sorting imports
   - pip install tox
 script:
   - tox
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,7 @@ jobs:
 install:
   - sudo apt-get install pandoc # pandoc for jupyter notebooks
   - sudo apt install graphviz # graphviz for class inheritance diagrams in docs
+  - pip install isort # isort for sorting imports
   - pip install tox
 script:
   - tox
--- a/src/probnum/__init__.py
+++ b/src/probnum/__init__.py
@@ -24,17 +24,11 @@ diffeq
 """
 # pylint: disable=wrong-import-order
 
-from . import diffeq
-from . import filtsmooth
-from . import linalg
-from . import quad
-from . import random_variables
-from . import utils
-
-from .random_variables import asrandvar, RandomVariable
-
 # -*- coding: utf-8 -*-
-from pkg_resources import get_distribution, DistributionNotFound
+from pkg_resources import DistributionNotFound, get_distribution
+
+from . import diffeq, filtsmooth, linalg, quad, random_variables, utils
+from .random_variables import RandomVariable, asrandvar
 
 try:
     # Change here if project is renamed and does not equal the package name
--- a/src/probnum/__init__.py
+++ b/src/probnum/__init__.py
@@ -22,8 +22,6 @@ filtsmooth
 diffeq
     Probabilistic solvers for ordinary differential equations.
 """
-# pylint: disable=wrong-import-order
-
 # -*- coding: utf-8 -*-
 from pkg_resources import DistributionNotFound, get_distribution
 
--- a/src/probnum/random_variables/_dirac.py
+++ b/src/probnum/random_variables/_dirac.py
@@ -5,8 +5,12 @@ from typing import Callable, TypeVar
 import numpy as np
 
 from probnum import utils as _utils
-from probnum.type import ArrayLikeGetitemArgType  # Argument Types
-from probnum.type import RandomStateArgType, ShapeArgType, ShapeType
+from probnum.type import (
+    ArrayLikeGetitemArgType,
+    RandomStateArgType,
+    ShapeArgType,
+    ShapeType,
+)
 
 from . import _random_variable
 
--- a/src/probnum/random_variables/_dirac.py
+++ b/src/probnum/random_variables/_dirac.py
@@ -5,16 +5,10 @@ from typing import Callable, TypeVar
 import numpy as np
 
 from probnum import utils as _utils
-from probnum.type import (
-    ShapeType,
-    # Argument Types
-    ArrayLikeGetitemArgType,
-    RandomStateArgType,
-    ShapeArgType,
-)
+from probnum.type import (ArrayLikeGetitemArgType,  # Argument Types
+                          RandomStateArgType, ShapeArgType, ShapeType)
 
 from . import _random_variable
-
 
 _ValueType = TypeVar("ValueType")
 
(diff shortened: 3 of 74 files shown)


.travis.yml
```diff
@@ -23,6 +23,7 @@ jobs:
 install:
   - sudo apt-get install pandoc # pandoc for jupyter notebooks
   - sudo apt install graphviz # graphviz for class inheritance diagrams in docs
+  - pip install isort # isort for sorting imports
```
nathanaelbosch, line 26: This dependency should not be necessary. When running `tox -e isort` tox will create a virtual environment and use the dependencies which are listed in `./tox.ini`, which includes isort.

On another note: I think it would be useful to add the isort check to our CI workflow. Right now we are using Github-actions for the linting (Black and pylint), so it would be great if you could add isort to our linting workflow (`.github/workflows/linting.yml`).

Let us know if you have any questions!

src/probnum/random_variables/_dirac.py
```diff
@@ -5,17 +5,11 @@
...
 from probnum import utils as _utils
-from probnum.type import (
-    ShapeType,
-    # Argument Types
-    ArrayLikeGetitemArgType,
-    RandomStateArgType,
-    ShapeArgType,
-)
+from probnum.type import ArrayLikeGetitemArgType  # Argument Types
+from probnum.type import RandomStateArgType, ShapeArgType, ShapeType
```
nathanaelbosch, line 9: I have the impression that here the automatic formatting didn't work perfectly, since the way I interpret this part of the code is that the `# Argument Types` comment refers to line 9, not 8. Could you change it in a way to keep the original structure while being isort compliant?

src/probnum/random_variables/_normal.py
```diff
@@ -8,14 +8,8 @@
...
-from probnum.type import (
-    ShapeType,
-    # Argument Types
-    ArrayLikeGetitemArgType,
-    FloatArgType,
-    RandomStateArgType,
-    ShapeArgType,
-)
+from probnum.type import ArrayLikeGetitemArgType  # Argument Types
+from probnum.type import FloatArgType, RandomStateArgType, ShapeArgType, ShapeType
```
nathanaelbosch, line 12: Same as in the previous comment. Just replicate whatever you end up doing in `src/probnum/random_variables/_dirac.py`.

src/probnum/random_variables/_random_variable.py
```diff
@@ -10,15 +10,14 @@
 import numpy as np
 
 from probnum import utils as _utils
+from probnum.type import ArrayLikeGetitemArgType  # Argument Types
```
nathanaelbosch, line 13: Same as in the previous comment. Just replicate whatever you end up doing in src/probnum/random_variables/_dirac.py.

nathanaelbosch (changes requested):
Thanks a lot for your initiative and your effort! I added some remarks, the biggest of which is the request to add isort to our github actions workflow to ensure that future work will always follow the isort standard. Please let us know if you have any questions, and again thanks a lot!

src/probnum/random_variables/_dirac.py
oriash93: I am not sure it's possible to configure `isort` to work that way. There is one option that is somewhat related - see [here ](https://github.com/PyCQA/isort#auto-comment-import-sections)- but I think it's a bit overkill. 

Is it possible to just remove the comment?

.travis.yml
oriash93: I set it up, let me know if it's good.

nathanaelbosch: The github action looks good to me, thanks!

tox.ini
```diff
@@ -4,7 +4,7 @@
 # and then run "tox" from this directory.
 
 [tox]
-envlist = py3, docs, benchmarks, black, pylint
+envlist = py3, docs, benchmarks, black, isort, format, pylint
```
nathanaelbosch, line 7: Good idea for adding isort to the list of evs to run when calling `tox`.

However, I think we want to keep the default behaviour as just a check for correctness, which does not modify existing code, so I would prefer it if you could remove `format` from this list again.

src/probnum/random_variables/_dirac.py
nathanaelbosch: I see, I also didn't find an easy way to keep the structure with the comment as before. I think removing the comment is a good idea and leads to the cleanest solution.

nathanaelbosch (changes requested):

src/probnum/__init__.py
```diff
@@ -24,17 +24,11 @@
...
-from . import linalg
-from . import quad
-from . import random_variables
-from . import utils
-
-from .random_variables import asrandvar, RandomVariable
-
 # -*- coding: utf-8 -*-
-from pkg_resources import get_distribution, DistributionNotFound
+from pkg_resources import DistributionNotFound, get_distribution
```
nathanaelbosch, line 28: @JonathanWenger what do you think about the changes to this file? I don't really have an opinion here, but the structure looks quite a bit different. I also just saw that we explicitly disable pylint here. Should we do the same for isort, or does it look fine the way it was formatted?

All other changes in this PR did not seem controversial to me, but if you have the time it could also be beneficial if you could skim through the changes to double-check.

oriash93: @nathanaelbosch Does anyone know why pylint was disabled here? We can make isort ignore this as well (comments, flags, etc.)

JonathanWenger: The changes are fine! Pylint was disabled since package imports were (unnecessarily) before third party imports. You can simply remove `# pylint: disable=wrong-import-order` with the changes you made. Following that we can merge this from my side.

nathanaelbosch (approved):
Thanks for updating the PR! Everything looks good to me.

EDIT: I wrote this before realizing that the pylint CI check fails. I requested one more change below that should fix it.

tests/test_filtsmooth/test_statespace/test_continuous/test_continuousmodel.py
```diff
@@ -1,9 +1,10 @@
 import unittest
 
 import numpy as np
-import probnum.random_variables as rvs
 
+import probnum.random_variables as rvs
 from probnum.filtsmooth.statespace.continuous import continuousmodel
+from probnum.random_variables import Dirac, Normal
```
nathanaelbosch, line 7: This line seems to be responsible for the pylint check to fail, probably due to the merge of the updated master branch. After removing it the CI should be satisfied and we should be able to merge the PR :)

nathanaelbosch (changes requested):

nathanaelbosch (approved):

nathanaelbosch:
Thank you once more for your contribution @oriash93 ! :)

PR #222, opened 2020-10-07
Pre commit additions

JonathanWenger:
Some useful additions to the default `pre-commit` configuration. In particular `isort` is now added as well in response to #221.


nathanaelbosch (approved):
I'm not 100% sure what exactly the added functionality does, but guessing from the config files I don't see any potential for problems with it so for me the PR is good to go.

If you think I should try it out locally in my machine before we merge let me know!

PR #223, opened 2020-10-07
Package restructuring and naming

JonathanWenger:
This PR makes the overall structure and naming of subpackages and modules more consistent and in line with the style guide. In particular `linops` is moved to the top level in the namespace. Also the API documentation structure is updated to match the actual structure of the package.

- [x] Refactored the linear operator subpackage out of `linalg` and pulled it up to `probnum`. 
- [x] more consistent naming of modules
- [x] improved structure for API documentation

Notebooks are not updated in this PR, but will be later in particular after major changes as part of #175 are merged.

### DIFF
--- a/benchmarks/random_variables.py
+++ b/benchmarks/random_variables.py
@@ -2,9 +2,9 @@
 Benchmarks for random variables.
 """
 
-import probnum.linear_operators as linear_operators
 import numpy as np
 
+import probnum.linear_operators as linear_operators
 from benchmarks.benchmark_utils import SPD_MATRIX_5x5
 from probnum import random_variables as rvs
 
--- a/benchmarks/random_variables.py
+++ b/benchmarks/random_variables.py
@@ -2,9 +2,9 @@
 Benchmarks for random variables.
 """
 
+import probnum.linear_operators as linear_operators
 import numpy as np
 
-import probnum.linalg.linops as linops
 from benchmarks.benchmark_utils import SPD_MATRIX_5x5
 from probnum import random_variables as rvs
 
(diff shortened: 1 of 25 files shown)


pnkraemer:
Good work, thank you! One quick question: do you think replacing the name `linops` in `probnum.linops` with `linear_operators`, i.e. `probnum.linear_operators` would be more in line with `probnum.random_variables`? If so, I remember having something like this in the style guide which we might want to update. In principle, I prefer the longer name but I understand if you dont.

JonathanWenger:
> Good work, thank you! One quick question: do you think replacing the name `linops` in `probnum.linops` with `linear_operators`, i.e. `probnum.linear_operators` would be more in line with `probnum.random_variables`? If so, I remember having something like this in the style guide which we might want to update. In principle, I prefer the longer name but I understand if you dont.

I think long term we want to group `random_variables` and `random_processes` under a joint subpackage `prob` again. This would be in line with the style guide which says that subpackages have shortened names while any level below is written with underscores. I don't have a very strong opinion but I think we should be consistent across the board. The advantage of the described convention is that most import statements are kept short this way.

JonathanWenger:
Thinking ahead I could imagine a package structure like:

```html
probnum/
├── diffeq/
├── filtsmooth/
├── linalg/
├── linops/
├── optim/
├── prob/
│   ├── random_variables
│   └── random_processes
├── quad/
├── filtsmooth/
└── utils/
```

In line with keeping things modular all PN method subpackages (`diffeq`, `filtsmooth`, `linalg`, `optim`, `quad`) depend on the core subpackages (`linops`, `prob`, `utils`) and potentially top level functions under `probnum`, e.g. `probnum.aslinop` or `probnum.asrandvar`.

pnkraemer:
Sounds good.

pnkraemer (approved):

PR #229, opened 2020-10-23
Refactor gaussfilters (WIP)

pnkraemer:
Hey @nathanaelbosch the tests are running again. I do not think I am done, but I am well on my way. If you want to take a look (e.g. at the documentations), go ahead. This is why I have made this PR public. I will do some more clean-ups (eg. in `ivp2filter` and in `odeprior`) but the gist of the proposed changes are there. I will tell you when I am done and then I can walk you through in person.


pnkraemer:
now that this one is reverted (sorry for merging accidentally), how do we get it back up?

PR #233, opened 2020-11-03
Remove unused code for (classic) quadrature and utility functions

### LINKED ISSUE #155: Refactor `probnum.quad` to resolve the pylint messages
At the time of writing, `probnum.quad` triggers the following `pylint` messages:
- [x] `attribute-defined-outside-init`
- [ ] `too-few-public-methods`
- [x] `redefined-builtin`
- [x] `arguments-differ`
- [x] `unused-argument`
- [x] `missing-module-docstring`

__Goal:__
Then, after fixing all errors/warnings, you should remove the exception from the corresponding line in `./tox.ini`, such that this message will be tracked in the future.

__To work on this issue__ you can check for these messages individually by calling:
```
pylint src/probnum/quad --disable=all --enable=<message>
```
Then, after fixing all errors/warnings, you should remove the exception from the corresponding line in `./tox.ini`, such that this message will be tracked in the future.

JonathanWenger:
This PR removes unused legacy code and restructures the `quad` subpackage in preparation for the addition of Bayesian quadrature to the package.

This closes  #155 .


alpiges:
Looks good, just two questions/points about the package structure:
1. I'd suggest to move `bayesquad.py` back under `quad` and rename `bayesian_quadrature` to `bq_methods` or something along these lines.
2. Why keep the BQ methods in a private module?

JonathanWenger:
> Looks good, just two questions/points about the package structure:
> 
>     1. I'd suggest to move `bayesquad.py` back under `quad` and rename `bayesian_quadrature` to `bq_methods` or something along these lines.
Done.
> 
>     2. Why keep the BQ methods in a private module?
The separation into private modules in a joint folder signifies what a user should and shouldn't import. With BQ my idea was to have the implementation of the methods under `quad/bq_methods/_bayesian_quadrature`. Now if a (power) user wants to build their own BQ method from different components (e.g. under `quad/acquisition_functions/_expected_improvement`) then the imports would look like
```python
from probnum.quad.bq_methods import BayesianQuadrature
from probnum.quad.acquisition_functions import expected_improvement
```
The additional private modules allow us as developers to have a more finely grained file structure with few classes / methods per module.

alpiges:
> With BQ my idea was to have the implementation of the methods under `quad/bq_methods/_bayesian_quadrature`

Generally the subfolder structure makes sense, but I don't see why we would need that in this particular case. Specifically I don't see why we would keep a folder `_bayesian_quadrature` under `bq_methods`. Why does it bear the general name `_bayesian_quadrature`? And what else should go underneath `bq_method`?

JonathanWenger:
> > With BQ my idea was to have the implementation of the methods under `quad/bq_methods/_bayesian_quadrature`
> 
> Generally the subfolder structure makes sense, but I don't see why we would need that in this particular case. Specifically I don't see why we would keep a folder `_bayesian_quadrature` under `bq_methods`. Why does it bear the general name `_bayesian_quadrature`? And what else should go underneath `bq_method`?

From your feedback this seems to be a case of premature abstraction. I am happy to change the structure. My thinking behind the structure was as follows:

My original thought was to have a folder `bq_method` which contains multiple classes describing different BQ methods. For example `WSABIBayesianQuadrature` would be in a file `bq_methods/_wsabi.py` and be a subclass of `BayesianQuadrature` with a specific set of components (i.e. prior, acquisition function, etc). A convenience interface if you will for those working with WSABI a lot. The `BayesianQuadrature` class is in the module of the same name in snake case. This style is PEP8 conform and somewhat more prevalent in pure object-oriented languages with one file per class.

alpiges:
I fully agree with that. But then `VanillaBayesianQuadrature` and `WSABI` should go into their own module. The latter we can also keep out for now since it might be lying around as a skeleton for a while.

src/probnum/quad/bq_methods/_bayesian_quadrature.py
```diff
@@ -0,0 +1,110 @@
...
+            # Update Distribution
+            self.fun0 = None
+
+        # Information on result
+        info = {"model_fit_diagnostic": None}
+
+        return integral, self.fun0, info
+
+
+class WSABIBayesianQuadrature(BayesianQuadrature):
```
alpiges, line 87: I suggest to call this class `WarpedBayesianQuadrature`.

src/probnum/quad/_bayesquad.py
```diff
@@ -0,0 +1,68 @@
...
+true value of the integral.
+"""
+
+from .bq_methods._bayesian_quadrature import (
+    VanillaBayesianQuadrature,
+    WSABIBayesianQuadrature,
+)
+
+
+def bayesquad(fun, fun0, domain, nevals=None, method="vanilla", **kwargs):
```
alpiges, line 17: There should also be an argument `integration_measure`, but we can add it alongside the integration measures at a later stage.

alpiges (approved):

Issue #117, opened 2020-07-23
Proposal: Change `update` return signature


nathanaelbosch:
This time with an issue, instead of commiting and waiting for your comment there ;) Also, this would be a change that requires broader code modifications.

**Motivation:**
Currently, `update` returns `updated_rv, covest, ccest, meanest`. This felt a bit cryptic and hard to understand for me.

**Proposal:**
A few first ideas/proposals on how to improve this:
1. Adding a docstring: This is the obvious improvement and I'll do that independently of the outcome of this discussion
2. Rename variables: I would have proposed `meas_cov`, `crosscov`, and `meas_mean`. The `_est` could be added back of course, but the fact that they describe the "measurement" is useful to me, and `crosscov` is used at other places.
3. Combine `meanest` and `covest` into `measurement_rv`: This would greatly improve the clarity for me. The output would then become `updated_rv, measurement_rv, crosscov`. But, **this would require changes accross many files**, which is why I wanted to discuss first, not after.

pnkraemer:
All valid points, and thanks for bringing this up before changing it. Let's talk about this in person in the next days?

pnkraemer:
> Combine meanest and covest into measurement_rv: This would greatly improve the clarity for me. The output would then become updated_rv, measurement_rv, crosscov.

I agree, lets do this asap(ish) :)

PR #240, opened 2020-11-11
Change update signature in Kalman

pnkraemer:
Solves #117

### DIFF
--- a/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
+++ b/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
@@ -99,7 +99,7 @@ class Kalman(BayesFiltSmooth):
         """
         data = np.asarray(data)
         predrv, _ = self.predict(start, stop, randvar)
-        filtrv, _, _, _ = self.update(stop, predrv, data)
+        filtrv, _, _ = self.update(stop, predrv, data)
         return filtrv
 
     def predict(self, start, stop, randvar, **kwargs):
--- a/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
+++ b/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
@@ -107,21 +107,24 @@ class Kalman(BayesFiltSmooth):
 
     def update(self, time, randvar, data, **kwargs):
         """
+        Gaussian filter update step. Consists of a measurement step and a conditioning step.
 
         Parameters
         ----------
         time
+            Time of the update.
         randvar
+            Random variable to be updated. Result of :meth:`predict()`.
         data
-        kwargs
+            Data to update on.
 
         Returns
         -------
-        Normal
+        filt_rv : Normal
             Updated Normal RV (new filter estimate).
-        Normal
+        meas_rv : Normal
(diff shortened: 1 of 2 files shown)


src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -99,20 +99,40 @@ def filter_step(self, start, stop, randvar, data, **kwargs):
...
+        Parameters
+        ----------
+        time
+        randvar
+        data
+        kwargs
+
+        Returns
+        -------
+        Normal
```
nathanaelbosch, line 120: What do you think about adding the variable names here for additional clarity?
So:
`filt_rv : Normal` 
`meas_rv : Normal`
`info : Dict`

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -99,20 +99,40 @@ def filter_step(self, start, stop, randvar, data, **kwargs):
...
         predrv, _ = self.predict(start, stop, randvar)
-        filtrv, _, _, _ = self.update(stop, predrv, data)
+        filtrv, _, _ = self.update(stop, predrv, data)
         return filtrv
 
     def predict(self, start, stop, randvar, **kwargs):
         return self.dynamod.transition_rv(randvar, start, stop=stop, **kwargs)
 
     def update(self, time, randvar, data, **kwargs):
+        """
```
nathanaelbosch, line 109: I know I'm annoying, but I think since you already started adding a docstring it might make sense to add a one-liner for the title, types for parameters and parameter descriptions.

What about:
"Update the state by conditioning on data"

Function signature with type hints:
`def update(self, time: FloatArgType, randvar: RandomVariable, data:np.ndarray, **kwargs):` (no idea how to handle kwargs.

Parameter description:
```
        Parameters
        ----------
        time
            Time at which the measurement model should be evaluated
        randvar
            State 
        data
            True, noiseless measurement 
```

Something like that?

nathanaelbosch:
Also thanks a lot for working on this! That helps me a lot with the stuff that I'm implementing :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: Shouldn't have started a docstring...

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: fair enough. Tbh, I though this would somehow mess up sphinx, so I didn't do it. Not sure whether this is true, so I will add it now.

nathanaelbosch: Oh true I didn't check that, I just looked at the numpy styleguide and there it seems like an ok thing to do. Sorry if the suggestion might not work correctly :/

pnkraemer:
included your suggestions. Sphinx sems to be fine with the extra doc in the return (it renders well).

nathanaelbosch (approved):
Looks good! Thanks again :)

PR #242, opened 2020-11-11
Implemented the dynamic diffusion and error estimation

nathanaelbosch:
I replaced the current estimators for the diffusion and for the errors with the ones used in [1], as well as in my personal work. I would argue that this provides a more reasonable default behaviour for now since we did never evaluate that the one currently implemented works more reliably. 

I did not start modularizing things so far. I thought that it might be useful to do so in multiple steps. Open questions / next steps for me would be:
- How do we best accomodate the two diffusion variants, i.e. 'time-fixed" and "dynamic"? My goal would be that it can be specified to use either the fixed or dynamic diffusion. However, personally I always use the dynamic one since my results seemed to indicate that it performs better.    
- Absolute and relative tolerances: Not really supported right now. `AdaptiveSteps` compares the error to a `tol_per_step`, which corresponds to an absolute tolerance. The `_rel_and_abs_error` also does not correspond to the procedure described by Hairer et. al. I think it's clear that we want this, but we have to figure out how to best separate the error estimation, scaling with abs and rel tolerances, accepting the step, and proposing a step size.
- The last part to have all my stuff in would be the multivariate diffusion models, but I admit that I'm not too keen on implementing these.

I will open corresponding issues to keep track of things. Let me know what you think about this PR and the plan in general!

[1] Schober, M., Särkkä, S. and Hennig, P.. A probabilistic model for the
numerical solution of initial value problems. Statistics and Computing, 2019.


pnkraemer:
I guess I am not supposed to review code yet, right? 

As to ideas for modularising: maybe we implement the diffusion calibration thing along the lines of

```python

class DiffusionCalibration:

    def __init__(self):
        self.best_guess = None

    def update(self, meas_rv):
        local_evidence = meas_rv.mean @ np.linalg.inv(meas_rv)
        if self.best_guess is None:
            self.best_guess =local_evidence
        else:
            self.best_guess += local_evidence  # wrong but shows idea

    def calibrate_locally(self, pred_rv, meas_rv):
        return pred_rv, meas_rv * self.best_guess

    def calibrate_globally(self, pred_rv, meas_rv):
        pass # swap actions with calibrate_locally for different MLE
``` 

which then can be used in the filter along the lines of

```python
    for t, d in zip(times, data):
        pred_rv = dynamic_model.transition_rv(...)
        meas_rv = measurement_model.transition_rv()
        if calibration is not None:
            calibration.update(meas_rv)
            pred_rv, meas_rv = calibration.calibrate_locally(pred_rv, meas_rv)
        current_rv = gaussian_update(meas_rv, pred_rv, crosscov)
        append(current_rv)
```
This would allow using your research discoveries in general filters.
Then, in the ODE solver, we can call
```python
def step(self, stuff):
while iterating:
    m, P = fetch_into_preconditioned_coordinates(current.mean, current.cov)
    m, P = self.gfilt.filter_step(m, P)  # takes care of calibration
    e = estimate_error(m, P)
    m, P = push_into_old_coordinates
m, P = self.calibration.calibrate_globally()  # may not do anything
return variables
```
where calibration would be taken care of in the filter step, or in the end. What do you think? 

The adaptive steps will need a bit of rewriting, but if you like, you can ignore this for now. I have rewriten adaptive steps for my project anyway (proportional control with proper internal norms and so on) and we can merge this separately?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -107,40 +124,44 @@ def undo_preconditioning(self, rv):
...
-        ordint, spatialdim = self.gfilt.dynamicmodel.ordint, self.ivp.dimension
-        h0_1d = np.eye(ordint + 1)[:, 0].reshape((1, ordint + 1))
-        projmat = np.kron(np.eye(spatialdim), h0_1d)
-        weights = np.ones(len(abserrors))
-        rel_error = (
-            (abserrors / np.abs(projmat @ currmn)) @ weights / np.linalg.norm(weights)
-        )
-        abs_error = abserrors @ weights / np.linalg.norm(weights)
-        return np.maximum(rel_error, abs_error)
+        z, S = meas_rv.mean, meas_rv.cov
```
pnkraemer, line 160: these dont seem to be used, right?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -43,35 +42,53 @@ def initialise(self):
...
+        return filt_rv, err
 
     def postprocess(self, times, rvs):
         """
         Rescale covariances with sigma square estimate,
         (if specified) smooth the estimate, return ODESolution.
         """
-        rvs = self._rescale(rvs)
+        if False:  # pylint: disable=using-constant-test
+            # will become useful again for time-fixed diffusion models
```
pnkraemer, line 82: thanks for this reminder, I think it will be useful in the future! 

Though (and this is by no means criticism of this occasion here!) it seems to me that those kind of hotfixes get more and more common, no? Maybe we keep an eye on this...

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -33,8 +33,7 @@ def __init__(self, ivp, gaussfilt, with_smoothing):
         if not issubclass(type(gaussfilt.dynamicmodel), ODEPrior):
             raise ValueError("Please initialise a Gaussian filter with an ODEPrior")
         self.gfilt = gaussfilt
-        self.sigma_squared_global = 0.0
-        self.sigma_squared_current = 0.0
+        self.sigma_squared_mle = 1.0
```
pnkraemer, line 36: aren't both sigmas kind of MLEs? (I remember you calling the time-varying one something like a quasi MLE or so). Since this current implementation is only considering the time-varying one, shall we maybe make this clear in this variable name? E.g. `sigma_square_local_mle` or sth? Thoughts?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: I can happily delete the whole `if` branch and add it later again. I just thought it's cleaner to keep it there since the ultimate goal is to implement a fixed diffusion model very soon, which requires that `if` but with some _real_ condition. Should I just remove it for now?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -43,35 +42,53 @@ def initialise(self):
...
+        # 3.1. Adjust the prediction covariance to include the diffusion
+        pred_rv = Normal(
+            pred_rv.mean, pred_rv.cov + (self.sigma_squared_mle - 1) * diffmat
+        )
+        # 3.2 Update the measurement covariance (measure again)
+        meas_rv, info = self.gfilt.measmod.transition_rv(pred_rv, t_new, **kwargs)
+
+        # 4. Update
         zero_data = 0.0
         filt_rv, meas_rv, info = self.gfilt.update(t_new, pred_rv, zero_data, **kwargs)
```
pnkraemer, line 66: is there a reason for calling a full update after the meas_rv is already computed (and updated!)? If I remember correctly, this update will call yet another `measmod.transition_rv`. It might be interesting to extract the update step from the update step (latter I think of as measurement + update, you know what I mean). This can then be called here.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: no, as I said: I think it will be useful in the future :) That was just a side-remark

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Thanks! I'll remove them

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: WIth `simga_squared_mle` I indeed mean the MLE for the _current_ sigma, but that exists for both the dynamic and time-fixed diffusion models. That is also the one that is needed for local error estimation.

In the fixed case, it corresponds to the global sigma, so I don't see a reason to have a separate variable for it. I'm not sure what to call the object that you previously stored in `sigma_squared_current` ("increment"?), but in any case: Do we actually need that anywhere? If not I think a `sigma_squared_mle` should be sufficient for both the dynamic and fixed model.

There is another open issue that's related to this topic: The sigmas need to be accounted for in the dense output in ODESolution and KalmanPosterior. In the "fixed" case that's just a number (the last MLE is the global MLE), in the "dynamic" case we need to save the full array. I'll check if we have an issue to track that and open one if not. But I think that does not have to be fixed in this PR.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: No real reason, I just didn't want to change too many things at the same time. But you're right, that should maybe still be included in this PR!

Should that change happen in `Kalman` or in `BayesFiltSmooth`? I would add a new method `measure(time, randvar)` and another one that performs only the conditioning, which are then both called in `update`. Do you have suggestions on the naming? Maybe something like `condition(state, measurement, crosscov)`? But maybe that sounds a bit too general

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: > WIth simga_squared_mle I indeed mean the MLE for the current sigma, but that exists for both the dynamic and time-fixed diffusion models. That is also the one that is needed for local error estimation.

Hmm okay, that makes sense. 

> Do we actually need that anywhere?

Not if we commit to the time-varying things for now, I guess.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: I guess it is best placed in `Kalman`? What I would do is split the `Kalman.update` in `measure` and sth like `condition_prediction` or so, the latter of which contains the linear algebra things and can be called again here. Naming? I dont know, in my head "update" is "measure" and "update", but in code this is clearly not helpful :D Maybe something really verbose until we have a better idea, like `condition_prediction_on_measurement` or so?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: >> Do we actually need that anywhere?
> Not if we commit to the time-varying things for now, I guess.
What I mean is: Even with the time-fixed sigma, do we ever explicitly need this local increment? I'm asking because the algorithm as I described it in my paper would not use that anywhere.

pnkraemer: Ah I see. I guess it was like this before, because the calibration was implemented in a callback, which did not return anything. I am writing this from memory though, which could be wrong. In general I dont see that it will be necessary, unless for implementation reasons, so I agree with you

nathanaelbosch: Oh I see, I already wondered why it was implemented in a callback but I understood the reason for it only now (and I think I actually implemented that wrongly in my Julia code so thanks a lot for helping me find that bug :D). It might very well be that I'll add a callback back in again then I guess, but maybe the "increment" and "global MLE" will then be stored in some "DiffusionEstimate" object.

pnkraemer: sounds wonderful

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Thanks a lot for the suggestions! I split up the update into measure and condition_state_on_measurement, and I call those in the ODESolver.step now :)

pnkraemer (approved):
Looks good to me, thanks for the updates! IMO we can merge it now :) Lets try to find an idea how to resolve this UKF mess soon :)

PR #243, opened 2020-11-11
Fix the order of callback and saving in ODESolver.solve

### LINKED ISSUE #241: `method_callback` is applied after storing the results, but the description says otherwise
The `method_callback` docstring:
```python
        """
        Optional callback. Can be overwritten.
        Do this as soon as it is clear that the current guess is accepted, but before storing it.
        No return. For example: tune hyperparameters (sigma).
        """
```
But, in `solve` it is used as
```python
            if steprule.is_accepted(stepsize, errorest):
                self.num_steps += 1
                t = t_new
                current_rv = proposed_rv
                times.append(t)
                rvs.append(current_rv)
                self.method_callback(
                    time=t_new, current_guess=proposed_rv, current_error=errorest
                )
```

Should the docstring be adjusted or the order in `solve`? @pnkraemer

nathanaelbosch:
Fixes #241

### DIFF
--- a/src/probnum/diffeq/odesolver.py
+++ b/src/probnum/diffeq/odesolver.py
@@ -39,13 +39,11 @@ class ODESolver(ABC):
 
             if steprule.is_accepted(stepsize, errorest):
                 self.num_steps += 1
-                t = t_new
-                current_rv = proposed_rv
-                times.append(t)
-                rvs.append(current_rv)
                 self.method_callback(
                     time=t_new, current_guess=proposed_rv, current_error=errorest
                 )
+                times.append(t_new)
+                rvs.append(proposed_rv)
 
             suggested_stepsize = self._suggest_step(stepsize, errorest, steprule)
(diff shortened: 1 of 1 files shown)


src/probnum/diffeq/odesolver.py
```diff
@@ -39,13 +39,11 @@ def solve(self, firststep, steprule, **kwargs):
 
             if steprule.is_accepted(stepsize, errorest):
                 self.num_steps += 1
-                t = t_new
-                current_rv = proposed_rv
```
pnkraemer, line 43: Am I super confused or is there no more `current_rv` now? What does the next predict step start with? Do the tests pass? The build seems to fail.

nathanaelbosch: Woops, good catch! Big mistake there, I'll fix that

pnkraemer (approved):
Looks good, thanks for updating! :)

PR #245, opened 2020-11-11
Iterated filtering and smoothing

pnkraemer:
Please don't hate me now, but I wrote some more code. In particular, iterated filtering and smoothing routines based on posterior linearisation (e.g. IEKF or IEKS, but applies out-of-the-box to UKF-style algorithms, too). 

I think it's good to have this established in a basic form, so @SchmidtJonathan can implement more of those methods (levenberg-marquardt [1] and in particular, ADMM [2])

Posterior linearisation uses the KalmanPosterior by default. Choices between IEKF and IEKS are made by specifying a `StoppingCriterion` (which contains methods like `continue_update_iterations` or `continue_filtsmooth_iterations`. As a first example, I implemented a stoppingcriterion which stops if a fixed-point is found. For ODEs we may want a slightly different criterion.

1. https://users.aalto.fi/~ssarkka/pub/lm-eks-camera.pdf
2. https://arxiv.org/pdf/2005.08275.pdf

I will request your review when the CI stops complaining :)  

------------------------
Update: CI seems happy. 

Even though I am assigning you now, take your time to start the review. I am aware that you are doing lots of other things currently, and also that I am demanding lots of your time with reviews. please only get to this whenever you feel like it!


src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -19,8 +19,13 @@ def __init__(self, non_linear_sde, num_steps):
         self.non_linear_sde = non_linear_sde
         self.num_steps = num_steps
 
-    def transition_realization(self, real, start, stop, **kwargs):
-        jacobfun = functools.partial(self.non_linear_sde.jacobian, state=real)
+    def transition_realization(self, real, start, stop, linearise_at=None, **kwargs):
```
nathanaelbosch, line 22: So `linearise_at` is a RV here, right? Since `real` is a np.ndarray, would it not make sense to have `linearise_at` to be a realization, too?

This could also collapse the code to
```python
jacobfun = functools.partial(self.non_linear_sde.jacobian, state=linearize_at if linearize_at is not None else real)
```
or even
```python
jacobfun = functools.partial(self.non_linear_sde.jacobian, state=linearize_at or real)
```
But if that's better is very subjective of course^^

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -32,8 +37,11 @@ def transition_realization(self, real, start, stop, **kwargs):
             dispmatfun=self.non_linear_sde.dispersionmatrix,
         )
 
-    def transition_rv(self, rv, start, stop, **kwargs):
-        jacobfun = functools.partial(self.non_linear_sde.jacobian, state=rv.mean)
+    def transition_rv(self, rv, start, stop, linearise_at=None, **kwargs):
+        if linearise_at is None:
+            jacobfun = functools.partial(self.non_linear_sde.jacobian, state=rv.mean)
+        else:
+            jacobfun = functools.partial(self.non_linear_sde.jacobian, state=rv.mean)
```
nathanaelbosch, line 44: You didn't actually linearise at the `linearise_at` spot ;)

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
```diff
@@ -0,0 +1,132 @@
...
+    def filter_step(
+        self, start, stop, current_rv, data, previous_posterior=None, **kwargs
+    ):
+        """Filter step of iterated filter.
+
+        Different to Kalman.filter in the sense that there may be multiple iterated updates for
+        a single predict, or multiple iterated predicts for a single update, or multiple iterated predicts
+        and updates in general.
+        This retrieves methods such as the iterated extended Kalman filter.
+        By further specifying a `previous_posterio` (KalmanPosterior), the first of those possibly
```
nathanaelbosch, line 38: small typo

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
```diff
@@ -0,0 +1,132 @@
...
+            Posterior distribution of a previous smoothing iteration. Optional.
+            If specified, posterior linearisation is applied.
+
+        Returns
+        -------
+        RandomVariable
+            Resulting filter estimate after the single step.
+        dict
+            Additional information provided by predict() and update().
+            Contains keys `pred_rv`, `info_pred`, `meas_rv`, `info_upd`.
```
nathanaelbosch, line 63: The alternative I see here is to merge the info dict with `info_pred` and `info_upd`. 
What do you think?

nathanaelbosch (commented):
Looks good!

I did not check the stoppingcriterion.py in detail and I could not really "verify" its correctness, since I'm not familiar with the topic at all.

Also, as far as I could tell the tests seem to be unittests for individual parts, notably the stoppingcriterion. Are there any tests that apply the method to some problem to make sure that it "runs" and that it outputs something "resonable"? Or maybe even something better than a non-iterated filter, to verify the correctness of the implementation to some degree? 
This is also not a request! I don't know myself how to properly test such code so I was just wondering about your opinion on this :)

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: linearise_at is a RV, which I think makes sense even in the transition_realization sense: the UKF for instance, linearises at the full gaussian RV. Or am I wrong here? 

The collapsing would work either way I guess:
```
jacobfun = functools.partial(self.non_linear_sde.jacobian, state=linearize_at.mean or real)
```
but personally I am not sure whether this makes it more readable or only shorter...

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: oooops^^

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
pnkraemer: Tbh, I dont know... I was thinking about that, too, but then decided against it. The reason being that this way, it is easy to access the single elements (RV or dict) the way that the sub-methods spit them out without doing some weird merging. Of course a dict of dicts is odd... either way, I think both work, but would slightly (60% to 40%) prefer the dict of dicts over flattening things. What is your opinion?

pnkraemer:
Thanks for the review! 

> Also, as far as I could tell the tests seem to be unittests for individual parts, notably the stoppingcriterion. Are there any tests that apply the method to some problem to make sure that it "runs" and that it outputs something "resonable"? Or maybe even something better than a non-iterated filter, to verify the correctness of the implementation to some degree?

'TestIteratedKalman.test_iterated_filtsmooth` checks that the iterated filter-smoother runs for like 5 iterations. It could be applied to like the pendulum, but I am not sure whether the error would actually reduce, because the problem is so well-behaved. Maybe we try something like an ODE, but that could also be a future PR? 

> I did not check the stoppingcriterion.py in detail and I could not really "verify" its correctness, since I'm not familiar with the topic at all.

This criterion basically checks if the increment compared to the previous iterations is very small and then stops. As to the general class: I tried to write it in a way that it combines both IEKF and IEKS in a single object. Maybe this will turn out to be super convenient, maybe it will bite us. I think it is a decent starting point though (for instance, I think it is much better than what I cooked up in the `StepRule` module initially).

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
nathanaelbosch: Yes if the UKF uses the full distribution it makes sense for `linearise_at` to be a RV. How about naming it `linearize_at_rv`? 

`state=linearize_at.mean or real` will fail though if `linearize_at is None` ;) I don't care too much about the if-else. I think I'd prefer removing it (maybe in favor of assigning a variable and then having a single call to `functools.partial`), but I understand if you prefer the readability with the if-else so feel free to keep it as is and just resolve.

pnkraemer: I think you have a good point with the single call to functools. I will pick a winner tomorrow :) I also think that the `_rv` think might be resolved more cleanly with a type hint? (Since midterm we want to use them anyway...). I will receonsider this thing tomorrow and then resolve if I consider it solved. Thanks for the feedback!

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
nathanaelbosch: With your comment and looking at the code once more, I think the current implementation is probably better than merging the dicts :+1:

nathanaelbosch:
> It could be applied to like the pendulum, but I am not sure whether the error would actually reduce, because the problem is so well-behaved.

Alright, then I guess there's not much value with adding such a test right now.

> Maybe we try something like an ODE, but that could also be a future PR?

Sure we can do that in a future PR once we have some more experience with using those for ODEs.

nathanaelbosch (approved):

PR #249, opened 2020-11-13
Refactor preconditioning

pnkraemer:
This is the current state of the preconditioning refactor. Just that you know what is going on here (and so I am commited to finishing...).

The main changes are that 1) the odeprior is broken up into orthogonal pieces: preconditining, projections, lti sdes, 2) filtering and smoothing know how to use the preconditioning, if it exists.


pnkraemer:
the kwargs in transition punched me in the face, thus I made them all explicit. If we need an additional input to transition_rv in the future, lets add it in there. Side effect, this fixed a few previously unnoticed things (like passing things into a function that dont belong there).

pnkraemer:
with this PR, kwargs confusion is entirely removed from filtsmooth. There is more in diffeq, though.

src/probnum/filtsmooth/bayesfiltsmooth.py
```diff
@@ -10,9 +10,9 @@ class BayesFiltSmooth(ABC):
     Bayesian filtering and smoothing.
     """
 
-    def __init__(self, dynamod, measmod, initrv):
-        self.dynamod = dynamod
-        self.measmod = measmod
+    def __init__(self, dynamic_model, measurement_model, initrv):
+        self.dynamic_model = dynamic_model
+        self.measurement_model = measurement_model
```
pnkraemer, line 15: these used to be properties, and stored internally as dynamod. I found this unnecessary, (i) having a property that just aliases a variable and (ii), using a property for this in general. This pattern also applies to the driftmatrix stuff in sde.

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -20,35 +19,41 @@ def __init__(self, non_linear_sde, num_steps):
...
     def transition_realization(self, real, start, stop, linearise_at=None, **kwargs):
+
         compute_jacobian_at = linearise_at.mean if linearise_at else real
-        jacobfun = functools.partial(
-            self.non_linear_sde.jacobian, state=compute_jacobian_at
-        )
+
+        def jacobfun(t, x=compute_jacobian_at):
+            # replaces functools (second variable may not be called x)
+            return self.non_linear_sde.jacobfun(t, x)
```
pnkraemer, line 27: now that jacobfun is a mere attribute, it is not ensured that the space variable is called state (it is entirely user specified). Thus, functools.partial is replaced with this wrapper.

src/probnum/filtsmooth/statespace/discrete_transition.py
```diff
@@ -31,80 +34,30 @@ class DiscreteGaussian(trans.Transition):
...
-    def transition_realization(self, real, start, stop=None, **kwargs):
-        newmean = self._dynamicsfun(start, real)
-        newcov = self._diffmatfun(start)
+        def if_no_jacobian(t, x):
+            raise NotImplementedError
+
+        self.jacobfun = jacobfun if jacobfun is not None else if_no_jacobian
+
+    def transition_realization(self, real, start, **kwargs):
+        # **kwargs swallow all irrelevant arguments for this function.
```
pnkraemer, line 52: now, this is the only purpose of kwargs. Do you think this is cleaner, too? Shall I remove the comment before merging?

docs/source/tutorials/ordinary_differential_equations.rst
```diff
@@ -13,6 +13,7 @@ See the notebooks below to learn more about how to use ProbNum's functionality f
 
    adaptive_steps_odefilter
    uncertainties_odefilters
+   odesolvers_from_scratch
```
pnkraemer, line 16: I worked in the non-`probsolve_ivp` universe to verify my implementation and decided to just glue the code into a notebook. Nothing fancy, just an illustration of how to solve ODEs without `probsolve_ivp`.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -24,47 +24,50 @@ def __init__(self, ivp, gaussfilt, with_smoothing):
...
-        * gaussfilt.dynamicmodel contains the prior,
+        * gaussfilt.dynamic_model contains the prior,
         * gaussfilt.measurementmodel contains the information about the
         ODE right hand side function,
         * gaussfilt.initialdistribution contains the information about
         the initial values.
         """
-        if not issubclass(type(gaussfilt.dynamicmodel), ODEPrior):
-            raise ValueError("Please initialise a Gaussian filter with an ODEPrior")
+        if not issubclass(type(gaussfilt.dynamic_model), pnfs.statespace.Integrator):
```
pnkraemer, line 33: Formerly `ODEPrior` objects are `Integrator` objects now. Thus, they are in `filtsmooth` instead of `diffeq`. The interface is identical though.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -24,47 +24,50 @@ def __init__(self, ivp, gaussfilt, with_smoothing):
...
-        discrete_dynamics = self.gfilt.dynamod.discretise(t_new - t)
-        diffmat = discrete_dynamics.diffusionmatrix(t_new)
+
+        # Read the diffusion matrix; required for calibration / error estimation
+        discrete_dynamics = self.gfilt.dynamic_model.discretise(t_new - t)
+        diffmat = discrete_dynamics.diffmat
 
         # 1. Predict
-        pred_rv, _ = self.gfilt.predict(t, t_new, current_rv, **kwargs)
+        pred_rv, _ = self.gfilt.predict(t, t_new, current_rv)
```
pnkraemer, line 53: the kwargs are resolved in `filtsmooth` and consequently, not needed here anymore.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -24,47 +24,50 @@ def __init__(self, ivp, gaussfilt, with_smoothing):
...
 
         # 2. Measure
-        meas_rv, info = self.gfilt.measure(t_new, pred_rv, **kwargs)
+        meas_rv, info = self.gfilt.measure(t_new, pred_rv)
 
         # 3. Estimate the diffusion (sigma squared)
-        self.sigma_squared_mle = self._estimate_diffusion(pred_rv, meas_rv)
+        self.sigma_squared_mle = self._estimate_diffusion(meas_rv)
```
pnkraemer, line 59: That was an unused input. Not sure how it got past pylint last PR...

STYLEGUIDE.md
```diff
@@ -88,7 +88,10 @@ Stick to the first few letters for abbreviations if they are sufficiently descri
 Further conventions are
 - `unit2unit`: convert between types or units, e.g. `mat2arr`: convert matrix to array or `s2ms`: convert seconds to milliseconds. Can also be used for simple adapter methods, along the lines of `filt2odefilt`.
 - `proj`: projection (if required: `projmat`, `projvec`, `projlinop`, ...)
-- `precond`: preconditioner
+- `precon`: preconditioner
+- `driftmat`: drift-matrix, `forcevec`: force-vector, `dispmat` dispersion-matrix,
```
pnkraemer, line 92: this reflects the changes made in `statespace.SDE` and the respective subclasses. Now there is only one way of accessing those matrices/functions, under the present naming scheme.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -140,11 +144,11 @@ def _estimate_local_error(self, pred_rv, t_new, calibrated_diffmat, **kwargs):
             Statistics and Computing, 2019.
         """
         local_pred_rv = Normal(pred_rv.mean, calibrated_diffmat)
-        local_meas_rv, _ = self.gfilt.measure(t_new, local_pred_rv, **kwargs)
+        local_meas_rv, _ = self.gfilt.measure(t_new, local_pred_rv)
         error = local_meas_rv.cov.diagonal()
-        return error
+        return np.sqrt(error)
```
pnkraemer, line 149: it needs to be the square-root of the diagonal, no? Or am I missing something? Please tell me if I am wrong.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -164,4 +168,75 @@ def _estimate_diffusion(self, pred_rv, meas_rv):
...
+        return self.gfilt.dynamic_model
+
+
+#####################################################################################################3
+#
+# below the code for using preconditioning in step():
+#
+# It seems slower though and we don't need it for stability in the forward pass?
+#
+#####################################################################################################3
```
pnkraemer, line 180: This. The ODESolver.step() is not using the preconditioner currently. I implemented it and it turned out that it was a bit slower than not using it (110-120% of the runtime). Since we only NEED it in the smoothing iteration, I'd suggest sticking to this for now? What do you think?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -209,41 +201,37 @@ def probsolve_ivp(
...
+     [0.66045798]
+     [0.74373419]
+     [0.81236692]
+     [0.86592532]
+     [0.90598047]
+     [0.93495402]
+     [0.9554465 ]
+     [0.96968678]
+     [0.97948238]
+     [0.98614302]]
```
pnkraemer, line 218: It's my impression that every single PR we change this doctest output, because there is some round-off difference. That is really annoying. Or is it just my machine?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -209,41 +201,37 @@ def probsolve_ivp(
...
     solver = GaussianIVPFilter(ivp, gfilt, with_smoothing=with_smoothing)
-    solution = solver.solve(firststep=firststep, steprule=stprl, **kwargs)
+    solution = solver.solve(firststep=firststep, steprule=stprl)
     return solution
 
 
-def _create_solver_inputs(
-    ivp, method, which_prior, tol, step, firststep, precond_step, **kwargs
-):
+def _create_solver_inputs(ivp, method, which_prior, tol, step, firststep, **kwargs):
```
pnkraemer, line 230: these kinds of functions need a serious update soon, now that much of the APi has changed. I guess the number of lines can be cut in half

src/probnum/diffeq/odesolution.py
```diff
@@ -89,13 +88,6 @@ def __init__(self, times, rvs, solver):
             self._y = _RandomVariableList(rvs)
         self._solver = solver
 
-    def _proj_normal_rv(self, rv, coord):
-        """Projection of a normal RV, e.g. to map 'states' to 'function values'."""
-        q = self._solver.prior.ordint
-        new_mean = rv.mean[coord :: (q + 1)]
-        new_cov = rv.cov[coord :: (q + 1), coord :: (q + 1)]
-        return Normal(new_mean, new_cov)
-
```
pnkraemer, line 98: For the moment, I removed those functions. I know you wrote them and I am sorry if you don't like it. They are not completely lost, so tell me if you want them back in :) My idea behind removing them was that now the projection is encapsulated in `Integrator`, so in the near future we can make projection functions there and use them here again. Are you okay with that?

src/probnum/diffeq/odesolution.py
```diff
@@ -116,30 +108,25 @@ def y(self):
         if self._y:  # hotfix
             return self._y
         else:
-            function_rvs = [self._proj_normal_rv(rv, 0) for rv in self._state_rvs]
+            projmat = self._solver.prior.proj2coord(coord=0)
```
pnkraemer, line 111: see above comment. This also applies to the ones below.

src/probnum/diffeq/odesolution.py
```diff
@@ -116,30 +108,25 @@ def y(self):
...
-        Time-discrete posterior estimates over states, without preconditioning.
-
-        Note that this does not correspond to ``self._kalman_posterior.state_rvs``:
-        Here we undo the preconditioning to make the "states" interpretable.
         """
-        state_rvs = _RandomVariableList(
-            [self._solver.undo_preconditioning(rv) for rv in self._kalman_posterior]
-        )
-        return state_rvs
+        return self._kalman_posterior.state_rvs
```
pnkraemer, line 129: whoop whoop finally we dont need to undo preconditioning everywhere :)

src/probnum/filtsmooth/__init__.py
```diff
@@ -1,42 +1,20 @@
...
     "DiscreteUKFComponent",
     "UnscentedTransform",
-    "Transition",
-    "SDE",
-    "LinearSDE",
-    "LTISDE",
-    "DiscreteGaussian",
-    "DiscreteLinearGaussian",
-    "DiscreteLTIGaussian",
     "FiltSmoothPosterior",
```
pnkraemer, line 15: this __init__ blew up, therfore I extracted `statespace` into its own module. Its' accessible via `probnum.filtsmooth.statespace`.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -188,34 +197,38 @@ def update(self, time, randvar, data, **kwargs):
             which is the crosscov between input RV and measured RV.
             The crosscov does not relate to the updated RV!
         """
-        meas_rv, info = self.measure(time, randvar, **kwargs)
+        # if required, preconditioning can be leveraged here, similarly to `smooth_step`
```
pnkraemer, line 200: shall this comment stay? I didn't implement this, because as of now, we only need preconditioning for smoothing.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -269,9 +288,49 @@ def smooth_step(self, unsmoothed_rv, smoothed_rv, start, stop, **kwargs):
             Time-point of the to-be-smoothed RV.
         stop : float
             Time-point of the already-smoothed RV.
+        intermediate_step :
+            Step-size to be taken by approximate transition methods.
+        """
+        if self.dynamic_model.precon is None:
+            return self._smooth_step_classic(
```
pnkraemer, line 295: I can't come up with a better name. I find that `..._preconditioning` and `..._no_preconditioning` look super similar. `classic` is the way I think about it, but please tell me if you dont like it :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -283,3 +342,42 @@ def smooth_step(self, unsmoothed_rv, smoothed_rv, start, stop, **kwargs):
...
+        unsmoothed_rv : RandomVariable
+            Filtering distribution at time t.
+        smoothed_rv : RandomVariable
+            Prediction at time t+1 of the filtering distribution at time t.
+        start : float
+            Time-point of the to-be-smoothed RV.
+        stop : float
+            Time-point of the already-smoothed RV.
+        """
+        # It is not clear to me how to best test this, except running IBM smoothing for high-ish order. (N)
```
pnkraemer, line 367: This. 

It is tested, because IBM-ode smoothing is tested, though I am not sure whether this is all we want.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -113,12 +113,7 @@ def _smooth_prediction(self, pred_rv, loc):
         next_idx = (self.locations < loc).sum()
         next_loc = self.locations[next_idx]
         next_rv = self._state_rvs[next_idx]
-        next_pred, crosscov = self.gauss_filter.predict(
-            start=loc, stop=next_loc, randvar=pred_rv
-        )
-        smoothed_rv = self.gauss_filter.smooth_step(
-            pred_rv, next_pred, next_rv, crosscov
-        )
+        smoothed_rv = self.gauss_filter.smooth_step(pred_rv, next_rv, loc, next_loc)
```
pnkraemer, line 116: this code was incorrect before this change. We tested it, but not the values, just the shape... We still do, because I am not sure how to test this well, but at least it is correct now :)

src/probnum/filtsmooth/gaussfiltsmooth/unscentedkalman.py
```diff
@@ -22,6 +22,8 @@ def __init__(
             raise TypeError("cont_model must be an SDE.")
         self.non_linear_sde = non_linear_sde
         self.ut = ut.UnscentedTransform(dimension, spread, priorpar, special_scale)
+        super().__init__()
```
pnkraemer, line 25: these take care of initialising (empty) preconditioners.

src/probnum/filtsmooth/statespace/discrete_transition.py
```diff
@@ -131,64 +86,43 @@ class DiscreteLinearGaussian(DiscreteGaussian):
...
+        new_crosscov = rv.cov @ dynamicsmat.T
+        new_cov = dynamicsmat @ new_crosscov + diffmat
+        return pnrv.Normal(mean=new_mean, cov=new_cov), {"crosscov": new_crosscov}
 
     @property
     def dimension(self):
-        return len(self.dynamicsmatrix(0.0).T)
+        # risky to evaluate at zero, but works well
+        # remove this -- the dimension of discrete transitions is not clear!
+        # input dim != output dim is possible...
```
pnkraemer, line 124: these comments show how unsure I am about this property... it works (for now), but seems like a very dangerous few lines of code to me...

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -0,0 +1,265 @@
...
+
+
+class IBM(Integrator, sde.LTISDE):
+    """Integrated Brownian motion in :math:`d` dimensions."""
+
+    def __init__(self, ordint, spatialdim, diffconst):
+        self.diffconst = diffconst
+
+        # initialise BOTH superclasses' inits.
+        # I don't like it either, but it does the job.
```
pnkraemer, line 67: any idea how to better initialise this multiple inheritance thing here? the code works, is kind of readable, but I am not sure how "best practice" this is.

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -0,0 +1,265 @@
...
+    """Integrated Ornstein-Uhlenbeck process in :math:`d` dimensions."""
+
+    def __init__(
+        self,
+        ordint: int,
+        spatialdim: int,
+        driftspeed: float,
+        diffconst: float,
+    ):
+        # Other than previously in ProbNum, we do not use preconditioning for IOUP by default.
```
pnkraemer, line 193: are you okay with this? I can use the `TaylorCoordinates` as a preconditioner here, too, but thought it might be cleaner to leave the preconditioner for IOUP, etc., open ended.

src/probnum/filtsmooth/statespace/preconditioner.py
```diff
@@ -0,0 +1,62 @@
...
+    def __call__(self, step) -> np.ndarray:
+        # if more than step is needed, add them into the signature in the future
+        raise NotImplementedError
+
+    @cached_property
+    def inverse(self) -> "Preconditioner":
+        raise NotImplementedError
+
+
+class TaylorCoordinates(Preconditioner):
```
pnkraemer, line 32: Do you like this name? I had `NordsieckCoordinates` at first, but since they are different and only similar in spirit, I chose this more general name. In fact, `TaylorCoordinates` may also be wrong, but general enough that (hopefully) no one will complain...

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -18,26 +20,37 @@ class SDE(transition.Transition):
...
 
-    def transition_rv(self, rv, start, stop, **kwargs):
+    def transition_rv(
+        self,
+        rv,
+        start,
+        stop=None,
+        step=None,
+        linearise_at=None,
+    ):
```
pnkraemer, line 51: since those remain (abstract) interfaces, I use the full signature and no `kwargs`.

src/probnum/filtsmooth/statespace/transition.py
```diff
@@ -36,29 +36,18 @@ class Transition(abc.ABC):
...
-        The input is either interpreted as a random variable or as a realization.
-        Accordingly, the respective methods are called: :meth:`transition_realization` or :meth:`transition_rv`.
-        """
-        if isinstance(arr_or_rv, RandomVariable):
-            return self.transition_rv(rv=arr_or_rv, start=start, stop=stop, **kwargs)
-        return self.transition_realization(
-            real=arr_or_rv, start=start, stop=stop, **kwargs
-        )
+    def __init__(self):
+        self.precon = None
```
pnkraemer, line 40: as you suggested, preconditioners are instance attributes now.

src/probnum/filtsmooth/statespace/transition.py
```diff
@@ -36,29 +36,18 @@ class Transition(abc.ABC):
...
     def transition_realization(
-        self, real: np.ndarray, start: float, stop: float = None, **kwargs
-    ) -> ("RandomVariable", Dict):
+        self,
+        real: np.ndarray,
+        start: float,
+        stop: Optional[float] = None,
+        step: Optional[float] = None,
+        linearise_at: Optional[RandomVariable] = None,
+    ) -> (RandomVariable, Dict):
```
pnkraemer, line 50: all possible positional arguments are tracked in this interface. I hope this makes it more clear, which kwargs are possible and which arent.

tests/test_diffeq/test_odesolution.py
```diff
@@ -137,19 +137,20 @@ def test_output_shape(self):
...
+    #     chi_squared = np.array(
+    #         [
+    #             chi_squared_statistic(
+    #                 sample, self.solution[:].mean, self.solution[:].cov #+ 1e-8 * np.eye(*self.solution[:].cov.shape)
+    #             )
+    #             for sample in five_samples
+    #         ]
+    #     ).mean()
+    #     self.assertLess(chi_squared, 10.0)
+    #     self.assertLess(0.1, chi_squared)
```
pnkraemer, line 153: these had to be removed, because the posterior covariance did not allow cholesky. There is still one of those 3 chi_square_statistics functions, but the other two broke all the time. I suppose it is because now the covariances are stored in the "real" coordinates.

tests/test_filtsmooth/test_statespace/test_integrator.py
```diff
@@ -0,0 +1,193 @@
+import unittest
+
```
pnkraemer, line 2: the tests in this module are the same tests as previously in `ODEPrior`, with very few additions.

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -0,0 +1,263 @@
...
+    """
+    An integrator is a special kind of SDE,
+    where the :math:`i`th coordinate models the :math:`i`th derivative.
+    """
+
+    def __init__(self, ordint, spatialdim):
+        self.ordint = ordint
+        self.spatialdim = spatialdim
+
+    def proj2coord(self, coord: int) -> np.ndarray:
```
pnkraemer, line 30: this currently returns a matrix. I understand that building and multiplying a projection matrix with lots of zeros might be a bit slow, and that ideally, we either (i) implement a `ProjectionMatrix` operator which does this more efficiently (by implementing `__matmul__` or (ii) by writing a function that slices. I'd postpone this, though. What do you think?

pnkraemer:
Currently, the documentation is not shown for the "new" statespace module. I will fix this tomorrow, and the code can be reviewed nevertheless.

pnkraemer:
finally, sphinx stops complaining :D

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
nathanaelbosch: Sounds good :) I think you could remove the comments here and below, the functools reference is not that understandable anymore after this PR

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
```diff
@@ -71,7 +77,11 @@ def filter_step(
             linearise_predict_at = previous_posterior(start)
             linearise_update_at = previous_posterior(stop)
         pred_rv, info_pred = self.predict(
-            start, stop, current_rv, linearise_at=linearise_predict_at
+            start,
+            stop,
+            current_rv,
+            linearise_at=linearise_predict_at,
+            intermediate_step=intermediate_step,
```
nathanaelbosch, line 84: Is this just something that was overlooked in the IEKS PR?

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
```diff
@@ -94,34 +104,34 @@ def iterated_filtsmooth(self, dataset, times, **kwargs):
             posterior = self.smooth(posterior)
         return posterior
 
-    def predict(self, start, stop, randvar, linearise_at=None, **kwargs):
+    def predict(self, start, stop, randvar, linearise_at=None, intermediate_step=None):
         """(Possibly iterated) prediction step."""
-        pred_rv, info_pred = self.dynamod.transition_rv(
-            randvar, start, stop=stop, linearise_at=linearise_at, **kwargs
+        pred_rv, info_pred = self.dynamics_model.transition_rv(
+            randvar, start, stop=stop, linearise_at=linearise_at, step=intermediate_step
```
nathanaelbosch, line 110: I'm a bit confused about `step=intermediate_step`. I'm not really into the actual code, so I'm just wondering, but did you consider renaming "step" => "intermediate_step"? Or the other way round?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: I like `classic`, and I also agree that `_preconditioning` and `no_preconditioning` do look similar. An alternative would be to use no suffix at all for the classic method, and just `_preconditioned` otherwise, though explicit is better than implicit.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: I think it's fine to not have it implemented, as you say that it should not matter for the forward pass. I'm fine with both keeping and deleting the comment.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: What about something like comparing `_smooth_step_with_preconditioning` to `_smooth_step`, first for a case in which both should output the same, and then for a case in which the latter fails but the first succeeds?

Though I'm not sure if that actually adds something compared to running a high-order low-tolerance IBM..

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: I'm very surprised that tests passed, if the previous version was incorrect. That didn't even really match the function signature??

src/probnum/filtsmooth/gaussfiltsmooth/unscentedkalman.py
```diff
@@ -43,17 +45,18 @@ def __init__(
...
         return self.disc_model.transition_realization(real, start, **kwargs)
 
     def transition_rv(self, rv, start, linearise_at=None, **kwargs):
-        if linearise_at is None:
-            sigmapts = self.ut.sigma_points(rv.mean, rv.cov)
-        else:
-            sigmapts = self.ut.sigma_points(linearise_at.mean, linearise_at.cov)
-        proppts = self.ut.propagate(start, sigmapts, self.disc_model.dynamics)
-        meascov = self.disc_model.diffusionmatrix(start)
+        compute_sigmapts_at = linearise_at if linearise_at else rv
```
nathanaelbosch, line 54: I know that this might even have been my own suggestion, but what do you think about explicitly checking `... if linearise_at is not None else ...`? I'm not sure if that would even matter, the only thing I could think of is if we linearize around some one-dim zero array.

src/probnum/filtsmooth/statespace/discrete_transition.py
nathanaelbosch: Yes I think this is cleaner than before :) It's also fine for me to remove the comments before merging.

src/probnum/filtsmooth/statespace/discrete_transition.py
nathanaelbosch: In this case I agree that the comments are useful! Maybe also open an issue to complement this?

src/probnum/filtsmooth/statespace/integrator.py
nathanaelbosch: I'm fine with postponing :)

src/probnum/filtsmooth/statespace/integrator.py
nathanaelbosch: I don't know either. To be honest, I don't even know what `super()` does in this case^^

Edit: I tried it out. It seems you could fix that to a single call of `super().__init__()`: https://stackoverflow.com/a/16310777

nathanaelbosch: Oh but I don't know if there is a clean way of handling the different arguments. So all in all I guess what you're doing here is fine :)

src/probnum/filtsmooth/statespace/integrator.py
nathanaelbosch: I don't have a real opinion on this, I never left the comfort zone of the IBM prior. Whatever you think is best here :)

src/probnum/filtsmooth/statespace/preconditioner.py
nathanaelbosch: TaylorCoordinates sounds like it is an established thing already. Would you write "Taylor coordinates" in a paper too? But I don't really know about any better naming here neither, I'm fine with whatever you come up with.

I do like the docstring though :D

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -105,85 +128,84 @@ def transition_rv(self, rv, start, stop, step, **kwargs):
...
+        return len(self.driftmatfun(0.0))
 
 
 class LTISDE(LinearSDE):
     """
     Linear time-invariant continuous Markov models of the
     form
-    dx = [F x(t) + u] dt + L dBt.
+
+    .. math:: d x_t = [G x_t + v] d t + L d w_t.
```
nathanaelbosch, line 150: I think I prefered the notation with functions instead of indices. But feel free to just resolve this and leave things as they are!

nathanaelbosch (commented):
I reviewed everything except for the tests for now. Overall I really like the changes!

One thing I just realised: Due to #255 there are a bunch of merge conflicts that need to be resolved :/ If I should help in any way let me know.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -24,47 +24,50 @@ def __init__(self, ivp, gaussfilt, with_smoothing):
...
-        * gaussfilt.dynamicmodel contains the prior,
+        * gaussfilt.dynamics_model contains the prior,
         * gaussfilt.measurementmodel contains the information about the
         ODE right hand side function,
         * gaussfilt.initialdistribution contains the information about
         the initial values.
         """
-        if not issubclass(type(gaussfilt.dynamicmodel), ODEPrior):
-            raise ValueError("Please initialise a Gaussian filter with an ODEPrior")
+        if not issubclass(type(gaussfilt.dynamics_model), pnfs.statespace.Integrator):
```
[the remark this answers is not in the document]
pnkraemer: true. I will refactor. To be honest, I only touch the ODEPrior bit here which is why it is not changed yet.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -24,47 +24,50 @@ def __init__(self, ivp, gaussfilt, with_smoothing):
...
+            raise ValueError(
+                "Please initialise a Gaussian filter with an Integrator (see filtsmooth.statespace)"
```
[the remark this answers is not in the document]
pnkraemer: hmmmm probably, but I dont know...

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: I guess we might want a unittest for adaptive steps that checks that the ratio maxstep/minstep (which is called separation radius, btw) is within known bounds (e.g. roughly equal to 100 for a set algorithm choice). If this square thing would happen, then the step-sizes would change enough.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: I would guess that I screwed something up in the uncertainty calibration. It is probably just as fast, though probably not faster, and since it is not needed here for stability reasons I would argue leaving it out for code readability and clarity

Shall I remove the commented code then? We can always go back to trying it out here in the future if we like?

src/probnum/filtsmooth/bayesfiltsmooth.py
```diff
@@ -36,53 +36,3 @@ def smoother_step(self, **kwargs):
             + "the Bayesian smoother {}.".format(type(self).__name__)
         )
         raise NotImplementedError(errormsg)
-
-    def predict(self, start, stop, randvar, **kwargs):
```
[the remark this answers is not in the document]
pnkraemer: because it is quite cumbersome to force a particle filter into predict and update. I know that the books define a filter like that, but in code, a PF looks different and predict/update is only sensible for Gaussian filters I'd argue.

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
pnkraemer: not sure what you mean, but if I remember, I added something in the IEKS things that were missing before. Could have been this input.

src/probnum/filtsmooth/gaussfiltsmooth/iterated_kalman.py
pnkraemer: I renamed it, because intermediate_step is something that is passed to `filter`, though not related at all to the mesh-width of the observations, and instead a quantity that is used in each prediction. Therefore I wanted to distinguish it from a `step` to avoid confusion. Does this make sense or is it unnecessary?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: I will throw a coin

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: I dont want people to run into the smooth_step_classic thing when they are actually looking for smooth_step. Therefore the explicity (is that a word?). Does that sound reasonable?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: It certainly does, though it might be extremely granular? I dont know, maybe we open an issue that this is not quite tested in an isolated way? Dont have to though

src/probnum/filtsmooth/statespace/integrator.py
pnkraemer: Yeah according to StackOverflow you can do this in a single super(), though in my code this never worked^^ The present thing worked though, which is why it is there :D

src/probnum/filtsmooth/statespace/integrator.py
pnkraemer: Lets not have preconditioning until it bites us then

src/probnum/filtsmooth/statespace/sde.py
pnkraemer: I'll consider this, and compare to the remaining code in sde.py

pnkraemer:
Thanks for the review and for offering help with the conflicts. I'll look into this (probably tomorrow) and let you know when the merge conflicts, as well as your demanded updates are included. Looking forward to getting this done and moving on to square roots and taylor mode! :)

src/probnum/filtsmooth/bayesfiltsmooth.py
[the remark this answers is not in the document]
nathanaelbosch: Thanks for clarifying :)

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: Me too. Was this even tested properly?

src/probnum/filtsmooth/statespace/preconditioner.py
pnkraemer: Indeed, that sounds like an established thing. How about `TaylorLikeCoordinates` or `NordsieckLikeCoordinates` to clarify what they are without lying? In view of the docstring, how about `BetterThanNordsieckCoordinates`? (just kidding... or am I?) In a paper I'd write rescaled coordinates or something along those lines.

src/probnum/filtsmooth/__init__.py
pnkraemer: I dont know... it was getting too big though, which is why I decided to split it up. As far as I understand, much of the statespace module will soon move to some different random-process-related module anyway?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: removed it.

src/probnum/filtsmooth/gaussfiltsmooth/unscentedkalman.py
pnkraemer: I dont think it would matter in the sense that the results are gonna be different. Or is that wrong? Since it is right underneath the signature, it will also be clear what is going on I guess.

pnkraemer:
@nathanaelbosch I re-requested your review, mostly as a reminder that there are still a few questions open for discussion. If you'd like me to act on any of those, just comment accordingly. All points that I considered a to-do are resolved now.

src/probnum/filtsmooth/__init__.py
pnkraemer: Shall we leave it here then or move it into toplevel with its own package?

nathanaelbosch: Ah one thing I still wondered about this: Is statespace only imported by filtsmooth, or also in other parts of the code?

I'm fine with leaving it here and moving it later, but if you think it's better to move it now feel free.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Sounds good!

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -181,66 +173,63 @@ def probsolve_ivp(
...
-     [0.73686744]
-     [0.8053776 ]
-     [0.85895587]
-     [0.89928283]
-     [0.92882899]
-     [0.95007559]
-     [0.96515825]
-     [0.97577054]
-     [0.9831919 ]]
+    >>> print(np.round(solution.y.mean, 2))
```
nathanaelbosch, line 180: Nice fix :D

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: Yes sure, I'm good with `_classic` and `_with_preconditioning` :+1:

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: Your call on what to do here, I don't have a really good overview of the current test coverage :/

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: Well it seems it's even not covered right now so I guess it was not :see_no_evil:

src/probnum/filtsmooth/__init__.py
pnkraemer: By diffeq, too, because there are the `Integrator` priors now.

pnkraemer: Lets leave it for the sake of a faster review and if I change my mind, this will be a quick PR in the future

src/probnum/filtsmooth/gaussfiltsmooth/unscentedkalman.py
nathanaelbosch: What I meant here is that if we have `linearize_at = np.array([0])`, then 
```python
compute_sigmapts_at = linearise_at if linearise_at else rv
```
results in `compute_sigmapts_at == rv`, and not `compute_sigmapts_at == linearize_at`, so for this one weird corner case it would actually result in unintended behaviour.

Though of course it's totally unlikely this will ever happen, so I guess it does not really matter^^

src/probnum/filtsmooth/statespace/preconditioner.py
nathanaelbosch: I like your proposal of `NordsieckLikeCoordinates`, since the way I understand it they are mostly motivated by those Nordsieck coordinates, but improved upon for our setting. I could be easily convinced for another name though

nathanaelbosch (commented):
I reviewed all non-test files again and answered to the open questions. I'll have a look at the tests right now, I just wanted to submit my comments for now :)

tests/test_diffeq/test_odesolution.py
nathanaelbosch: Sounds good

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
```diff
@@ -81,6 +81,7 @@ def test_call_extrapolation(self):
 
 
 class TestKalmanPosteriorSampling(CarTrackingDDTestCase, NumpyAssertions):
+    # There is no check as to whether the samples make sense...
```
nathanaelbosch, line 84: Should we open an issue for this? If not, feel free to just resolve

nathanaelbosch (commented):
Finally reviewed the tests :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: I will open an issue

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: lets open an issue about this (and the lack of tests below, too). Seems like we have to extend the test coverage of kalmanposterior a bit.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
pnkraemer: I will

pnkraemer:
Thanks for the review. I have implemented all things that were left (to the best of my understanding). @nathanaelbosch dare to review again? Or how shall we proceed?

nathanaelbosch (approved):
Thanks for all your work! :)

PR #256, opened 2020-11-19
Reduced strictness of codecov via config file

JonathanWenger:
Codecov checks often failed in PRs in the past where no or only minor changes were made to the codebase. The introduction of the config file reduces the strictness of codecov and should mostly avoid such issues in the future.


nathanaelbosch (approved):
This seems really useful! Thanks :)

Issue #239, opened 2020-11-09
more ode example problems


pnkraemer:
Would be great to have more ODE example problems. Currently we only have `logistic`, `lotkavolterra`, `fitzhughnagumo`. 

Further, I'd like to also have
* three-body (I will post code below)
* seir (https://en.wikipedia.org/wiki/Compartmental_models_in_epidemiology#The_SEIR_model)
* rigid-body (http://www.ece.northwestern.edu/local-apps/matlabhelp/techdoc/math_anal/diffeq8.html)
* van-der-pol (also http://www.ece.northwestern.edu/local-apps/matlabhelp/techdoc/math_anal/diffeq8.html); the default here should probably a non-stiff version...

and perhaps (if necessary, in some simplified version)
* hodgkin-huxley (https://gist.github.com/giuseppebonaccorso/60ce3eb3a829b94abf64ab2b7a56aaef) 
because I catch myself doing every example with lotka-volterra. That is super boring, but probably due to its availability.

pnkraemer:
Three body code (variables may require naming that is consistent with the rest of the probnum source; the snippet is shamelessly stolen) 


```python
def threebody(t,Y):
    # defining the ODE:
    # assume Y = [y1,y2,y1',y2']
    mu = 0.012277471 #  (standardised moon mass)
    mp = 1 - mu    D1 = ((Y[0] + mu)**2 + Y[1]**2)**(3/2)
    D2 = ((Y[0] - mp)**2 + Y[1]**2)**(3/2)    y1p = Y[0] + 2 * Y[3] - mp * (Y[0] + mu) / D1 - mu * (Y[0] - mp) / D2
    y2p = Y[1] - 2 * Y[2] - mp * Y[1] / D1 - mu * Y[1] / D2    return np.array([Y[2],Y[3],y1p,y2p])# defining the initial value
iv = np.array([ 0.994, 0, 0, -2.00158510637908252240537862224 ])
initrv = rvs.Dirac(iv)# defining the time span
Tmax = 17.0652165601579625588917206249
timespan = (0, Tmax)ivp = IVP(timespan, initrv, threebody)
```

pnkraemer:
@schmidtjonathan, while you're working on this, what do you think about moving the examples into a separate file, called `examples.py` or something along these lines? It might slim down `ivp.py` a little bit...

schmidtjonathan:
[...]

PR #262, opened 2020-11-24
More ODE examples

schmidtjonathan:
Related to Issue #239 
Adds more examples of initial value problems to `ivp.py` in the `diffeq` module, along with tests thereof.


pnkraemer:
I unassigned myself as a reviewer, though this doesn't mean a lot. I only did this so you can notify me when this PR is ready to be reviewed, by requesting my review. Does this work for you? Thanks a lot already!

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -40,10 +40,10 @@ def logistic(timespan, initrv, params=(3.0, 1.0)):
     timespan : (float, float)
         Time span of IVP.
     initrv : RandomVariable,
-        RandomVariable that  describes the belief over the initial
-        value. Usually its distribution is Constant (noise-free)
-        or Normal (noisy). To replicate "classical" initial values
-        use the Constant distribution.
+        Scalar-valued RandomVariable (shape: :code:`()`) that describes the belief
```
JonathanWenger, line 43: Just a quick pointer. We recently agreed on a standard on how to document shapes when using type hints. Here's a guide: https://probnum.readthedocs.io/en/latest/development/adding_to_the_api_documentation.html#Docstrings

schmidtjonathan: Thanks, I'll adapt the docstrings to this format

src/probnum/diffeq/ode/__init__.py
```diff
@@ -1,4 +1,15 @@
...
+    "ODE",
+    "IVP",
+    "logistic",
+    "fitzhughnagumo",
+    "seir",
+    "rigidbody",
+    "vanderpol",
+    "threebody",
+    "lotkavolterra",
+]
```
pnkraemer, line 15: Do we want this here? Or is it sufficient to have an __all__ in `diffeq/__init__.py`? I know this was here before, but maybe now is as good as ever to remove it

src/probnum/diffeq/ode/ivp.py
```diff
@@ -6,196 +6,7 @@
...
-    y1, y2 = y
-    return np.array([a * y1 - b * y1 * y2, -c * y2 + d * y1 * y2])
-
-
-def lv_jac(t, y, params):
-    """Jacobian for Lotka-Volterra."""
-    a, b, c, d = params
-    y1, y2 = y
-    return np.array([[a - b * y2, -b * y1], [d * y2, -c + d * y1]])
+__all__ = ["IVP"]
```
pnkraemer, line 9: same here. Where do we want the __all__ things? Is there something in the styleguide?

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
+# pylint: disable=unused-variable
```
pnkraemer, line 1: why this? Is it because most ODEs are autonomous, i.e. ignore t?

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+    where :math:`y_0= y(t_0)` is the initial value.
+
+    Parameters
+    ----------
+    timespan : (float, float)
+        Time span of IVP.
+    initrv : RandomVariable,
+        *(shape=())* -- Scalar-valued RandomVariable that describes the belief
+        over the initial value. Usually its distribution is Constant (noise-free)
+        or Normal (noisy) with scalar mean and scalar variance.
```
pnkraemer, line 45: thanks for tidying this up!!!

src/probnum/diffeq/ode/ivp.py
schmidtjonathan: > Use __all__ = [...] in __init__.py files to fix the order in which the methods are visible in the documentation. This also avoids importing unnecessary functions via import statements from ... import *.

From [here](https://github.com/probabilistic-numerics/probnum/blob/master/STYLEGUIDE.md). It is probably not all that important if you want to expose all the functionality from the module, but as it is made explicit in the styleguide I thought I'd better include it

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+        \begin{pmatrix}
+            \frac{-\beta y_1 y_3}{N} \\
+            \frac{\beta y_1 y_3}{N} - \alpha y_2 \\
+            \alpha y_2 - \gamma y_3 \\
+            \gamma y_3
+        \end{pmatrix}
+
+    for some parameters :math:`(\alpha, \beta, \gamma, N)`.
+    N is a constant population count.
+    Default is :math:`(\alpha, \beta, \gamma, N)=(0.3, 0.3, 0.1, 10^7)`.
```
pnkraemer, line 238: isn't it that N is defined by the sum of values in initrv? Why is it defined as a parameter?

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+
+    Returns
+    -------
+    IVP
+        IVP object describing the SEIR model IVP with the prescribed
+        configuration.
+    """
+
+    def rhs(t, y):
+        return seir_rhs(t, y, params)
```
pnkraemer, line 262: according to codecov, this is not tested. Any idea why?

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: and isnt there a better name than N? E.g. `population_count` or something?

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+
+        f(t, y) =
+        \begin{pmatrix}
+            y_2 y_3 \\
+            -y_1 y_3 \\
+            -0.51 \cdot y_1 y_2
+        \end{pmatrix}
+
+    The ODE system has no parameters.
+
```
pnkraemer, line 293: dont you say with the other ODEs that the problem includes a jacobian? Not too important, though

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+    return np.array([y2, mu * (1.0 - y1 ** 2) * y2 - y1])
+
+
+def vanderpol_jac(t, y, params):
+    y1, y2 = y
+    if isinstance(params, float):
+        mu = params
+    else:
+        (mu,) = params
```
pnkraemer, line 393: maybe remove the brackets?

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+
+    return np.array([[0.0, 1.0], [-2.0 * mu * y2 * y1 - 1.0, mu * (1.0 - y1 ** 2)]])
+
+
+def threebody(timespan, initrv, params=0.012277471):
+    r"""Initial value problem (IVP) based on a three-body problem.
+
+    The three-body problem is defined as follows:
```
pnkraemer, line 401: can you mention p. 129f in hairer & Wanner (1) solving non-stiff diffeqs? It is surprisingly hard to find a reference for this

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+    Let the initial conditions be :math:`y = (y_1, y_2, \dot{y}_1, \dot{y}_2)^T`.
```
pnkraemer, line 402: Can you also write that the threebody problem is turned into a first order linearisation in this function? IN the future we probably wont do this anymore.

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,464 @@
...
+    def rhs(t, y):
+        return threebody_rhs(t, y, params)
+
+    return IVP(timespan, initrv, rhs)
+
+
+def threebody_rhs(t, y, params):
+    y1, y2, y1_dot, y2_dot = y
+    if isinstance(params, float):
+        mu = params
```
pnkraemer, line 455: isnt this the moon mass or sth like this? If so, this might be a neat variable name. If not, keep the mu

tests/test_diffeq/test_ode/test_ivp_examples.py
```diff
@@ -0,0 +1,156 @@
...
+            lg1.jacobian(0.1, random_point) @ random_direction, fd_approx, rtol=1e-2
+        )
+
+    def test_seir(self):
+        """Test the SEIR ODE convenience function."""
+        rv = Constant(np.array([1.0, 0.0, 0.0, 0.0]))
+        lg1 = ivp_examples.seir(self.tspan, rv)
+        self.assertEqual(issubclass(type(lg1), ivp.IVP), True)
+        lg2 = ivp_examples.seir(self.tspan, rv, params=(1.0, 1.0, 1.0, 1.0))
+        self.assertEqual(issubclass(type(lg2), ivp.IVP), True)
```
pnkraemer, line 96: why is the seir jacobian not tested?

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: There does not exist a test case for that. As of now, we do not test evaluations for IVPs for which no Jacobian is implemented. Shall I write according tests?

pnkraemer (commented):
Thank you lots for your work on this. All of my comments are very minor. This looks very good already!

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: I think this can be removed. At least pylint does not complain if I do so. Fixed

src/probnum/diffeq/ode/ivp.py
pnkraemer: I see... though ode is not a real module in the sense that all its functions are pulled up to the diffeq namespace anyway. To me it is not clear whether this means that it needs its own __all__ or not... I am fine with keeping it until then, though especially here it seems a bit redundant...

tests/test_diffeq/test_ode/test_ivp_examples.py
```diff
@@ -0,0 +1,156 @@
...
+    """Test cases for example IVPs: Lotka-Volterra, etc."""
+
+    def setUp(self):
+        self.tspan = (0.0, 4.212)
+
+    def test_logistic(self):
+        """Test the logistic ODE convenience function."""
+        rv = Constant(0.1)
+        lg1 = ivp_examples.logistic(self.tspan, rv)
+        self.assertEqual(issubclass(type(lg1), ivp.IVP), True)
```
pnkraemer, line 20: lets use `isinstance` instead of `issubclass(type`. I am aware that you didnt introduce it, but it would be great to have it fixed now that we're going over it :)

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: True. It would be sensible to save this value somewhere, though. Otherwise, the same value `population_count` would have to be computed in each evaluation of the `rhs`. So either we compute it in each evaluation, we provide it as a parameter (and give it a useful name), or we add it as an attribute to the `IVP` object, which would be maybe the fastest, yet somehow "hackiest" solution. What do you prefer?

tests/test_diffeq/test_ode/test_ivp_examples.py
```diff
@@ -0,0 +1,156 @@
...
+from tests.testing import NumpyAssertions
+
+
+class TestExamples(unittest.TestCase, NumpyAssertions):
+    """Test cases for example IVPs: Lotka-Volterra, etc."""
+
+    def setUp(self):
+        self.tspan = (0.0, 4.212)
+
+    def test_logistic(self):
```
pnkraemer, line 16: what do you thjink about a test of calling the ODE function rhs? This could either be in test_logistic or in another test, I dont care. Wouldnt this also remove lots of the code coverage complaint?

tests/test_diffeq/test_ode/test_ivp_examples.py
JonathanWenger: Just as an addition: If you're writing a test you'll get an even more expressive output if you use `self.assertIsInstance(obj, cls)`, but otherwise what Nico said :).

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: Right, I'll fix that. Thanks!

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: The brackets are actually added by `black`. Are you suggesting to remove the brackets or the entire if condition?

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: Yes, I mentioned this in the docstring. But I will rename the variable, as well

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: how about reading N off the initial random variable into a parameter, which is then fixed by
```python
    def rhs(t, y):
        return seir_rhs(t, y, params) 
```
anyway? I am not sure what the best solution here is, though passing an explicit parameter N that might contradict the initial state (and subsequently, all other states) seems error prone

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: I wrote a corresponding comment in the tests. Bottom line: that would be great, I cant believe we havent done this yet.

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: interesting... then lets keep it! There is no arguing with black ;)

tests/test_diffeq/test_ode/test_ivp_examples.py
pnkraemer: ... because there is none, right? If so, my bad :) (why is there none?)

tests/test_diffeq/test_ode/test_ivp_examples.py
pnkraemer: nice, thanks for the pointer! I should look into all the unittest assert options again at some point...

JonathanWenger: If you want a quick overview, check out the sections in the guide: https://probnum.readthedocs.io/en/latest/development/unit_testing.html

tests/test_diffeq/test_ode/test_ivp_examples.py
schmidtjonathan: There is, now :)

tests/test_diffeq/test_ode/test_ivp_examples.py
schmidtjonathan: It would. I'm going to take care of this

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: Done ✅

src/probnum/diffeq/ode/ivp_examples.py
schmidtjonathan: Done ✅

schmidtjonathan:
@pnkraemer The `__all__` statement in `diffeq.ode.__init__.py` is necessary in order to allow other modules to import the `ivp_examples`. So I added them back. It appears that they can be safely removed from the `ivp.py` and `ivp_examples.py` files, though (which I did).

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,495 @@
...
+
+    return IVP(timespan, initrv, rhs, jac=jac)
+
+
+def vanderpol_rhs(t, y, params):
+    y1, y2 = y
+    if isinstance(params, float):
+        mu = params
+    else:
+        (mu,) = params
```
pnkraemer, line 406: I think this if could be replaced by `mu = float(*params)` which would get rid of this decision tree here. I dont think it is absolutely necessary though, so your call :) I am fine with either solution

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: see comment above :)

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: sorry for the mixed signals, this above comment is what happens if one "just does a quick review" in between two things... A first order linearisation of an ODE is bogus. I meant a first order ODE... Sorry!

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,495 @@
...
+
+    return IVP(timespan, initrv, rhs)
+
+
+def threebody_rhs(t, y, params):
+    y1, y2, y1_dot, y2_dot = y
+    if isinstance(params, float):
+        standardized_moon_mass = params
+    else:
+        (standardized_moon_mass,) = params
```
pnkraemer, line 487: see comment above

tests/test_diffeq/test_ode/test_ivp_examples.py
pnkraemer: Thanks a lot for this! In principle, I think one could refactor those tests out in the future, though for now, I like it as is.

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,495 @@
...
+    Parameters
+    ----------
+    timespan : (float, float)
+        Time span of IVP.
+    initrv : RandomVariable,
+        *(shape=(2, ))* -- Vector-valued RandomVariable that describes the belief
+        over the initial value. Usually its distribution is Constant (noise-free)
+        or Normal (noisy) with :math:`2`-dimensional mean vector and
+        :math:`2 \times 2`-dimensional covariance matrix.
+        To replicate "classical" initial values use the Constant distribution.
```
pnkraemer, line 380: "Constant random variable" instead of "Constant distribution"?

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: is this still relevant or resolved?

src/probnum/diffeq/ode/ivp_examples.py
```diff
@@ -0,0 +1,495 @@
...
+
+
+def vanderpol_rhs(t, y, params):
+    y1, y2 = y
+    if isinstance(params, float):
+        mu = params
+    else:
+        (mu,) = params
+
+    return np.array([y2, mu * (1.0 - y1 ** 2) * y2 - y1])
```
pnkraemer, line 408: one thing I am always confused about with van der pol is whether it is mu (1 - y1^2)y1 - y2 or whether it is mu((1-y1^2)y1 - y2). I know the sources suggest the former, but https://benchmarks.sciml.ai/html/StiffODE/VanDerPol.html suggests the latter. In my experiments, I only got the stiff versions to work with the latter. @nathanaelbosch do you know anything about this? You are the van der pol expert.

src/probnum/diffeq/ode/ivp_examples.py
pnkraemer: maybe also in van der pol

src/probnum/diffeq/ode/ivp_examples.py
nathanaelbosch: Phew apparently I never caught that :/ I did use the latter in my experiments, but I don't have any tests about the former. Thanks a lot for pointing that out!

pnkraemer (approved):
Thanks for doing this so thoroughly!

PR #232, opened 2020-10-26
Random processes

JonathanWenger:
# In a Nutshell
This PR adds random processes as a way to represent uncertainty about a function and functions with stochastic output. As a precursor covariance functions / kernels are added in #263. Random process arithmetic will not be added in this PR.

# Classes

- `RandomProcess`: Objects generalizing functions by adding uncertainty via a covariance function.
- `GaussianProcess`: Standard Gaussian processes without conditioning functionality.
- `GaussMarkovProcess`: Gaussian processes with the Markov property. Mostly used in filtering. Generalizes classes like `FiltSmoothPosterior`, `KalmanPosterior`, `LinearSDEModel`, `LTISDEModel`, `DiscreteGaussianModel`. and provides a common interface for them.

# TODOs

### Random Processes
- [ ] Benchmarks

### Gauss-Markov Processes
- [ ] Basic implementation of Levy processes (Brownian Motion / Wiener, Ornstein-Uhlenbeck) as an example
*Note: These should either be defined via `LinearSDEModel` or at least inherit from these models under `filtsmooth` in some way.*

This closes #44.

PR #263, opened 2020-11-25
Kernels

JonathanWenger:
This PR adds an interface for (positive semi-definite) kernels / covariance functions and some basic implementations of common kernels. This is a first step towards the implementation of random processes (see #232). Kernel arithmetic is not yet implemented in this first PR.

# Classes

- `Kernel`: Kernel interface.
- `WhiteNoise`: White noise kernel.
- `Linear`: Linear kernel
- `Polynomial`: Polynomial kernel
- `RatQuad`: Rational Quadratic kernel
- `ExpQuad`: Exponentiated Quadratic / RBF kernel
- `Matern`: Matern kernel

# Other
- Benchmarks for computation of kernel matrices

# Examples
For example now code like this becomes possible.
```python
>>> import numpy as np
>>> from probnum.kernels import ExpQuad
>>> K = ExpQuad(input_dim=1, lengthscale=0.1)
>>> K(np.array([[1], [.1], [.5]]))
array([[1.00000000e+00, 2.57675711e-18, 3.72665317e-06],
       [2.57675711e-18, 1.00000000e+00, 3.35462628e-04],
       [3.72665317e-06, 3.35462628e-04, 1.00000000e+00]])
```


benchmarks/ivpsolvers.py
```diff
@@ -20,12 +20,11 @@ class IVPSolve:
     params = [["eks0", "ekf0"], ["with", "without"], ["ibm4", "ioup4", "matern92"]]
 
     def setup(self, method, precond, prior):
-        # pylint: disable=attribute-defined-outside-init,invalid-name,unused-argument
+        # pylint: disable=invalid-name
```
pnkraemer, line 23: thanks. why are these not necessary anymore? Are they ignored globally?

docs/source/public_api/kernels.rst
```diff
@@ -0,0 +1,12 @@
+probnum.kernels
+========================
+
+Kernels or covariance functions.
```
pnkraemer, line 4: Sorry if this is late, but I think now is still better than in the future: 

What are your thoughts on changing `Kernel` to `CovarianceKernel`. The term "kernel" is associated with so many things. (Same for "covariance" btw.). I personally am never sure whether I prefer "kernel" or "covariance function" so I tend to use both. This would make the names a bit longer, but would avoid any sort of confusion... I guess?

docs/source/public_api/kernels.rst
```diff
@@ -0,0 +1,12 @@
...
+
+Kernels describes the spatial or temporal variation of a random process. If
+evaluated at two sets of points a kernel is defined as the covariance of the
+values of the random process at these locations.
```
pnkraemer, line 8: Can you imagine any use case that is not a Gaussian processes? If not, maybe change this text here accordingly?

src/probnum/__init__.py
```diff
@@ -23,7 +25,8 @@
 
 from pkg_resources import DistributionNotFound, get_distribution
 
-from . import diffeq, filtsmooth, linalg, linops, quad, random_variables, utils
+from . import diffeq, filtsmooth, kernels, linalg, linops, quad, random_variables, utils
+from .kernels import Kernel
```
pnkraemer, line 29: what is the reason for this extra import?

src/probnum/diffeq/ode/ivp.py
```diff
@@ -30,10 +30,10 @@ def logistic(timespan, initrv, params=(3.0, 1.0)):
     timespan : (float, float)
         Time span of IVP.
     initrv : RandomVariable,
-        RandomVariable that  describes the belief over the initial
+        RandomVariable that describes the belief over the initial
         value. Usually its distribution is Constant (noise-free)
-        or Normal (noisy). To replicate "classical" initial values
-        use the Constant distribution.
+        or Normal (noisy). To replicate 'classical' initial values
+        use a Constant random variable.
```
pnkraemer, line 36: ... thanks for this correction, you're of course right! :)

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,73 @@
...
+            x1 = np.atleast_2d(x1)
+            pdists = scipy.spatial.distance.cdist(
+                x0 / self.lengthscale, x1 / self.lengthscale, metric="euclidean"
+            )
+
+        # Kernel matrix
+        kernmat = np.exp(-(pdists ** 2) / 2.0)
+        return self._transform_kernelmatrix(
+            kerneval=kernmat, x0_shape=x0_originalshape, x1_shape=x1_originalshape
+        )
```
pnkraemer, line 73: So do I understand correctly that the decision whether `__call__` returns a kernel matrix or a single kernel evaluation is made by specifying  `input_dim` and then checking at the call, whether the user put in a single point or a stack of points? That is a good idea, and certainly one way of doing this.

Though, and this is just to mention a possible alternative, not because I think it is superior, wouldnt the interface be more didactic if there was a function `__call__(self, ...)` and another one `matrix(self, ...)` both of which would do different things? This would have the advantage that if something of the wrong shape is passed, an error is raised, instead of doing something that the user does not expect. In this case, one could also avoid the `input_dim` decision at initialisation. This might be irrelevant, but in my code I at some point converged to the `call` *and* the `matrix`, because (i) it was easier to test, and (ii) saved *a lot* of case distinctions in each function.

I am not saying this is better, but I'd like to hear your opinion on this decision :) Thanks a lot for all of this work btw!

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,73 @@
...
+    array([[1.00000000e+00, 3.72665317e-06, 1.92874985e-22],
+           [3.72665317e-06, 1.00000000e+00, 3.72665317e-06],
+           [1.92874985e-22, 3.72665317e-06, 1.00000000e+00]])
+    """
+
+    def __init__(self, input_dim: IntArgType, lengthscale: ScalarArgType = 1.0):
+        self.lengthscale = _utils.as_numpy_scalar(lengthscale)
+        super().__init__(input_dim=input_dim, output_dim=1)
+
+    def __call__(self, x0: _InputType, x1: Optional[_InputType] = None) -> np.ndarray:
```
pnkraemer, line 51: What are the  reasons for making x1 optional? Wouldn't it be a more clear communication to user that if they want the matrix corresponding to a single point set, they need to plug in both? Again, not a very strong opinion, but I think it is easier to explain that a kernel takes *two* inputs, instead of explaining that a kernel *usually* takes two inputs.

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+
+    # pylint: disable="invalid-name"
+    def __init__(
+        self,
+        input_dim: IntArgType,
+        output_dim: IntArgType = 1,
+        kernelfun: Optional[
+            Callable[[_InputType, Optional[_InputType]], np.ndarray]
+        ] = None,
```
pnkraemer, line 51: what would you say is the preferred way of making a kernel? Initialising here, or subclassing? And how does a user know at this point, how the matrix decision is handled, i.e. what happens if the provided kernelfun is not vectorised?

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+    ):
+        self.__input_dim = np.int_(_utils.as_numpy_scalar(input_dim))
+        self.__output_dim = np.int_(_utils.as_numpy_scalar(output_dim))
```
pnkraemer, line 54: why the double underscores here? I only know that they are used to make a class attribute not "over-writable" by a subclass. Is this something that should be the case for input dim and output dim?

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+        Creates a kernel / covariance function from a (non-vectorized) function
+        :math:`k : \\mathbb{R}^{d_{in}} \\times \\mathbb{R}^{d_{in}} \\rightarrow
+        \\mathbb{R}`. When a kernel matrix is computed the given function ``fun`` is
+        applied to all pairs of inputs.
+
+        Parameters
+        ----------
+        fun
+            (Non-vectorized) covariance function.
+        """
```
pnkraemer, line 257: I do not understand what this function does. (I get the name and the docstring, but not the code)...

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+        self.__input_dim = np.int_(_utils.as_numpy_scalar(input_dim))
+        self.__output_dim = np.int_(_utils.as_numpy_scalar(output_dim))
+        self.__kernelfun = (
+            self._as_vectorized_kernel_function(fun=kernelfun)
+            if kernelfun is not None
+            else None
+        )
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}>"
```
pnkraemer, line 62: clever ;) maybe we adopt something like this as default printing for other classes, too?

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+                (x0.ndim == 0 and self.input_dim > 1)  # Scalar input
+                or (x0.ndim == 1 and x0.shape[0] != self.input_dim)  # Vector input
+                or (x0.ndim >= 2 and x0.shape[1] != self.input_dim)  # Matrix input
+            ):
+                raise ValueError(err_msg)
+            equal_inputs = True
+        else:
+            # Promote unequal shapes
+            if x0.ndim < 2 and x1.ndim == 2:
+                x0 = np.atleast_2d(x0)
```
pnkraemer, line 178: is that something we want? Or do we want an exception if this happens?

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+            if x1.ndim < 2 and x0.ndim == 2:
+                x1 = np.atleast_2d(x1)
+            if x0.ndim != x1.ndim:  # Shape mismatch
+                raise ValueError(err_msg)
+
+            # Check shapes
+            if (
+                (x0.ndim == 0 and self.input_dim > 1)  # Scalar input
```
pnkraemer, line 186: how about `def is_scalar_input(x0): return x0.ndim == 0 and self.input_dim > 1)` and similar functions in line 183 or so?

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,302 @@
...
+        return x0, x1, equal_inputs
+
+    def _transform_kernelmatrix(
+        self,
+        kerneval: Union[np.ndarray, ScalarArgType],
+        x0_shape: ShapeArgType,
+        x1_shape: ShapeArgType,
+    ) -> Union[np.float_, np.ndarray]:
+        """Transform the kernel matrix based on the given arguments.
+
```
pnkraemer, line 207: I am not sure I understand what this function does and why it is necessary, sorry to say this so bluntly...

src/probnum/kernels/_linear.py
```diff
@@ -0,0 +1,61 @@
...
+    def __call__(self, x0: _InputType, x1: Optional[_InputType] = None) -> np.ndarray:
+
+        # Check and reshape inputs
+        x0, x1, equal_inputs = self._check_and_transform_input(x0, x1)
+        x0_originalshape = x0.shape
+        x1_originalshape = x1.shape
+
+        # Compute kernel matrix
+        x0 = np.atleast_2d(x0)
+        x1 = np.atleast_2d(x1)
```
pnkraemer, line 55: If I remember correctly, these 7 lines crop up in (almost?) every `__call__` implementation. Maybe there is another way of dealing with this?

src/probnum/kernels/_matern.py
```diff
@@ -0,0 +1,115 @@
...
+        elif self.nu == 1.5:
+            scaled_pdists = np.sqrt(3) * pdists
+            kernmat = (1.0 + scaled_pdists) * np.exp(-scaled_pdists)
+        elif self.nu == 2.5:
+            scaled_pdists = np.sqrt(5) * pdists
+            kernmat = (1.0 + scaled_pdists + scaled_pdists ** 2 / 3.0) * np.exp(
+                -scaled_pdists
+            )
+        elif self.nu == np.inf:
+            kernmat = np.exp(-(pdists ** 2) / 2.0)
```
pnkraemer, line 102: I'd advise against this. it is a neat little gimmick, but strictly speaking, not the matern kernel anymore. Especially, since there is the `ExpQuad` class.

src/probnum/kernels/_polynomial.py
```diff
@@ -0,0 +1,67 @@
...
+
+from ._kernel import Kernel
+
+_InputType = np.ndarray
+
+
+class Polynomial(Kernel[_InputType]):
+    """Polynomial kernel.
+
+    Covariance function defined by :math:`k(x_0, x_1) = (x_0^\\top x_1 + c)^q`.
```
pnkraemer, line 18: Do the parameters (i.e. `c`) here align with the `Linear` kernel? Would it not be easier to maintain if `Linear` was `Polynomial(exponent=1)`? I am worried about the API if we decide to allow products of kernels, in which case one could get the `Polynomial` by multiplying `Linear` kernels. If `c` does not align with both, we might need some awkward API changes...

src/probnum/kernels/_rational_quadratic.py
```diff
@@ -0,0 +1,85 @@
...
+
+    Parameters
+    ----------
+    input_dim :
+        Input dimension of the kernel.
+    lengthscale :
+        Lengthscale of the kernel. Describes the input scale on which the process
+        varies.
+    alpha :
+        Scale mixture. Positive constant determining the weighting between different
```
pnkraemer, line 32: any particular reason for calling this `alpha` and not `scale_mixture` or something along those lines?

src/probnum/kernels/_white_noise.py
```diff
@@ -0,0 +1,52 @@
...
+
+import probnum.utils as _utils
+from probnum.type import IntArgType, ScalarArgType
+
+from ._kernel import Kernel
+
+_InputType = np.ndarray
+
+
+class WhiteNoise(Kernel[_InputType]):
```
pnkraemer, line 15: is there a use case for the white noise kernel? I get all the others, but am not sure whether `WhiteNoise` is useful for e.g. Bayesian quadrature.

src/probnum/utils/arrayutils.py
```diff
@@ -47,10 +45,9 @@ def atleast_1d(*rvs):
 def as_colvec(
     vec: Union[np.ndarray, "probnum.RandomVariable"]
 ) -> Union[np.ndarray, "probnum.RandomVariable"]:
-    """Transform the given vector or random variable to column format.
-
-    Given a vector (or random variable) of dimension (n,) return an array with
-    dimensions (n, 1) instead. Higher-dimensional arrays are not changed.
+    """Transform the given vector or random variable to column format. Given a vector
+    (or random variable) of dimension (n,) return an array with dimensions (n, 1)
+    instead. Higher-dimensional arrays are not changed.
```
pnkraemer, line 50: isn't this the kind of docstring style that pylint and docformatter dislike? If not, no complaints, I am just a bit confused.

tox.ini
```diff
@@ -64,10 +64,11 @@ commands =
...
          pylint src/probnum/filtsmooth --disable="duplicate-code,no-self-use,too-many-locals,too-many-arguments,unused-argument,missing-module-docstring,missing-function-docstring" --jobs=0
+         pylint src/probnum/kernels --disable="too-few-public-methods,invalid-name" --jobs=0
          pylint src/probnum/linalg --disable="attribute-defined-outside-init,too-many-statements,too-many-instance-attributes,too-complex,protected-access,too-many-lines,no-self-use,too-many-locals,redefined-builtin,arguments-differ,abstract-method,too-many-arguments,too-many-branches,duplicate-code,unused-argument,fixme,missing-module-docstring" --jobs=0
-         pylint src/probnum/linops --jobs=0
+         pylint src/probnum/linops --disable="attribute-defined-outside-init,too-many-statements,too-many-instance-attributes,too-complex,protected-access,too-many-lines,no-self-use,too-many-locals,redefined-builtin,arguments-differ,abstract-method,too-many-arguments,too-many-branches,duplicate-code,unused-argument,fixme,missing-module-docstring" --jobs=0
```
pnkraemer, line 69: where does this explosion come from? It doesnt seem to me that there was a lot of change in linops (if any?).

src/probnum/kernels/_matern.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Parameters
+    ----------
+    input_dim :
+        Input dimension of the kernel.
+    lengthscale :
+        Lengthscale of the kernel. Describes the input scale on which the process
+        varies.
+    nu :
+        Hyperparameter controlling differentiability.
```
pnkraemer, line 39: is nu a good name? Wouldn't maybe `reg` (regularity) or `smoothness` or so be more descriptive? FYI: in state space models, the respective value of `nu` is referred to as the order (if I remember correctly, `ordint` for order of integration).

pnkraemer (commented):
Thanks for all of your good work on this! It was very easy to review, which speaks for the quality especially of the documentation. There are a few points I did not understand. Let's resolve them in the respective comment threads, but overall, I like this PR.

benchmarks/ivpsolvers.py
JonathanWenger: yes exactly.

docs/source/public_api/kernels.rst
JonathanWenger: Long term this will move into `probnum.prob.kernels` which I think will alleviate this confusion. To me `CovarianceKernel` sounds somewhat doubled, alternatively I could also imagine `CovarianceFunction`.

docs/source/public_api/kernels.rst
JonathanWenger: Student-t processes come to mind. You can define a covariance function for arbitrary random processes. Maybe this is an argument for naming the class `CovarianceFunction`. This would also nicely correspond to a `Covariance` class for random variables, which might be useful in the future.

src/probnum/__init__.py
JonathanWenger: I removed the Kernel import. This should not be at top-level, I agree.

src/probnum/kernels/_exponentiated_quadratic.py
JonathanWenger: This is mainly for efficiency reasons. If you know, that you are computing the pairwise covariance, this often allows for more efficient computation of the kernel matrix. In particular this also saves you 50% memory.

src/probnum/kernels/_kernel.py
JonathanWenger: I am not sure. I think it makes the function a little more usable. As a new user of GPy I was often confused why I couldn't evaluate the kernel function (which stemmed from the fact that I was passing a vector instead of an array).

tox.ini
JonathanWenger: When `linops` was refactored out of `linalg` the exceptions were still there, but are not caught by `pylint`. In a future refactor these have to be removed one by one.

src/probnum/utils/arrayutils.py
JonathanWenger: I am not sure I understand. I think this was just an automatic reformatting by `docformatter`. What part of it would pylint or docformatter "dislike"?

src/probnum/kernels/_white_noise.py
JonathanWenger: Once kernel arithmetic is added one may want to define a sum kernel with a white noise kernel. Take Bayesian optimization for example. If you assume your function can only be evaluated up to noise, then this corresponds to a Gaussian likelihood or equivalently a sum kernel with a White noise component and an exact likelihood.

src/probnum/kernels/_matern.py
JonathanWenger: I removed it.

tests/test_kernels/test_matern.py
```diff
@@ -0,0 +1,33 @@
...
+        lengthscale = 1.25
+        kernmat_rbf = kerns.ExpQuad(lengthscale=lengthscale, input_dim=x0.shape[1])
+        kernmat_matern = kerns.Matern(
+            lengthscale=lengthscale, nu=np.inf, input_dim=x0.shape[1]
+        )
+        self.assertAllClose(
+            kernmat_rbf(x0, x1),
+            kernmat_matern(x0, x1),
+            msg="RBF and Matern kernel are not equivalent for nu=infty.",
+        )
```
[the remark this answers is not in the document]
JonathanWenger: That's good feedback, I adjusted the test.

src/probnum/kernels/_rational_quadratic.py
JonathanWenger: Mostly I stuck to conventions set by `gpytorch` and `gpy`. I think for kernels some of these parameters are so set in stone, that they are more recognizable by their greek letters.

src/probnum/kernels/_polynomial.py
JonathanWenger: very good point. I changed the API of `Linear` to match `Polynomial`.

src/probnum/kernels/_matern.py
JonathanWenger: I think this is one of those edge cases where the mathematical symbol is so well known, that it is more descriptive than choosing another name. In particular `smoothness` is not quite correct, because of the added 0.5.

src/probnum/kernels/_kernel.py
JonathanWenger: I decided to force the user to subclass Kernel by removing the initialization with `kernelfun` and making the class abstract.

src/probnum/kernels/_kernel.py
JonathanWenger: Changed to single underscore.

src/probnum/kernels/_kernel.py
JonathanWenger: Function was removed since initialization with a function is no longer possible.

src/probnum/kernels/_exponentiated_quadratic.py
JonathanWenger: I've thought about this a bit and I have decided against separating them for the following reasons.
1. The user can still pass only two vectors to `matrix` and will obtain the same behaviour as for `__call__` which means there are two ways of doing the same thing without a clear separation of which one is the correct one. In particular one returns a single entry matrix and another a scalar. This breaks the way we specify covariances to one dimensional random variables.
2. While one would save the argument `input_dim` and `_check_and_transform_input` one would have to specify two functions for each subclass of `Kernel`, namely `__call__` and `matrix` which would lead to some code duplication. 

That being said, one cannot evaluate the diagonal of the kernel matrix without forming the full kernel matrix in the current setup. One solution would be your suggestion with the above caveats, or adding an argument to `__call__(x0, x1, diag=False)`, which can then offer a time/space efficient function to compute just the diagonal of the kernel matrix.

src/probnum/kernels/_kernel.py
JonathanWenger: I changed it somewhat and tried to make the docstring more clear.

src/probnum/kernels/_linear.py
JonathanWenger: That was good feedback. I moved most of this into convenience methods of the superclass which make it easy to subclass `Kernel`.

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,68 @@
...
+        # Check and reshape inputs
+        x0, x1, kernshape = self._check_and_reshape_inputs(x0, x1)
+
+        # Compute pairwise euclidean distances ||x0 - x1|| / l
```
pnkraemer, line 55: super minor and petty comment: in the code, you compute ||x0/l - x1-l|| but the docstring says otherwise... again: very minor, and I am almost feeling bad for writing this...

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,68 @@
...
+           [3.72665317e-06, 1.00000000e+00, 3.72665317e-06],
+           [1.92874985e-22, 3.72665317e-06, 1.00000000e+00]])
+    """
+
+    def __init__(self, input_dim: IntArgType, lengthscale: ScalarArgType = 1.0):
+        self.lengthscale = _utils.as_numpy_scalar(lengthscale)
+        super().__init__(input_dim=input_dim, output_dim=1)
+
+    def __call__(self, x0: _InputType, x1: Optional[_InputType] = None) -> np.ndarray:
+        # Check and reshape inputs
```
pnkraemer, line 52: I think now that the implementation is nicely condensed, there is no need for comments like this anymore?

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,68 @@
...
+            )
+        else:
+            pdists = scipy.spatial.distance.cdist(
+                x0 / self.lengthscale, x1 / self.lengthscale, metric="euclidean"
+            )
+
+        # Compute kernel matrix
+        kernmat = np.exp(-(pdists ** 2) / 2.0)
+
+        return Kernel._reshape_kernelmatrix(kernmat, newshape=kernshape)
```
pnkraemer, line 68: what does this do? Reshape the output if it is a scalar? Because the remaining shape should be handled correctly anyway, no? Please correct me if I am wrong.

also, why do you call this method as a staticmethod here? `self._reshape...` should work as well (or make it a function to start with?) That could be personal preferance on my side though, so it's fair if you disagree.

src/probnum/kernels/_kernel.py
```diff
@@ -0,0 +1,273 @@
...
+import numpy as np
+
+import probnum.utils as _utils
+from probnum.type import IntArgType, ShapeArgType, ShapeType
+
+_InputType = TypeVar("InputType")
+
+
+class Kernel(Generic[_InputType], abc.ABC):
```
pnkraemer, line 14: I think this was a good shout, at least for the foreseeable future. It seems to make the implementation a bit more condensed.

src/probnum/kernels/_linear.py
```diff
@@ -0,0 +1,54 @@
...
+
+from ._kernel import Kernel
+
+_InputType = np.ndarray
+
+
+class Linear(Kernel[_InputType]):
+    """Linear kernel.
+
+    Linear covariance function defined by :math:`k(x_0, x_1) = x_0^\\top x_1 + c`.
```
pnkraemer, line 18: just to make sure: you did not want a shift here, as in `(x1 - a)(x2-a) + b`? I mean, this is also something that can be done in a future PR, I just want to figure out the objective :)

src/probnum/kernels/_matern.py
pnkraemer: > I think this is one of those edge cases where the mathematical symbol is so well known, that it is more descriptive than choosing another name

Is it though? gpytorch does it like this, and wikipedia writes it, but I have already seen many instances where people do not use \nu (e.g. in numerics, people usually use "m").  Though in any case, feel free to go ahead with this, I merely wanted to propose alternatives :)

src/probnum/kernels/_matern.py
```diff
@@ -0,0 +1,109 @@
...
+        # Kernel matrix computation dependent on differentiability
+        if self.nu == 0.5:
+            kernmat = np.exp(-pdists)
+        elif self.nu == 1.5:
+            scaled_pdists = np.sqrt(3) * pdists
+            kernmat = (1.0 + scaled_pdists) * np.exp(-scaled_pdists)
+        elif self.nu == 2.5:
+            scaled_pdists = np.sqrt(5) * pdists
+            kernmat = (1.0 + scaled_pdists + scaled_pdists ** 2 / 3.0) * np.exp(
+                -scaled_pdists
```
pnkraemer, line 95: Is this formula correct? I think it is not, in the sense that `scaled_pdists = np.sqrt(5) * pdists` should be `scaled_pdists = np.sqrt(5) * pdists / 3.0` and then `kernmat = (1.0 + scaled_pdists + scaled_pdists ** 2) * np.exp(-scaled_pdists) ` should be the call (https://en.wikipedia.org/wiki/Mat%C3%A9rn_covariance_function)

src/probnum/kernels/_matern.py
```diff
@@ -0,0 +1,109 @@
...
+            )
+        elif self.nu == np.inf:
+            kernmat = np.exp(-(pdists ** 2) / 2.0)
+        else:
+            pdists[pdists == 0.0] += np.finfo(float).eps
```
pnkraemer, line 100: why this?

src/probnum/kernels/_white_noise.py
pnkraemer: I was wondering less about future applications, but more so about immediate ones. Isn't BO out of scope for the foreseeable future? 
>  If you assume your function can only be evaluated up to noise, then this corresponds to a Gaussian likelihood or equivalently a sum kernel with a White noise component and an exact likelihood.

Is this the way we would model such noisy evaluations though? 

Just to be clear: I do not disagree with having a white noise kernel and also not with the way it is implemented! I was just wondering whether it has a purpose in the near future or not (we will have to maintain the code if it is there, so it better be useful). If you want it in, we put it in!

src/probnum/kernels/_rational_quadratic.py
pnkraemer: But isn't it a bit odd that we have word-names for everything (lengthscale, etc.) but greek letters nu and alpha? In any case, I am okay with it, but just wanted to voice thisopinion at some point...

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,68 @@
...
+            pdists = scipy.spatial.distance.squareform(
+                scipy.spatial.distance.pdist(x0 / self.lengthscale, metric="euclidean")
+            )
+        else:
+            pdists = scipy.spatial.distance.cdist(
+                x0 / self.lengthscale, x1 / self.lengthscale, metric="euclidean"
+            )
+
+        # Compute kernel matrix
+        kernmat = np.exp(-(pdists ** 2) / 2.0)
```
pnkraemer, line 66: Thanks for all of this! One last design suggestion: it seems that implementation of matern, rational quadratic, expquad, and perhaps some more in the future could be achieved by implementing them as stationary covariances, i.e. k(x, y) = g(|x-y|). In this case, ExpQuad would be roughly a single line. Building the kernel matrix is the same everywhere, and makes up about 90 percent of the  `__call__` implementation. 
Of course, there might be reasons to not do it (e.g. derivatives might get a bit harder to compute because of chain rule mess). The advantage would be though that it is also very easy to switch the underlying norms from l2 to say linf. 
Have you thought about this option? Do you think it is convenient or unnecessary?

pnkraemer (commented):
Thanks a lot for the updates Jonathan! I think the clarity is very good, to the point that now I was able to ask more content-related (and sometimes, admittedly, super petty) questions. Once the open discussions are resolved, I can give you my approval :)

src/probnum/kernels/_exponentiated_quadratic.py
```diff
@@ -0,0 +1,68 @@
...
+    def __init__(self, input_dim: IntArgType, lengthscale: ScalarArgType = 1.0):
+        self.lengthscale = _utils.as_numpy_scalar(lengthscale)
+        super().__init__(input_dim=input_dim, output_dim=1)
+
+    def __call__(self, x0: _InputType, x1: Optional[_InputType] = None) -> np.ndarray:
+        # Check and reshape inputs
+        x0, x1, kernshape = self._check_and_reshape_inputs(x0, x1)
+
+        # Compute pairwise euclidean distances ||x0 - x1|| / l
+        if x1 is None:
```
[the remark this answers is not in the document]
JonathanWenger: It used to be this way, but then I would have to return a flag whether the arguments were equal. Now `x1` is `None` if it's not passed. Every kernel that for example relies on `scipy.spatial.pdist` and `scipy.spatial.cdist` benefits from the knowledge whether `x0` and `x1` are the same or not.

src/probnum/kernels/_exponentiated_quadratic.py
JonathanWenger: It reshapes the output to the correct shape as determined by the input and output dimension of the kernel. This allows you to implement a kernel with an efficient way to compute the kernel matrix without having to think about what the right output shape is. All you need to know if you implement `__call__` is that you get one or two arrays N x d and you need to return some array. The correct shape is then determined by `_reshape_kernelmatrix`. The function is static because it doesn't rely on `self`. It's just `reshape` with a corner case for scalars.

src/probnum/kernels/_matern.py
JonathanWenger: I disagree. If you look under the section **Simplification for \nu half integer**, the term is `1 + np.sqrt(5)d / rho + 5 d**2 / (3 rho**2)`, so formulating this with `scaled_pdists = np.sqrt(5)/3` introduces a 1/3 fraction for the linear term in `d` which is not in the definition.

src/probnum/kernels/_matern.py
JonathanWenger: The modified bessel function is only defined on the domain (0, \infty). I added a comment to that effect.

src/probnum/kernels/_linear.py
JonathanWenger: We may want this in the future when we have optimization of kernel hyperparameters, but for now this just corresponds to a fixed shift of the data, which can be part of the preprocessing pipeline.

src/probnum/kernels/_exponentiated_quadratic.py
JonathanWenger: I think this is a very good idea. I just don't know yet whether there should be an abstract interface `StationaryKernel` or just an attribute `is_stationary` so I haven't added this here yet. But for future PRs I think this is one of the first things to add. I lean towards the class since it makes implementing new stationary kernel easy.

src/probnum/kernels/_exponentiated_quadratic.py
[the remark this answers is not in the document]
pnkraemer: I am not sure we are talking about the same thing here... what I meant was that in line 160-162 of `self._check_and_reshape_inputs` one finds the following code
```python
        if x1 is None:
            x1 = x0
            equal_inputs = True ```
that would mean that x1 is not `None` after the call anymore?

File: src/probnum/kernels/_exponentiated_quadratic.py (hunk 31 above)
Comment (pnkraemer) [reply src/probnum/kernels/_exponentiated_quadratic.py:66]: Alright, I agree that this is beyond the current PR.

File: src/probnum/kernels/_exponentiated_quadratic.py (hunk 26 above)
Comment (pnkraemer) [reply src/probnum/kernels/_exponentiated_quadratic.py:68]: Thanks a lot for the thorough answer! 

My comment about the staticmethod was more like "why is this a static method inside `Kernel` and not a function in `_kernel.py`".  Then again, if a class has a staticmethod, I am more used to calling it with `self.some_staticmethod` instead of `MyClass.some_staticmethod` both of which work. However, there may well be good practice that I am missing though, I am not doing a lot with static methods.

File: src/probnum/kernels/_kernel.py [hunk 33]
```diff
@@ -0,0 +1,273 @@
...
+
+            if (
+                (x0.ndim == 0 and self.input_dim > 1)  # Scalar input
+                or (x0.ndim == 1 and x0.shape[0] != self.input_dim)  # Vector input
+                or (x0.ndim >= 2 and x0.shape[1] != self.input_dim)  # Matrix input
+            ):
+                raise ValueError(err_msg)
+            equal_inputs = True
+        else:
+            # Promote unequal shapes
```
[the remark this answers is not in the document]
pnkraemer: Pinging the author here :) This is no real review, rather a good practice question I may need to adopt :)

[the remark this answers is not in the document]
pnkraemer: wait, have I asked the same question twice? There is a very similar thread below. If so, please ignore :D

src/probnum/kernels/_kernel.py
pnkraemer: Okay, as long as this is an explicit choice.  I personally tend to throw errors, but that does not have to mean anything :) I dont think we'd break backwards compatibility too much if we decide to remove this in the future.

src/probnum/kernels/_matern.py
pnkraemer: true, not sure what I was thinking here...

src/probnum/utils/arrayutils.py
pnkraemer: > What part of it would pylint or docformatter "dislike"?
I thought docformatter makes a linebreak after the first sentence (here after "format"). Though I just mention this for clarity of my comment, not because anything needs to be done here.

pnkraemer:
Would be great if you could take another look at my answers to the ongoing discussions (especially the "x1 is none" and "staticmethod" ones). Once my confusion is settled there, I will approve.

src/probnum/kernels/_exponentiated_quadratic.py
JonathanWenger: This is probably somewhat of a design choice, but this method is tailored to be used inside the `__call__` function of kernels which why semantically it makes sense to tie it to the class itself in my opinion. No other use beyond that is intended. If you think this is not valid I am happy to change it. Now calling it with `self` means that if a subclass overrides the static method, this would take precedence. To avoid bugs or at least force the developer to declare their intent I typically use the class name (since this makes explicit which method should be called). But I think this is also debatable.

src/probnum/kernels/_kernel.py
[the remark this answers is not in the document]
JonathanWenger: I have thought about this quite a bit and I am not sure what the best practice is we should adopt. Python functions since it is dynamically typed seems to encourage functions to be vectorized and not care about input shape much. Then again modifying input arguments may be unintuitive to the user. I think we have to discuss this in the software meeting and decide on a few things related to this question. Personally I have struggled a lot with `GPy` initially when I tried the library first simply because it was always expecting 2d row vectors when evaluating the kernel function.

JonathanWenger:
> Would be great if you could take another look at my answers to the ongoing discussions (especially the "x1 is none" and "staticmethod" ones). Once my confusion is settled there, I will approve.

Done! I'm glad you insisted on these since it simplified the code. Thanks for taking another look.

src/probnum/kernels/_exponentiated_quadratic.py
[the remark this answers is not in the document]
pnkraemer: Okay cool thanks for taking care of this. Please dont hate me now, but I think the same updates now need to be done in the other kernel classes as well (at least in Polynomial, which is where I looked last?). Though if you like, you can also merge already and make an issue about this and fix it later. Whatever you reply, I'll give you the approval then :)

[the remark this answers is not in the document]
JonathanWenger: I don't think so. `_check_and_reshape_inputs` now returns `None` for `x1` if only `x0` was passed. This can then be checked in the subclasses to potentially use a more efficient method (such as `pdist`) to compute the kernel matrix.

JonathanWenger:
@pnkraemer In accordance with our move to `pytest` decided at the last software meeting, I refactored the tests. All should be done now hopefully.

pnkraemer (approved):
Looks good, thanks for your patience! 

And well done with the migration to pytest. I thought with #297 I would be the pioneer, but that was wrong apparently :)

PR #264, opened 2020-11-26
Automatic version bump on release

JonathanWenger:
Previously one had to manually change the version string in `setup.cfg` prior to making a release on Github to ensure it would be automatically uploaded to PyPi. With this PR the Github Action triggered on creation of a release (but not for draft releases) changes the version string to the tag given in the release via a commit and then publishes the package to PyPI.


JonathanWenger:
@marvinpfoertner I haven't found a strong reason to move the version number into `__init__.py` beyond avoiding boilerplate. I quite like the "all-in-one-place" `setup.cfg` so I am inclined to leave things as they are in this PR. Let me know if you have any feedback on the Github Action.

Issue #247, opened 2020-11-11
Diffeq: Proportional step size control with both absolute and relative tolerances


nathanaelbosch:
The step size controller which is implemented right now does not support relative tolerances. Ideally we would like to have a controller which closely corresponds to the proportional control algorithm described in Hairer's _Solving Ordinary Differential Equations I_. For a more practically oriented referece see also https://diffeq.sciml.ai/stable/extras/timestepping/.

This topic is closely related to error estimation. In total, the steps are the following:
- Estimate a d-dimensional local error
- Scale this d-dimensional error elementwise, considering the absolute and relative tolerances
- Compute a scalar norm of this scaled local error, typically `E = np.sqrt(np.sum(err**2)/length(err))`
- Reject if E>1, accept otherwise
- Compute the new step, as done currently in `AdaptiveSteps.suggest` (without scaling once more with `tol_per_step`)

__The main question here:__ For each of these elementary steps above, where exactly in the code should they happen? Which of them are in `step`, `_estimate_local_error`, `AdaptiveSteps`, or in `solve`?

pnkraemer:
One other change related to `StepRule`: I'd move `firststep` into it as well, and create new attributes `minstep` and `maxstep`. 

As to your question, I am not entirely sure. My initial guess would've been:

> Estimate a d-dimensional local error

In the solver, because this is likely different for each.

> Scale this d-dimensional error elementwise, considering the absolute and relative tolerances
> Compute a scalar norm of this scaled local error, typically E = np.sqrt(np.sum(err**2)/length(err))

The best place of this is not clear to me. In principle, this is the same for every adaptivity, right? Though to me, it seems like more of a steprule concept, because `E` is tied very closely to the next suggestion step and this way we could ensure that they match. My opinion is not set in stone, though.


> Reject if E>1, accept otherwise
> Compute the new step, as done currently in AdaptiveSteps.suggest (without scaling once more with tol_per_step)

I think this belongs to the steprule, as the last one is different for different control strategies (PI, etc.). 
For reference, below is the proportional control code I used recently. I am not sure how optimal this is, but it does a job :) 

> 
> class ProportionalControl(pnd.StepRule):
>     """Adaptive step size selection based on proportional control."""
> 
>     def __init__(
>         self,
>         atol,
>         rtol,
>         localconvrate,
>         limitchange=(0.2, 5.0),
>         safetyscale=0.9,
>         firststep=1.0,
>         minstep=1e-24,
[...]

PR #269, opened 2020-12-02
Steprule is better

pnkraemer:
These are the improvements discussed in #247. Among other things, 
* firststep is in StepRule
* Proportional control is implemented with atol, rtol, internal norms, etc.
* The tests for adaptive steps are updated accordingly
* ODESolvers have an order.


src/probnum/diffeq/odesolver.py
```diff
@@ -12,11 +11,12 @@
...
     """Interface for ODESolver."""
 
-    def __init__(self, ivp):
+    def __init__(self, ivp, order):
         self.ivp = ivp
+        self.order = order  # RK45 has order=5, IBM(q) has order=q
         self.num_steps = 0
 
-    def solve(self, firststep, steprule, **kwargs):
+    def solve(self, steprule, atol, rtol, **kwargs):
```
pnkraemer, line 19: I chose atol and rtol as arguments to solve, instead of parameters of steprule because in my opinion, they are not part of the step-size-selection. A single steprule should be possible to reuse with different tolerances. Thoughts?

src/probnum/diffeq/odesolver.py
```diff
@@ -28,14 +28,20 @@ def solve(self, firststep, steprule, **kwargs):
         """
         t, current_rv = self.initialise()
         times, rvs = [t], [current_rv]
-        stepsize = firststep
+        stepsize = steprule.firststep
```
pnkraemer, line 31: All steprules have a firststep now, but for e.g. constant steps, this is chosen automatically.

src/probnum/diffeq/odesolver.py
```diff
@@ -28,14 +28,20 @@ def solve(self, firststep, steprule, **kwargs):
...
             proposed_rv, errorest = self.step(t, t_new, current_rv, **kwargs)
-
-            if steprule.is_accepted(stepsize, errorest):
+            internal_norm = steprule.errorest_to_internalnorm(
+                errorest=errorest,
+                proposed_state=proposed_rv.mean,
+                current_state=current_rv.mean,
+                atol=atol,
+                rtol=rtol,
+            )
```
pnkraemer, line 43: I understand that making the computation of this internal norm thingy (naming inspired by SciML) part of the steprule is controversial, I added a comment in the code in AdaptiveSteps to explain why

src/probnum/diffeq/steprule.py
```diff
@@ -1,16 +1,21 @@
...
 
 class StepRule(ABC):
     """(Adaptive) step size rules for ODE solvers."""
 
+    def __init__(self, firststep):
+        self.firststep = firststep
+
     @abstractmethod
-    def suggest(self, laststep, errorest, **kwargs):
+    def suggest(self, laststep, errorest, localconvrate=None):
```
pnkraemer, line 13: since solvers have orders, localconvrate cannot be an attribute of StepRule anymore, because it is initialised independently. In any case, now that we have it like this,I find it more convenient as an argument anyway.

src/probnum/diffeq/steprule.py
```diff
@@ -19,66 +24,109 @@ def is_accepted(self, proposedstep, errorest, **kwargs):
         """
         raise NotImplementedError
 
+    # 'internalnorm' is named after the respective variable used by SciML.
```
pnkraemer, line 27: thoughts? I can also remove the comment once you've seen it :)

src/probnum/diffeq/steprule.py
```diff
@@ -19,66 +24,109 @@ def is_accepted(self, proposedstep, errorest, **kwargs):
...
+        if step < self.minstep:
+            raise RuntimeError("Step-size smaller than minimum step-size")
+        if step > self.maxstep:
+            raise RuntimeError("Step-size larger than maximum step-size")
         return step
 
-    def is_accepted(self, proposedstep, errorest, **kwargs):
-        return errorest * proposedstep < self.tol_per_step
+    # Looks unnecessary, though maybe we want tolerance per unit step
+    # in which case it is good to have such method in here.
```
pnkraemer, line 116: This. Also, for constant steps we want to accept every single time, without computing anything, so I find it convenient to have this method here.

src/probnum/diffeq/steprule.py
```diff
@@ -19,66 +24,109 @@ def is_accepted(self, proposedstep, errorest, **kwargs):
...
+    def is_accepted(self, laststep, scaled_error, localconvrate=None):
+        return scaled_error < 1
+
+    # In here, because (i) we do not want to compute it for constant steps,
+    # and in fact, we don't even want to think about which value atol and rtol should have;
+    # (ii) having it in StepRule makes it easier to test, because the class is more light-weight;
+    # (iii) who knows, maybe there are other ways of dealing with this.
```
pnkraemer, line 123: See comment above.

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -167,13 +167,17 @@ def setUp(self):
 
     def test_kf_ibm1_stdev(self):
         """Standard deviation at end point roughly equal to tolerance."""
-        sol = probsolve_ivp(self.ivp, tol=self.tol, which_prior="ibm1", method="ekf0")
+        sol = probsolve_ivp(
+            self.ivp, atol=self.tol, rtol=self.tol, which_prior="ibm1", method="ekf0"
```
pnkraemer, line 171: I kept the delta small (out of laziness, honestly) by setting both atol and rtol to self.tol

tests/test_diffeq/test_odesolution.py
```diff
@@ -106,7 +106,7 @@ class TestODESolutionAdaptive(TestODESolution):
 
     def setUp(self):
         super().setUp()
-        self.solution = probsolve_ivp(self.ivp, which_prior="ibm2", tol=0.1)
+        self.solution = probsolve_ivp(self.ivp, which_prior="ibm2", atol=0.1, rtol=0.1)
```
pnkraemer, line 109: this is the new atol, rtol interface now.

tests/test_diffeq/test_steprule.py
```diff
@@ -23,23 +23,53 @@ def test_is_accepted(self):
...
+        received = self.asr.errorest_to_internalnorm(
+            errorest, proposed_state, current_state, atol, rtol
+        )
+        self.assertAlmostEqual(expected, received)
+
+    def test_errorest_to_internalnorm_2d(self):
+        errorest = np.array([0.1, 0.2])
+        proposed_state = np.array([1.0, 3.0])
+        current_state = np.array([2.0, 3.0])
+        atol = 0.1
```
pnkraemer, line 67: these tests are new.

pnkraemer:
The new code coverage check is brutal, it points out exactly which lines I didnt care testing... No change for hiding here. The missing bits are some error raisers regarding minstep and the creator functions in probsolve_ivp. How important do you think  it is to solve the latter? I expect a rewrite of probsolve_ivp to happen anytime soon, so maybe it is a waste of time to do this? The minstep tests I will write.

src/probnum/diffeq/odesolver.py
nathanaelbosch: I'm still unsure if `firststep` should be part of steprule, or if it should just be initialized via `step` (in `probsolve_ivp`) but still explicitly passed. I have the impression though that it might be more productive to talk about this directly instead of writing?

src/probnum/diffeq/odesolver.py
nathanaelbosch: Honestly I think in our OOP code it kind of makes sense. But for me it raises the question if "atol" and "rtol" should not be part of the steprule in this case?

Also I never completely understood why it's called "internalnorm" there, so if you have any better idea it would be perfectly fine for me to change that :D

src/probnum/diffeq/odesolver.py
nathanaelbosch: How do you mean "should be possible to reuse with different tolerances"? My first impression is that it could actually make sense to have atol and rtol as part of the steprule.

src/probnum/diffeq/steprule.py
```diff
@@ -19,66 +24,109 @@ def is_accepted(self, proposedstep, errorest, **kwargs):
...
+            raise RuntimeError("Step-size smaller than minimum step-size")
+        if step > self.maxstep:
+            raise RuntimeError("Step-size larger than maximum step-size")
         return step
 
-    def is_accepted(self, proposedstep, errorest, **kwargs):
-        return errorest * proposedstep < self.tol_per_step
+    # Looks unnecessary, though maybe we want tolerance per unit step
+    # in which case it is good to have such method in here.
+    def is_accepted(self, laststep, scaled_error, localconvrate=None):
```
nathanaelbosch, line 117: `is_accepted(self, scaled_error)` should be sufficient now :)

src/probnum/diffeq/steprule.py
```diff
@@ -19,66 +24,109 @@ def is_accepted(self, proposedstep, errorest, **kwargs):
...
+        return scaled_error < 1
+
+    # In here, because (i) we do not want to compute it for constant steps,
+    # and in fact, we don't even want to think about which value atol and rtol should have;
+    # (ii) having it in StepRule makes it easier to test, because the class is more light-weight;
+    # (iii) who knows, maybe there are other ways of dealing with this.
+    def errorest_to_internalnorm(
```
nathanaelbosch, line 124: What do you think about moving this to the base class? This scaling would also be needed in a PI-controller.

nathanaelbosch: Ok I think I understand why: You overwrite it in the constant step rule. I guess the additional computational effort is quite low so that it should not matter too much if it gets called during constant steps, so I think I'd still be in favor of having it in the base class. Optimizing the constant steps requires a bit more work anyways, for example the local errors are computed in the solver step but are never used. What do you think?

nathanaelbosch (commented):
Thanks a lot for working on this and for the PR!

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -27,7 +27,8 @@ def probsolve_ivp(
     ivp,
     method="ekf0",
     which_prior="ibm1",
-    tol=None,
+    atol=None,
```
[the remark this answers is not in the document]
pnkraemer: At some point in the future, yes. For now, let's discuss this in a separate issue I guess. I expect this will boil down to whether we are confident enough in our current adaptive step implementation to make this the default.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -27,7 +27,8 @@ def probsolve_ivp(
...
+    rtol=None,
     step=None,
```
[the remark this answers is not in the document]
pnkraemer: in principle, yes. Though I think this is more about the API of the ODE solvers (aka "how convenient should the convenience function be after all"), rather than the current steprules.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -223,42 +224,45 @@ def probsolve_ivp(
...
-    gfilt, firststep, stprl = _create_solver_inputs(
-        ivp, method, which_prior, tol, step, firststep, precond_step, **kwargs
+    _check_step_tol(step, atol, rtol)
+    gfilt, stprl = _create_solver_inputs(
+        ivp, method, which_prior, step, firststep, precond_step, **kwargs
     )
     with_smoothing = method[-2] == "s" or method[-1] == "s"
     solver = GaussianIVPFilter(ivp, gfilt, with_smoothing=with_smoothing)
-    solution = solver.solve(firststep=firststep, steprule=stprl, **kwargs)
+    solution = solver.solve(steprule=stprl, atol=atol, rtol=rtol, **kwargs)
```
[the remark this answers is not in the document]
pnkraemer: This is how it was before, though I find it a bit awkward to save the steprule in the solver. The decision of which solver to use is kind of orthogonal to the decision which steprule to use, or what do you think? I also think it is a bit unnecessary to let the ODEsolver carry a steprule around even though it is only needed in solve. What do you think?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -223,42 +224,45 @@ def probsolve_ivp(
...
     if precond_step is None:
         precond_step = step or 1.0
     _prior = _string2prior(ivp, which_prior, precond_step, **kwargs)
-    if tol is not None:
-        stprl = steprule.AdaptiveSteps(tol, _prior.ordint + 1, **kwargs)
+    if step is not None:
+        stprl = steprule.ConstantSteps(step)
+    else:
         if firststep is None:
             firststep = ivp.tmax - ivp.t0
```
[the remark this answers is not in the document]
pnkraemer: One could certainly argue about that :D I'd doubt that this makes sense, though in my experience adaptive step sizes find a smaller, better step rather quickly, whereas if this first step is too large, going up to a larger, better step will take a few iterations.

src/probnum/diffeq/steprule.py
pnkraemer: Not sure. Another reason for doing it like this would be that there are ode solvers which cannot compute error estimates (hello particle filter), and as such, are only usable with constant steps. In these cases I tend to return `errorest=np.nan` which would lead to an error if this one here is computed for constant steps. This, and the reasons I explained in the comments below make me think that it is rather well-suited here. I would be lying though, if I said that I think it is clearly better...

src/probnum/diffeq/odesolver.py
pnkraemer: > Honestly I think in our OOP code it kind of makes sense. But for me it raises the question if "atol" and "rtol" should not be part of the steprule in this case?

I dont have a strong opinion here. I kind of like that it is not part of the steprule:
I think that at least mathematically, it is rather elegant to have them separately, in the sense that (in PN algorithm language) it decouples action rules from stopping criteria. Also, we can setup a problem once (solver, prior, steprule etc.) and then solve the problem for different tolerances without redefining old objects (e.g. in workprecision diagrams)

src/probnum/diffeq/odesolver.py
pnkraemer: To be honest, I dont understand what you are talking about...

nathanaelbosch: Sorry if my statement was confusing!

My proposal would be to revert to explicitly passing `firststep`, which, when the solver is called via `probsolve_ivp`, is set to `step`. But this relates also to how the steprule is chosen (inferred from `step` and `tol` vs an explicit argument). Does that make more sense now?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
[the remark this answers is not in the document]
nathanaelbosch: Sure, I opened a new issue for it #272

src/probnum/diffeq/odesolver.py
pnkraemer: Yes, thanks for clarifying. I think I disagree with this, sorry to say this so openly... One of the reasons is that for random-step solvers (do you remember the abdulle/garegnani paper?) it is easiest to randomise steprules instead of solvers. The handling of firststep is then best kept in the steprule. But also in general, I think the first step that the solver takes fits within the scope of steprule (much more than atol/rtol do, IMO). What happens in probsolve_ivp is a different matter to me, I think it may be fun to reconceptualise the code in there soon.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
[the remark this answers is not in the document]
nathanaelbosch: > The decision of which solver to use is kind of orthogonal to the decision which steprule to use, or what do you think?

The vocabulary is not so precise here I think. If "solver" means Euler vs RK4 vs ODE filter then I totally agree! But I think one can also think about the "solver" as the full procedure going from problem to a solution estimate, which includes the step selection algorithm, so it's not so clear for me.

> I also think it is a bit unnecessary to let the ODEsolver carry a steprule around even though it is only needed in solve. What do you think?

This is a really good argument. If the steprule is only used in a single function of the solver, then it's hard to justify saving it in the solver itself. So in summary I agree with the current implementation :)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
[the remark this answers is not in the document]
nathanaelbosch: I think the speed of going up or going down only depends on `limitchange=(0.1, 5.0)`, though with this default choice you are totally right of course. An argument in favor of starting smaller and going up would be that the steps on the way are accepted so the effort is not for nothing. I'll open an issue for it so we don't have to resolve this here :)  #273

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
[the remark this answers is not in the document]
pnkraemer: > The vocabulary is not so precise here I think. If "solver" means Euler vs RK4 vs ODE filter then I totally agree! But I think one can also think about the "solver" as the full procedure going from problem to a solution estimate, which includes the step selection algorithm, so it's not so clear for me.

I meant solver as in `ODESolver`. You're right that this can mean the entire process, though here, I wasn't thinking in such general terms :)

src/probnum/diffeq/steprule.py
nathanaelbosch: I'm sorry but I don't understand your comment :/

nathanaelbosch: Ah I see! I believe that should be handled differently though, with E <- E / h, since the error per step or per unit should also be the one that's used for step size suggestions. Does that make sense?

src/probnum/diffeq/odesolver.py
pnkraemer: atol and rtol are in steprule now

src/probnum/diffeq/odesolver.py
pnkraemer: called "errorest_to_norm" now, and since steprule has attributes atol and rtol, this should indeed be a method

src/probnum/diffeq/steprule.py
pnkraemer: removed

src/probnum/diffeq/steprule.py
pnkraemer: true. Didnt change the code here either way, though, and the comment is removed.

nathanaelbosch (approved):
Looks great! Thanks a lot for the work, and for your patience with my slow and sometimes picky reviews ;)

pnkraemer:
Thank you! 
Btw: I do not think anything was picky, quite the contrary actually :)

PR #276, opened 2020-12-07
Problem Zoo

### LINKED ISSUE #268: Create problems subpackage
### Goal
A set of standard problems for the use of probabilistic numerical methods for users to quickly experiment and prototype.

```python
import probnum
from probnum.testproblems import LotkaVolterra
```
Suggested names for this subpackage are `testproblems`, `problemlib`, `problems`, ...

### Examples
Example problems are predefined ODEs, sparse matrices and linear systems (discretized Dirichlet problem,...).

### Decision
No actual data (e.g. matrices in `npz` format or else) should lie in this subpackage nor ProbNum itself. Rather loader functions lie under `probnum.problems` which download the problems from another location, similar to dataset loaders in other ML frameworks.

### ToDos
- [ ] Refactor example ODE problems

JonathanWenger:
Creates the structure for a problem zoo subpackage collecting example problems for probabilistic numerical methods used for experimentation and rapid prototyping by package users.

This closes #268.


JonathanWenger:
@schmidtjonathan just fyi: you may or may not want to move your newly implemented test problems to the structure implemented here as part of #262.

docs/source/public_api/filtsmooth.rst
```diff
@@ -19,8 +19,9 @@ probnum.filtsmooth.statespace
 
 Probabilistic State Space Models.
 
-This package implements continuous-discrete and discrete-discrete state space models,
-which are the basis for Bayesian filtering and smoothing, but also probabilistic ODE solvers.
+This package implements continuous-discrete and discrete-discrete state
+space models, which are the basis for Bayesian filtering and smoothing,
+but also probabilistic ODE solvers.
```
pnkraemer, line 24: do we have different automatic formatting tools? I remember that changes like these crop up a few times and if it is something that can be resolved for good, it would declutter PRs.

docs/source/public_api/testproblems.rst
```diff
@@ -0,0 +1,24 @@
+probnum.testproblems
+========================
+
+Test problems for probabilistic linear solvers.
```
pnkraemer, line 4: Do you have a strong opinion on the naming? I am not sure, whether `testproblems` communicates the right thing. We do not only use it for tests, but because it is convenient to collect them somewhere and then I think we will use them everywhere. How about, alternatively, something like `problemzoo` (which is inspired by pytorch's modelzoo).

src/probnum/filtsmooth/statespace/__init__.py
```diff
@@ -1,3 +1,10 @@
+"""Probabilistic State Space Models.
+
+This package implements continuous-discrete and discrete-discrete state
+space models, which are the basis for Bayesian filtering and smoothing,
+but also probabilistic ODE solvers.
+"""
+
```
pnkraemer, line 7: did you manually put this in here? If so, thanks a lot!

src/probnum/testproblems/__init__.py
```diff
@@ -0,0 +1 @@
+"""Test problems for probabilistic linear solvers."""
```
pnkraemer, line 1: is there a reason that the testproblems init has only this specific sentence here? Why only for the linear solver? :)

src/probnum/testproblems/linalg/_random_spd_matrix.py
```diff
@@ -0,0 +1,158 @@
...
+    random_state: Optional[RandomStateArgType] = None,
+) -> np.ndarray:
+    """Random symmetric positive definite matrix.
+
+    Constructs a random symmetric positive definite matrix from a given spectrum. An
+    orthogonal matrix :math:`Q` with :math:`\\operatorname{det}(Q)` (a rotation) is
+    sampled with respect to the Haar measure and the spectrum is rotated accordingly
+    resulting in :math:`A=Q \\operatorname{diag}(\\lambda_1, \\dots, \\lambda_n)
+    Q^\\top`. If no spectrum is provided, one is randomly drawn from a Gamma
+    distribution.
```
pnkraemer, line 24: Holy moly it's good that we store those functions somewhere now, given what happens in the random spd functions :D

src/probnum/testproblems/ode/__init__.py
```diff
@@ -0,0 +1,4 @@
+"""Test problems involving ordinary differential equations."""
```
pnkraemer, line 1: to be added.... :)

tests/test_random_variables/test_normal.py
```diff
@@ -9,45 +9,10 @@
 import probnum
 from probnum import linops
 from probnum import random_variables as rvs
+from probnum.testproblems.linalg import random_spd_matrix
```
pnkraemer, line 12: I approve! I think this declutters this test filt quite a lot! :)

tox.ini
```diff
@@ -68,6 +68,7 @@ commands =
          pylint src/probnum/linops --jobs=0
          pylint src/probnum/quad --disable="too-few-public-methods" --jobs=0
          pylint src/probnum/random_variables --disable="missing-function-docstring" --jobs=0
+         pylint src/probnum/testproblems --jobs=0
```
pnkraemer, line 71: ... lets try to keep it this way ;)

pnkraemer (commented):
Thanks a lot for taking care of this! It's in great shape, and I have only very minor comments. The only comment that is potentially worth discussing is the name `testproblems` (see below).

docs/source/public_api/filtsmooth.rst
JonathanWenger: I think this occurred since I played around with how we design the documentation (read from the docstrings or text in restructuredText files). If we both use docformatter via the pre-commit hook, there shouldn't be any differences. Alternatively we could add docformatter to Github Actions as well via tox.

docs/source/public_api/testproblems.rst
JonathanWenger: I don't. I changed it to `problem_zoo`, but I'm open to other suggestions.

src/probnum/testproblems/__init__.py
JonathanWenger: Oops, that was meant to be somewhere else. Changed it to a general description.

JonathanWenger:
> Thanks a lot for taking care of this! It's in great shape, and I have only very minor comments. The only comment that is potentially worth discussing is the name `testproblems` (see below).

Thanks for the review! I changed the name to `problem_zoo`, which I really like.

docs/source/public_api/problem_zoo.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Subpackage offering implementations of standard example problems or convenient
+interfaces to benchmark problems for probabilistic numerical methods. These test
+problems are meant for rapid experimentation and prototyping.
+
+.. automodapi:: probnum.problem_zoo
+    :no-heading:
+    :no-main-docstr:
+
+probnum.problem_zoo.ode
```
pnkraemer, line 14: sorry, it seems that I missed this last time. Do you think that until there is significant overlap between `quad`, `linalg`, etc., it would be best to shadow the module structure in here? In this case this would probably be better called `diffeq`. Though, we can also postpone this until very soon when I move the ivp examples in here.

src/probnum/problem_zoo/linalg/_random_spd_matrix.py
```diff
@@ -0,0 +1,158 @@
...
+    if not 0 <= density <= 1:
+        raise ValueError(f"Density must be between 0 and 1, but is {density}.")
+    chol = np.eye(dim)
+    num_off_diag_cholesky = int(0.5 * dim * (dim - 1))
+    num_nonzero_entries = int(num_off_diag_cholesky * density)
+
+    if num_nonzero_entries > 0:
+        # Draw entries of lower triangle (below diagonal) according to sparsity level
+        entry_ids = np.mask_indices(n=dim, mask_func=np.tril, k=-1)
+        _idx = random_state.choice(
```
pnkraemer, line 148: please dont hate me for this comment, but I dont think you need the leading underscore? Seems to me like the namespace of this variable does not clash with anything.

docs/source/public_api/problem_zoo.rst
JonathanWenger: I agree that we should mostly shadow the package structure here.

Let's discuss this when talking about the API today, but i think it might make sense to separate ODEs and PDEs since the methods are so different. So maybe in the main package we also want `probnum.ode(s)` and `probnum.pde(s)`. 

I or Marvin may add discretized PDE problems to the model zoo in the near future (think Galerkin methods) which I think intended to put in `probnum.model_zoo.pde`.

src/probnum/problems/__init__.py
```diff
@@ -0,0 +1 @@
+"""Definitions and collection of problems solved by probabilistic linear solvers."""
```
pnkraemer, line 1: let's fix the "linear solvers" bit at some point, but for now I dont care :)

src/probnum/problems/zoo/__init__.py
```diff
@@ -0,0 +1 @@
+"""Collection of test problems for probabilistic linear solvers."""
```
pnkraemer, line 1: "linear solvers" everywhere.... ;)

pnkraemer (approved):
There are a few "linear solvers" where there should be "pn methods", but for now I am okay with this. We will probably have a few iterations through the problems in the future, so I think we can merge anyway.

Thanks for working on this!

src/probnum/problems/zoo/__init__.py
JonathanWenger: My brain just can't stop itself!

PR #278, opened 2020-12-09
Fixed RK4 in statistics of linear SDEs

### LINKED ISSUE #252: RK4 wrong
The RK4 step in `sde.linear_sde_statistics` is wrong. It should be 
```
def _rk4_step(mean, cov, time, step, fun):
    """Do a single RK4 step to compute the solution."""
    m1, c1 = fun(time, mean, cov)
    m2, c2 = fun(time + step / 2.0, mean + step * m1 / 2.0, cov + step * c1 / 2.0)
    m3, c3 = fun(time + step / 2.0, mean + step * m2 / 2.0, cov + step * c2 / 2.0)
    m4, c4 = fun(time + step , mean + step * m3, cov + step * c3)
    mean = mean + step * (m1 + 2 * m2 + 2 * m3 + m4) / 6.0
    cov = cov + step * (c1 + 2 * c2 + 2 * c3 + c4) / 6.0
    time = time + step
    return mean, cov, time

```
which is different to the current implementation, because only here the time is advanced: `t -> t + step/2`, etc. Thus for time-varying systems, the solution will be incorrect.

pnkraemer:
Fixes #252 .

### DIFF
--- a/src/probnum/filtsmooth/statespace/sde.py
+++ b/src/probnum/filtsmooth/statespace/sde.py
@@ -280,9 +280,9 @@ def linear_sde_statistics(rv, start, stop, step, driftfun, jacobfun, dispmatfun)
 def _rk4_step(mean, cov, time, step, fun):
     """Do a single RK4 step to compute the solution."""
     m1, c1 = fun(time, mean, cov)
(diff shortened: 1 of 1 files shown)


nathanaelbosch (approved):
Great :)

Issue #279, opened 2020-12-09
Problems Classes


JonathanWenger:
Define via classes what problems are solved by probabilistic numerical methods in `probnum/problems/_problems.py`. Examples of this are

```python
class IVProblem:
    def __init__(ode, y_0):
    ....

class QuadProblem:
    def __init__(f, domain, measure):
    ...

class LinearSystem:
   def __init__(A, b):
   ...
```

pnkraemer:
How about a simple named tuple along the lines of 
```
class IVProblem(NamedTuple):

    ode : "ODE"
    y0 : "RandomVariable"

class QuadProblem(NamedTuple):

    f : callable
    domain : callable  # domain(x) is True: domain(other_x) is False
```

Respectively data classes or the like?

JonathanWenger:
You can use a backport of `dataclasses` in the same way we use one for `cached_property` for Python versions < 3.8: https://pypi.org/project/dataclasses/

In `setup.cfg` we have:
```ini
# Dependencies of the project (semicolon/line-separated):
install_requires =
    numpy<=1.19.1; python_version<="3.6"
    numpy; python_version>"3.6"
    scipy>=1.4
    cached_property; python_version<"3.8"
```
and in the import section of `_random_variable.py`:
```python
try:
    # functools.cached_property is only available in Python >=3.8
    from functools import cached_property
except ImportError:
    from cached_property import cached_property

```

pnkraemer:
neat, thanks for the pointer!

PR #281, opened 2020-12-09
Problem interface

pnkraemer:
This is a work in progress regarding #279 .

I decided to go for something like named tuples, because I think this problem interface should be super light-weight and I do not encourage anyone to let it be more than a collection of attributes that define a PN-worthy problem.
I chose dataclasses, because namedtuples give rise to some really awkward documentation. Dataclasses seem to do the job well. If we decide to change any of these decisions, it should be pretty straightforward. 

Edit: Well well, for some reason I though we'd dropped 3.6 support. Looks like I was wrong.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,6 +38,7 @@ install_requires =
     numpy; python_version>"3.6"
     scipy>=1.4
     cached_property; python_version<"3.8"
+    dataclasses; python_version<"3.7"
 # The usage of test_requires is discouraged, see `Dependency Management` docs
 # tests_require = pytest; pytest-cov
 # Require a specific Python version, e.g. Python 2.7 or >= 3.4
--- a/src/probnum/problems/__init__.py
+++ b/src/probnum/problems/__init__.py
@@ -1,5 +1,5 @@
 """Definitions and collection of problems solved by probabilistic numerical methods."""
 
-from ._problems import IVProblem, LinearSystem, QuadProblem, RegressionProblem
+from ._problems import IVProblem, LinearSystem, QuadratureProblem, RegressionProblem
 
-__all__ = ["RegressionProblem", "IVProblem", "LinearSystem", "QuadProblem"]
+__all__ = ["RegressionProblem", "IVProblem", "LinearSystem", "QuadratureProblem"]
--- a/src/probnum/problems/__init__.py
+++ b/src/probnum/problems/__init__.py
@@ -1,5 +1,15 @@
 """Definitions and collection of problems solved by probabilistic numerical methods."""
 
-from ._problems import IVProblem, LinearSystem, QuadratureProblem, RegressionProblem
+from ._problems import (
+    InitialValueProblem,
+    LinearSystem,
+    QuadratureProblem,
+    RegressionProblem,
+)
 
-__all__ = ["RegressionProblem", "IVProblem", "LinearSystem", "QuadratureProblem"]
+__all__ = [
+    "RegressionProblem",
+    "InitialValueProblem",
+    "LinearSystem",
+    "QuadratureProblem",
+]
--- a/src/probnum/problems/__init__.py
+++ b/src/probnum/problems/__init__.py
@@ -1,5 +1,5 @@
 """Definitions and collection of problems solved by probabilistic numerical methods."""
 
-from ._problems import IVProblem, LinearSystemProblem, QuadProblem, RegressionProblem
+from ._problems import IVProblem, LinearSystem, QuadProblem, RegressionProblem
 
-__all__ = ["RegressionProblem", "IVProblem", "LinearSystemProblem", "QuadProblem"]
+__all__ = ["RegressionProblem", "IVProblem", "LinearSystem", "QuadProblem"]
--- a/src/probnum/problems/__init__.py
+++ b/src/probnum/problems/__init__.py
@@ -1,5 +1,5 @@
 """Definitions and collection of problems solved by probabilistic numerical methods."""
 
-from ._problems import *
+from ._problems import IVProblem, LinearSystemProblem, QuadProblem, RegressionProblem
 
 __all__ = ["RegressionProblem", "IVProblem", "LinearSystemProblem", "QuadProblem"]
--- a/src/probnum/problems/__init__.py
+++ b/src/probnum/problems/__init__.py
@@ -1,1 +1,5 @@
 """Definitions and collection of problems solved by probabilistic numerical methods."""
+
+from ._problems import *
+
+__all__ = ["RegressionProblem", "IVProblem", "LinearSystemProblem", "QuadProblem"]
--- a/src/probnum/problems/_problems.py
+++ b/src/probnum/problems/_problems.py
@@ -1,13 +1,16 @@
-"""Definitions of problems currently solved by probabilistic numerical methods."""
+"""Definitions of problems currently solved by probabilistic numerical methods.
+
+If a problem is not listed here, we have not implemented a probabilistic
+numerical method for it yet.
+"""
 
 import dataclasses
 import typing
 
 import numpy as np
+import scipy.sparse
 
-import probnum.filtsmooth.statespace
-import probnum.linops
-import probnum.type
+import probnum
 
 
 @dataclasses.dataclass
@@ -20,9 +23,7 @@ class RegressionProblem:
 
     observations: np.ndarray
     locations: np.ndarray
-
-    # Not sure how to best deal with this re. dependencies.
-    likelihood: probnum.filtsmooth.statespace.DiscreteGaussian = None
+    likelihood: "probnum.filtsmooth.statespace.DiscreteGaussian" = None
 
 
 @dataclasses.dataclass
@@ -55,7 +56,7 @@ class IVProblem:
 
 @dataclasses.dataclass
 class LinearSystemProblem:
-    r"""Solve a linear system of equations.
+    r"""Linear system of equations.
 
     Compute :math:`x` from :math:`Ax=b`.
     Solved by the probabilistic linear solver in mod:`probnum.linalg`
@@ -71,13 +72,18 @@ class LinearSystemProblem:
            [0., 0., 1.]]), b=array([0, 1, 2]))
     """
 
-    A: typing.Union[probnum.linops.LinearOperator, np.ndarray]
-    b: np.ndarray  # Can probnum.linalg handle multiple RHSs?
+    A: typing.Union[
+        np.ndarray,
+        "scipy.sparse.spmatrix",
+        "probnum.linops.LinearOperator",
+        "probnum.RandomVariable",
+    ]
+    b: typing.Union[np.ndarray, "probnum.RandomVariable"]
 
 
 @dataclasses.dataclass
 class QuadProblem:
-    r"""Numerically approximate an integral.
+    r"""Numerically compute an integral.
 
     Compute the integral
 
--- a/src/probnum/problems/_problems.py
+++ b/src/probnum/problems/_problems.py
@@ -1,1 +1,53 @@
 """Definitions of problems solved by probabilistic numerical methods."""
+
+import typing
+
+import probnum.filtsmooth.statespace
+import probnum.linops
+import probnum.type
+
+
+class RegressionProblem(typing.NamedTuple):
+    r"""Regression problems.
+
+    Find a stochastic process that fits data given a likelihood (realised by a :obj:`Transition`). Solved by, for instance, Kalman filtering and smoothing.
+    """
+
+    observations: np.ndarray
+    locations: np.ndarray
+    likelihood: probnum.filtsmooth.statespace.DiscreteGaussian  # Not sure how to deal with this.
+
+
+class IVProblem(typing.NamedTuple):
+    r"""ODE initial value problems.
+
+    Examples
+    --------
+    >>> ivp = IVProblem(lambda t, x: x*(1-x), 0., 3., 0.1)
+    >>> ivp
+    IVProblem(f=1.0, t0=1.0, tmax=1.0, y0=1.0)
+    >>> ivp.f(ivp.t0, ivp.y0)
+    0.09
+    """
+
+    f: typing.Callable[[float, np.ndarray], np.ndarray]
+    t0: float
+    tmax: float
+    y0: np.ndarray  # Formerly (and unnecessarily) a RandomVariable
+
+
+class LinearSystemProblem(typing.NamedTuple):
+    r"""Solve a linear system of equations: compute :math:`x` from :math;`Ax=b`."""
+
+    A: Union[probnum.linops.LinearOperator, np.ndarray]
+    b: np.ndarray  # Can linalg handle multiple RHSs?
+
+
+class QuadratureProblem(typing.NamedTuple):
+    r"""Numerically approximate an integral.
+
+    Compute :math:`\int_\Omega f(x) \diff \mu(x)` for a function :math:`f: \Omega \rightarrow \mathbb{R}`. For the time being, :math:`\mu` is the Lebesgue measure.
+    """
+
+    integrand: typing.Callable[[np.ndarray], np.ndarray]
+    domain: Callable[[np.ndarray], bool]  # Up for discussion...
--- a/src/probnum/problems/_problems.py
+++ b/src/probnum/problems/_problems.py
@@ -8,7 +8,6 @@ import dataclasses
 import typing
 
 import numpy as np
-import scipy.sparse
 
 import probnum
 
--- a/src/probnum/problems/_problems.py
+++ b/src/probnum/problems/_problems.py
@@ -130,7 +130,7 @@ class LinearSystem:
 
 
 @dataclasses.dataclass
-class QuadProblem:
+class QuadratureProblem:
     r"""Numerical computation of an integral.
 
     Compute the integral
--- a/src/probnum/problems/_problems.py
+++ b/src/probnum/problems/_problems.py
@@ -113,9 +113,9 @@ class LinearSystem:
     --------
     >>> A = np.eye(3)
     >>> b = np.arange(3)
-    >>> lsp = LinearSystemProblem(A, b)
-    >>> lsp
-    LinearSystemProblem(A=array([[1., 0., 0.],
+    >>> lin_sys = LinearSystem(A, b)
+    >>> lin_sys
+    LinearSystem(A=array([[1., 0., 0.],
            [0., 1., 0.],
            [0., 0., 1.]]), b=array([0, 1, 2]))
     """
@@ -162,13 +162,13 @@ class QuadratureProblem:
     >>> np.round(qp1d.integrand(0.2), 2)
     0.04
     >>> qp1d.lower_bd
-    0.42
+    0.41
     >>>
     >>> lower_bd = [0., 0.]
     >>> upper_bd = [1., 1.]
     >>> qp2d = QuadratureProblem(integrand, lower_bd=lower_bd, upper_bd=upper_bd)
     >>> qp2d.upper_bd
-    [1., 1.]
+    [1.0, 1.0]
     """
 
     integrand: typing.Callable[[np.ndarray], np.ndarray]
--- a/src/probnum/problems/_problems.py
+++ b/src/probnum/problems/_problems.py
@@ -15,10 +15,29 @@ import probnum
 
 @dataclasses.dataclass
 class RegressionProblem:
-    r"""Regression problems.
+    r"""Regression problem.
 
     Fit a stochastic process to data, given a likelihood (realised by a :obj:`Transition`).
     Solved by Kalman filtering and smoothing in :mod:`probnum.filtsmooth`.
+
+    Parameters
+    ----------
+    observations
+        Observations of the latent process.
+    locations
+        Grid-points on which the observations were taken.
+    likelihood
+        Likelihood of the observations; that is, relation between the latent process and the observed values. Encodes for example noise.
+
+    Examples
+    --------
+    >>> obs = [11.4123, -15.5123]
+    >>> loc = [0.1, 0.2]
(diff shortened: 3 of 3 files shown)


setup.cfg
```diff
@@ -38,6 +38,7 @@ install_requires =
     numpy; python_version>"3.6"
     scipy>=1.4
     cached_property; python_version<"3.8"
+    dataclasses; python_version<"3.7"
```
pnkraemer, line 41: Dataclasses need python >= 3.7. The import is always `import dataclasses`.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+
+@dataclasses.dataclass
+class RegressionProblem:
+    r"""Regression problem.
+
+    Fit a stochastic process to data, given a likelihood (realised by a :obj:`Transition`).
+    Solved by Kalman filtering and smoothing in :mod:`probnum.filtsmooth`.
+
+    Parameters
+    ----------
```
pnkraemer, line 24: I think they should be documented as parameters, not as attributes. I am not 100% sure though.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+    >>> ivp = IVProblem(f, t0=0., tmax=3., y0=0.1)
+    >>> ivp.t0, ivp.tmax, ivp.y0
+    (0.0, 3.0, 0.1)
+    >>> np.round(ivp.f(ivp.t0, ivp.y0), 2)
+    0.09
+    """
+
+    f: typing.Callable[[float, np.ndarray], np.ndarray]
+    t0: float
+    tmax: float  # Bold move: remove this? This is not really part of an IVP.
```
pnkraemer, line 85: @nathanaelbosch usually tmax is in such definitions. Though, isn't it more elegant to not have it in here? What do you think?

pnkraemer: Oh btw, this will replace the `IVP` object soon. It does the same, but is so much less code.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+    >>> def f(t, x):
+    ...     return x*(1-x)
+    >>> ivp = IVProblem(f, t0=0., tmax=3., y0=0.1)
+    >>> ivp.t0, ivp.tmax, ivp.y0
+    (0.0, 3.0, 0.1)
+    >>> np.round(ivp.f(ivp.t0, ivp.y0), 2)
+    0.09
+    """
+
+    f: typing.Callable[[float, np.ndarray], np.ndarray]
```
pnkraemer, line 83: I am going for this single letter thing here. an alternative would be `ode`, but `f` should be super clear to everyone I guess. Calling `ivp.ode` or `ivp.rhs` is pretty cumbersome and got on my nerves in the past.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+    t0: float
+    tmax: float  # Bold move: remove this? This is not really part of an IVP.
+    y0: typing.Union[probnum.type.FloatArgType, np.ndarray]
```
pnkraemer, line 86: RV init conds are not part of the problems we can solve (reliably). So I removed them here and once we know how to deal with this, we can make corresponding problems. Also, always calling `initrv.mean` seems unnecessarily heavyweight to me.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+           [0., 0., 1.]]), b=array([0, 1, 2]))
+    """
+
+    A: typing.Union[
+        np.ndarray,
+        "scipy.sparse.spmatrix",
+        "probnum.linops.LinearOperator",
+        "probnum.RandomVariable",
+    ]
+    b: typing.Union[np.ndarray, "probnum.RandomVariable"]
```
pnkraemer, line 120: @JonathanWenger I copied the type hints from `problinsolve`. Does this still fit your scope?

JonathanWenger: Yes it should.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+class QuadProblem:
+    r"""Numerical computation of an integral.
+
+    Compute the integral
+
+        .. math::
+            \int_\Omega f(x) \, \text{d} \mu(x)
+
+    for a function :math:`f: \Omega \rightarrow \mathbb{R}`.
+    For the time being, :math:`\mu` is the Lebesgue measure.
```
pnkraemer, line 139: @alpiges I guess you could do more, but I don't know how to realise this in code. So for now, there is no measure in the fields of a `QuadProblem`.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+        np.ndarray,
+        "scipy.sparse.spmatrix",
+        "probnum.linops.LinearOperator",
+        "probnum.RandomVariable",
+    ]
+    b: typing.Union[np.ndarray, "probnum.RandomVariable"]
+
+
+@dataclasses.dataclass
+class QuadProblem:
```
pnkraemer, line 130: `QuadratureProblem` seemed too long for me.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
...
+    ...     return 0 < x < 1
+    >>> qp = QuadProblem(integrand, domain)
+    >>> np.round(qp.integrand(0.2), 2)
+    0.04
+    >>> qp.domain(0.2)
+    True
+    """
+
+    integrand: typing.Callable[[np.ndarray], np.ndarray]
+    domain: typing.Callable[[np.ndarray], bool]  # Up for discussion...
```
pnkraemer, line 163: @alpiges this is only a draft -- I don't know how you usually pick domains in code.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,157 @@
-"""Definitions of problems solved by probabilistic numerical methods."""
+"""Definitions of problems currently solved by probabilistic numerical methods.
+
+If a problem is not listed here, we have not implemented a probabilistic
+numerical method for it yet.
+"""
+
+import dataclasses
```
pnkraemer, line 7: the import is the same for all python versions (even though for 3.6 we need the additional dependency `import dataclasses` -- see `setup.cfg`).

src/probnum/problems/_problems.py
```diff
@@ -1 +1,163 @@
...
+    >>> np.round(ivp.f(ivp.t0, ivp.y0), 2)
+    0.09
+    """
+
+    f: typing.Callable[[float, np.ndarray], np.ndarray]
+    t0: float
+    tmax: float  # Bold move: remove this? This is not really part of an IVP.
+    y0: typing.Union[probnum.type.FloatArgType, np.ndarray]
+    df: typing.Callable[[float, np.ndarray], np.ndarray] = None
+    ddf: typing.Callable[[float, np.ndarray], np.ndarray] = None
```
pnkraemer, line 92: again, (almost) single letters -- I find them sufficiently descriptive here. What do you think?

src/probnum/problems/_problems.py
pnkraemer: An alternative would be something similar to (https://github.com/SciML/Quadrature.jl  --  see bottom of the readme). Let me know if this is more in line with your problem types [or change it yourself in the future :)]

pnkraemer:
If the CI passes, this should do it. As soon as this happens, I will request reviews.

This PR is an attempted first step towards streamlining PN algorithms in `ProbNum`. By defining specific problems that we can solve (and subsequently communicating problems we can not solve) we can more clearly communicate with a user. If all PN algorithms, no matter if linalg, quad, or filtsmooth, act on such problem instances it is easy for a user to use all of the algorithms: know one, know them all.

I guess there are some more advantages to implementing base problems, who knows. This is inspired by what happens for instance in DiffeqBase (SciML; https://github.com/SciML/DiffEqBase.jl/tree/master/src).

src/probnum/problems/_problems.py
```diff
@@ -1 +1,163 @@
...
+
+
+@dataclasses.dataclass
+class LinearSystemProblem:
```
JonathanWenger, line 98: I suggest LinearSystem here. This breaks with the *Problem convention, but there is no such thing as a linear system problem.

src/probnum/problems/_problems.py
JonathanWenger: I'd argue for QuadratureProblem or IntegrationProblem. `QuadProblem` is somewhat nondescriptive and `RegressionProblem` has the same number of letter as `QuadratureProblem`. In particular `QuadProblem` may be confused with quadratic program.

src/probnum/problems/__init__.py
```diff
@@ -1 +1,5 @@
 """Definitions and collection of problems solved by probabilistic numerical methods."""
+
+from ._problems import *
```
JonathanWenger, line 3: Star imports are discouraged since they potentially import unwanted modules, functions or classes. Instead explicitly write out the classes and functions you are importing. For example currently `probnum.diffeq.filtsmooth` exports numpy as `probnum.diffeq.filtsmooth.np`:

```python
dir(probnum.diffeq.odefiltsmooth)
Out[4]: 
['GaussianIVPFilter',
 '__builtins__',
 '__cached__',
...
 'ivp2filter',
 'np',
 'pnfs',
 ...
```

src/probnum/problems/_problems.py
```diff
@@ -1 +1,165 @@
...
+    >>> lsp = LinearSystemProblem(A, b)
+    >>> lsp
+    LinearSystemProblem(A=array([[1., 0., 0.],
+           [0., 1., 0.],
+           [0., 0., 1.]]), b=array([0, 1, 2]))
+    """
+
+    A: typing.Union[
+        np.ndarray,
+        "scipy.sparse.spmatrix",
```
JonathanWenger, line 124: Currently the docs build is failing since sphinx cant resolve the type hint. An import of `scipy.sparse` should fix this.

src/probnum/problems/_problems.py
alpiges: I think the domain should be defined as box constraints (lower and upper bound for each dimension of x). Although there could be a domain method which checks whether x is in the domain. Just -- why?
Also the example domain will not work in multiple dimensions.

src/probnum/problems/_problems.py
alpiges: I agree.

src/probnum/problems/_problems.py
pnkraemer: > I think the domain should be defined as box constraints

Thanks a lot for the feedback, I will change it! :) 

> Just -- why?

I thought box constraints only work for rectangular (do I mean "hypercubic" or something like this?) domains. What about triangles, circles, L-shaped, domains, etc.? (see for instance https://github.com/nschloe/quadpy#line-segment-c1) 

>  Also the example domain will not work in multiple dimensions.

I agree. It is just a simple example... :)

src/probnum/problems/_problems.py
pnkraemer: alright! 

> but there is no such thing as a linear system problem.

... though one could probably argue about this ;)

src/probnum/problems/_problems.py
pnkraemer: Alright, this is easy to change

src/probnum/problems/_problems.py
alpiges: > I thought box constraints only work for rectangular (do I mean "hypercubic" or something like this?) domains. What about triangles, circles, L-shaped, domains, etc.? (see for instance https://github.com/nschloe/quadpy#line-segment-c1)
Yes, I agree, just that's pretty far on the horizon. As long as we can to kernel integrals in closed form, we need the explicit bounds of the domain.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,202 @@
...
+    Parameters
+    ----------
+    observations
+        Observations of the latent process.
+    locations
+        Grid-points on which the observations were taken.
+    likelihood
+        Likelihood of the observations; that is, relation between the latent process and the observed values.
+        Encodes for example noise.
+    solution
```
pnkraemer, line 35: I don't really have a strong opinion on naming. I guess alternatives would be "true_solution", "analytical" (which is how SciML does it), "analytical_solution", "sol", or others. If you think one of them is strictly better than others, please voice your opinion. I guess the only thing to take into account is to communicate that calling "problem.solution" will **not** solve your problem, therefore something overly specific or even tedious to call might not be a bad idea.

pnkraemer:
Alright guys, I guess it is ready for review now. I am assigning you **all**, because it affects everyone and I think we should get this right at this moment. Though, feel free to only review the problem type that affects your own line of work. Everything I am not 100% sure about is marked with a comment. Please feel free to comment other things, too.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
-"""Definitions of problems solved by probabilistic numerical methods."""
+"""Definitions of problems currently solved by probabilistic numerical methods.
+
+If a problem is not listed here, we have not implemented a probabilistic
```
JonathanWenger, line 3: Petty opinion: This comment seems obsolete to me or at least I would change it to passive voice, i.e. ..., there is no implemented probabilistic numerical method for it yet.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+import probnum.random_variables as pnrv
+import probnum.type as pntp
+
+
+@dataclasses.dataclass
+class RegressionProblem:
+    r"""Regression problem.
+
+    Fit a stochastic process to data, given a likelihood (realised by a :obj:`DiscreteGaussian` transition).
+    Solved by Kalman filtering and smoothing in :mod:`probnum.filtsmooth`.
```
JonathanWenger, line 24: Isn't RegressionProblem too broad here? There is at least a Markovian assumption if this is solved by Kalman filtering, right? I would recommend making either the docstring more general or changing the name.

src/probnum/problems/_problems.py
JonathanWenger: I think `solution` is a good general naming. If `solution=None` and therefore not set, then one has to call `ProbabilisticNumericalMethod.solve(problem)` to obtain an approximate one.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+    # For testing and benchmarking
+    solution: typing.Optional[typing.Callable[[float, np.ndarray], np.ndarray]] = None
+
+
+@dataclasses.dataclass
+class LinearSystem:
+    r"""Linear system of equations.
+
+    Compute :math:`x` from :math:`Ax=b`.
+    Solved by the probabilistic linear solver in :mod:`probnum.linalg`
```
JonathanWenger, line 118: suggestion: "solved by probabilistic linear solvers in..."

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+
+    Parameters
+    ----------
+    A
+        Square matrix or linear operator.
+    b
+        Right-hand side vector or matrix.
+    solution
+        Closed form, analytic solution to the problem. Used for testing and benchmarking.
```
JonathanWenger, line 127: For linear systems "closed form" doesn't really make sense. I think "True solution" or "Analytic solution" is a better naming for `LinearSystem`.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+class LinearSystem:
+    r"""Linear system of equations.
+
+    Compute :math:`x` from :math:`Ax=b`.
+    Solved by the probabilistic linear solver in :mod:`probnum.linalg`
+
+    Parameters
+    ----------
+    A
+        Square matrix or linear operator.
```
JonathanWenger, line 123: In the future we might have solvers for non-square problems (theory for it exists) so I would suggest just "System matrix or linear operator."

JonathanWenger (approved):
Great stuff! I really like the clarity and structure the Problem classes add. Thanks :grin:

src/probnum/problems/_problems.py
nathanaelbosch: I don't really see why tmax should not be in there, it seems like a strict part of the problem definition for me. Though I could imagine a `timespan` argument, which is a tuple `(t0, tmax)` to make the function signature a bit compacter. The current api is completely fine too though.

nathanaelbosch (approved):

src/probnum/problems/_problems.py
pnkraemer: No, in fact it is not (at least that's what I hope). This should include any GP regression problems. I see that it might not be clear, because the likelihood is defined by something in `statespace`, but that is just a sub-package artefact (and one of the reasons I want to pull up statespace and rename it). In 1d, with markovian structure, Kalman filters solve it. The problem definition is more general than that though.

src/probnum/problems/_problems.py
pnkraemer: I tend to think that only the initial time point is relevant for the problem definition. Only as soon as one starts to think about solutions, the final time point comes into play (existence of solutions, etc.). I see that removing it would be quite an exotic choice, though, so I think putting it in is fine. 

> Though I could imagine a timespan argument, which is a tuple (t0, tmax) to make the function signature a bit compacter. The current api is completely fine too though.

I could add a property that returns this? Or not do it for now, I mean, it is not a gamechanger is it

src/probnum/problems/_problems.py
pnkraemer: Alright (I just copied the docstring in `linearsolver.py`.

src/probnum/problems/_problems.py
pnkraemer: Let's try to put this information everywhere then, by perhaps renaming the `solution` field into `analytical_solution`? I find it okay if it is cumbersome to call, we will only need it rarely (and mostly internally).

src/probnum/problems/_problems.py
nathanaelbosch: I don't think a property for this is really useful, if anything I'd just prefer it for the creation of the problem. It just feels a bit more balanced, since that way each "component' of the problem gets 1 slot (vector field, initial value, time domain), but again it's quite a minor thing so I think the current implementation is also fine :)

pnkraemer: In this case I would leave it as is: I believe calling `ivp.t0, ivp.tmax` is more descriptive than `ivp.timespan[0], ivp.timespan[1]`. If you disagree, voice your opinion, but it seems to me that you dont have a super strong opinion either :)

src/probnum/problems/_problems.py
JonathanWenger: I think `solution` is good, it's just the docstring that doesn't match here. Easy fix would be to just drop "Closed form" from the docstring.

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+    likelihood: typing.Optional[pnfs.statespace.DiscreteGaussian] = None
+
+    # For testing and benchmarking
+    solution: typing.Optional[
+        typing.Callable[[np.ndarray], typing.Union[float, np.ndarray]]
+    ] = None
+
+
+@dataclasses.dataclass
+class IVProblem:
```
alpiges, line 60: Wouldn't it make more sense to write this out `InitialValueProblem`? I feel like this would be more in line with the other problem classes, but also I don't have a strong opinion.

src/probnum/problems/_problems.py
alpiges: I think this is fine for now, we can add measures later once they are in `probnum.quad`

JonathanWenger (approved):

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+    0.41
+    >>>
+    >>> lower_bd = [0., 0.]
+    >>> upper_bd = [1., 1.]
+    >>> qp2d = QuadratureProblem(integrand, lower_bd=lower_bd, upper_bd=upper_bd)
+    >>> qp2d.upper_bd
+    [1.0, 1.0]
+    """
+
+    integrand: typing.Callable[[np.ndarray], typing.Union[float, np.ndarray]]
```
alpiges, line 196: f could also be a `RandomProcess`?

src/probnum/problems/_problems.py
```diff
@@ -1 +1,204 @@
...
+    lower_bd: typing.Union[pntp.FloatArgType, np.ndarray]
+    upper_bd: typing.Union[pntp.FloatArgType, np.ndarray]
+    output_dim: typing.Optional[int] = 1
+
+    # For testing and benchmarking
+    solution: typing.Optional[typing.Union[float, np.ndarray]] = None
```
alpiges, line 202: `solution` could also be a `RandomVariable`

alpiges (approved):

src/probnum/problems/_problems.py
pnkraemer: If that is something you intend to cover with BQ, sure! We can update the type hint as soon as we have random processes!

src/probnum/problems/_problems.py
pnkraemer: If this is something that you want BQ to cover, I will update :)

src/probnum/problems/_problems.py
pnkraemer: Turns out that both nathanael and I agree with you :) Thanks for the comment!

src/probnum/problems/_problems.py
alpiges: Yes, that should go be part of BQ for sure!

src/probnum/problems/_problems.py
alpiges: Maybe leave the `RandomProcess` part out for now, still can add this later

PR #284, opened 2020-12-14
Remove star imports from diffeq and filtsmooth (and filtsmooth.statespace)

pnkraemer:
The name says it all.

No more star imports in `__init__.py` files (in diffeq and filtsmooth). This avoids weird exports of `statespace.numpy` and the like.

### DIFF
--- a/src/probnum/diffeq/ode/__init__.py
+++ b/src/probnum/diffeq/ode/__init__.py
@@ -1,6 +1,14 @@
-from .ivp import *
-from .ivp_examples import *
-from .ode import *
+from .ivp import IVP
+from .ivp_examples import (
+    fitzhughnagumo,
+    logistic,
+    lotkavolterra,
+    rigidbody,
+    seir,
+    threebody,
+    vanderpol,
+)
+from .ode import ODE
 
 __all__ = [
     "ODE",
--- a/src/probnum/filtsmooth/__init__.py
+++ b/src/probnum/filtsmooth/__init__.py
@@ -1,8 +1,19 @@
 """Bayesian Filtering and Smoothing."""
 
-from .bayesfiltsmooth import *
-from .filtsmoothposterior import *
-from .gaussfiltsmooth import *
+from .bayesfiltsmooth import BayesFiltSmooth
+from .filtsmoothposterior import FiltSmoothPosterior
+from .gaussfiltsmooth import (
+    ContinuousEKFComponent,
+    ContinuousUKFComponent,
+    DiscreteEKFComponent,
+    DiscreteUKFComponent,
+    FixedPointStopping,
+    IteratedKalman,
+    Kalman,
+    KalmanPosterior,
+    StoppingCriterion,
+    UnscentedTransform,
+)
 
 # Public classes and functions. Order is reflected in documentation.
 __all__ = [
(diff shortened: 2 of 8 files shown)


src/probnum/diffeq/ode/__init__.py
```diff
@@ -1,6 +1,14 @@
-from .ivp import *
-from .ivp_examples import *
-from .ode import *
+from .ivp import IVP
+from .ivp_examples import (
```
schmidtjonathan, line 2: You do not import the `*_rhs()` methods from `ivp_examples`. I personally find that desirable, I just wanted to check whether that was a deliberate choice.

src/probnum/filtsmooth/__init__.py
```diff
@@ -1,8 +1,19 @@
...
+    ContinuousUKFComponent,
+    DiscreteEKFComponent,
+    DiscreteUKFComponent,
+    FixedPointStopping,
+    IteratedKalman,
+    Kalman,
+    KalmanPosterior,
+    StoppingCriterion,
+    UnscentedTransform,
+)
```
schmidtjonathan, line 17: Shouldn't we also import the `statespace` module here?

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         self.dispmatfun = dispmatfun
         self.jacobfun = jacobfun
         self.precon = None
```
schmidtjonathan, line 31: `self.precon = None` is already set in `Transition.__init__()`. I suggest removing this line and moving the `super().__init__()` call to the top of `SDE`'s initializer.

src/probnum/filtsmooth/__init__.py
```diff
@@ -1,8 +1,19 @@
 """Bayesian Filtering and Smoothing."""
 
-from .bayesfiltsmooth import *
-from .filtsmoothposterior import *
-from .gaussfiltsmooth import *
+from .bayesfiltsmooth import BayesFiltSmooth
+from .filtsmoothposterior import FiltSmoothPosterior
+from .gaussfiltsmooth import (
```
schmidtjonathan, line 5: Actually, I think you forgot to replace the `*`-imports in `filtsmooth.gaussfiltsmooth.__init__.py`

schmidtjonathan (changes requested):
That's a lot more clean and concise, thanks for taking care of this! I added a few comments, most of which are minor. The imports from `filtsmooth.gausfiltsmooth.__init__.py` are probably most important.

src/probnum/diffeq/ode/__init__.py
pnkraemer: This was a deliberate choice

src/probnum/filtsmooth/__init__.py
pnkraemer: is this how it works? I thought having it separate would be sufficient...

src/probnum/filtsmooth/__init__.py
pnkraemer: hmmmmm seems it. I thought I had... thanks for the pointer!

schmidtjonathan (approved):
:+1:

src/probnum/filtsmooth/__init__.py
schmidtjonathan: Before merging, you might want to reconsider that. I think, it works as it is, but I see no reason why in the `filtsmooth.__init__.py` we should explicitly import `gaussfiltsmooth` but not `statespace`.

schmidtjonathan:
I added one more comment you might want to consider [here](https://github.com/probabilistic-numerics/probnum/pull/284#discussion_r543533797)

PR #286, opened 2020-12-17
Diffconst as `transition` argument instead of `Integrator` property

nathanaelbosch:
WIP! The description will be edited over the course of the development.

I also tried to make my commits as meaningful as possible so you can also have a look at them individually, but what I did was basically:
1. I removed `diffconst` from all `Integrator`s, completely, and temporarily "fix" the tests (`tests/test_filtsmooth/test_statespace/test_integrator.py)` by setting `DIFFCONST = 1.0`
2. Added a `diffusion` parameter to all "leaves", ie. functions which do not just pass this diffusion parameter down to some other function. In particular, these seem to be: (i) `DiscreteGaussian.transition_realization`, (ii) `DiscreteLinearGaussian.transition_rv`, and (iii) `sde._increment_fun`. _Please check that I did not miss any!_
3. Added a `diffusion` parameter to all `statespace` components, as well as `Kalman`, the EKF-components and the UKF-components (but not yet IEKS).
4. Changed the tests again in order to pass a specific `diffusion` parameters to the transitions, and they still pass :) 

Questions / things left to do:
- [x] How should the diffusion be properly implemented for the UnscentedKalman?
- [x] I didn't yet add something for the IEKS
- [x] I guess I should add additional tests? I'm not sure what they should do, any pointers?


src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -17,7 +17,9 @@ def __init__(self, non_linear_sde, num_steps):
         self.num_steps = num_steps
         super().__init__()
 
-    def transition_realization(self, real, start, stop, linearise_at=None, **kwargs):
+    def transition_realization(
+        self, real, start, stop, linearise_at=None, diffusion=1.0, **kwargs
```
nathanaelbosch, line 21: Do you have any preferences for the ordering of parameters? 

Edit: Looking into at it now, I would maybe move the diffusion parameter before `linearise_at` since it appears in all transition functions.

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -67,16 +73,18 @@ def __init__(self, disc_model):
...
 
-    def transition_rv(self, rv, start, linearise_at=None, **kwargs):
+    def transition_rv(self, rv, start, linearise_at=None, diffusion=1.0, **kwargs):
         diffmat = self.disc_model.diffmatfun(start)
         compute_jacobian_at = linearise_at.mean if linearise_at is not None else rv.mean
         jacob = self.disc_model.jacobfun(start, compute_jacobian_at)
         mpred = self.disc_model.dynamicsfun(start, rv.mean)
         crosscov = rv.cov @ jacob.T
-        cpred = jacob @ crosscov + diffmat
+        cpred = jacob @ crosscov + diffusion * diffmat
```
nathanaelbosch, line 87: This is one of the "leaves" where the parameter is used, not just passed.

src/probnum/filtsmooth/statespace/discrete_transition.py
```diff
@@ -48,10 +48,10 @@ def if_no_jacobian(t, x):
         self.jacobfun = jacobfun if jacobfun is not None else if_no_jacobian
         super().__init__()
 
-    def transition_realization(self, real, start, **kwargs):
+    def transition_realization(self, real, start, diffusion=1.0, **kwargs):
 
         newmean = self.dynamicsfun(start, real)
-        newcov = self.diffmatfun(start)
+        newcov = diffusion * self.diffmatfun(start)
```
nathanaelbosch, line 54: Another one of the "leaves" here, and below in `transition_rv`

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -85,7 +83,7 @@ def _forcevec(self):
     @property
     def _dispmat(self):
         dispmat_1d = np.zeros(self.ordint + 1)
-        dispmat_1d[-1] = self.diffconst
+        dispmat_1d[-1] = 1.0  # Standard diffusion
```
nathanaelbosch, line 86: Should I remove the comment?

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -301,7 +322,9 @@ def _increment_fun(time, mean, cov, driftfun, jacobfun, dispmatfun):
     jacobian = jacobfun(time)
     mean_increment = driftfun(time, mean)
     cov_increment = (
-        cov @ jacobian.T + jacobian @ cov.T + dispersion_matrix @ dispersion_matrix.T
+        cov @ jacobian.T
+        + jacobian @ cov.T
+        + diffusion * dispersion_matrix @ dispersion_matrix.T
```
nathanaelbosch, line 327: The last "leaf". It should be correct this way since diffusion = sigma^2, but please double-check.

tests/test_filtsmooth/test_statespace/test_integrator.py
```diff
@@ -30,7 +30,7 @@ def test_proj2coord(self):
 
 
 STEP = np.random.rand()
-DIFFCONST = np.random.rand()
+DIFFCONST = np.random.rand() ** 2
```
nathanaelbosch, line 33: "diffusion" means sigma^2 now

tests/test_filtsmooth/test_statespace/test_integrator.py
```diff
@@ -57,7 +57,7 @@ def test_proj2coord(self):
 
 class TestIBM(unittest.TestCase, NumpyAssertions):
     def setUp(self):
-        self.sde = pnfs.statespace.IBM(ordint=2, spatialdim=2, diffconst=DIFFCONST)
+        self.sde = pnfs.statespace.IBM(ordint=2, spatialdim=2)
```
nathanaelbosch, line 60: In all these tests I basically changed the same stuff: I removed the diffusion from the constructor, and instead added the same diffusion into all `transition` functions that are called in the tests

nathanaelbosch:
@pnkraemer I think you can have a look already, most things should be there :) In addition to the general review, I would particularly appreciate your input for the UKF and for testing.

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: I think I agree. 

Two minor questions here: 1. do we want a default set to 1 or a None (for documentation/communication reasons I think this is a valid question); related: 2. do we want to name it "diffusion" or something like "special_diffusion" to communicate that this should only be touched if one knows what is going on?

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: nice! I am not sure though, whether we want both a diffusion and a diffmat... I find this hard to parse, looking at the code...

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -75,7 +75,9 @@ def filter(self, dataset, times, intermediate_step=None, linearise_at=None):
             rvs.append(filtrv)
         return KalmanPosterior(times, rvs, self, with_smoothing=False)
 
-    def filter_step(self, start, stop, current_rv, data, intermediate_step=None):
+    def filter_step(
+        self, start, stop, current_rv, data, intermediate_step=None, diffusion=1.0
```
pnkraemer, line 79: so the filter() method gets a diffusion constant? Why exactly? Isnt this something we want to set to 1 all the time and calibrate secretly, instead of asking the user to specify it at filter time?

src/probnum/filtsmooth/statespace/integrator.py
pnkraemer: No I think it is good, at least for the next few weeks/months until we got used to the interface.

pnkraemer: Though maybe call it unit diffusion instead of standard diffusion? Just a suggestion, feel free to ignore

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
nathanaelbosch: I did it like this just to touch on the least amount of code. I guess the alternative would be to rename `diffmat` and `diffmatfun` into something else?

src/probnum/filtsmooth/statespace/sde.py
pnkraemer: I think this is correct

src/probnum/filtsmooth/statespace/transition.py
```diff
@@ -149,6 +154,8 @@ def transition_rv(
         linearise_at :
             For approximate transitions , for instance ContinuousEKFComponent,
             this argument overloads the state at which the Jacobian is computed.
+        diffusion :
+            Diffusion (or diffusion estimate) for this transition
```
pnkraemer, line 158: maybe mention that this field is usually used to update diffusions with calibrated versions thereof?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: Good thing you comment on this! To be honest, I was not really sure in which parts of Kalman you want to have the diffusion, so I added it to `predict`, `smooth_step` (and related) and `filter_step`. I guess we do need it for the smoothing, right? Therefore it felt natural to also add it into `filter_step`, but I don't really know.

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: I agree, both with the fact that it is good to touch the least amount of code, and also with the sentiment that diffmat might be changed instead of diffusion...

pnkraemer:
That was quick! Thanks! 

## UKF
About the UKF: let's look at `DiscreteUKFComponent.transition_rv`:
```python
    def transition_rv(self, rv, start, linearise_at=None, **kwargs):
        compute_sigmapts_at = linearise_at if linearise_at is not None else rv
        sigmapts = self.ut.sigma_points(
            compute_sigmapts_at.mean, compute_sigmapts_at.cov
        )
        proppts = self.ut.propagate(start, sigmapts, self.disc_model.dynamicsfun)
        meascov = self.disc_model.diffmatfun(start)
        mean, cov, crosscov = self.ut.estimate_statistics(
            proppts, sigmapts, meascov, rv.mean
        )
        return Normal(mean, cov), {"crosscov": crosscov}
```
I think the line ` meascov = self.disc_model.diffmatfun(start)` needs to be turned into `meascov = diffusion * self.disc_model.diffmatfun(start)`. What do you think?

## IEKS
I think making sure that in IteratedKalman.predict and IteratedKalman.update and IteratedKalman.filter_step, a diffusion parameter is possible, will do the trick.


## Tests
I think we can remove a bunch of stuff (i.e. all the diffconst calls). Then, I suppose it is wise to test the diffusion leaves for actually using the diffusion (you have a better overview than me which those are), e.g. by plugging 1000 in there and checking that the outcome reflects this. Other than that, maybe we only check this argument wherever there is a bugfix, e.g., we could check that the KalmanPosterior handles diffusions. That could be a future PR though.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: I think that this is wise, but I suppose on filter(), smooth() and filtsmooth() we do not want it (cp. with ODESolver.solve(), where we probably dont want it either). Instead, in those functions, I'd argue that it is the calibration's job to handle sigma. What do you think?

nathanaelbosch: What do you mean with "it is the calibration's job to handle the sigma"? I don't really understand where we want to get / how we want to pass the diffusions down, if not by function arguments.

pnkraemer: Everything I meant was that I dont think it is a good idea to allow a user to change  the sigma behaviour while calling filter() or filtsmooth(), because we want to nudge a user towards "trusting" the calibration. Is that more clear? Sorry, I am aware this might not be the most clear explanation. if you have more questions, we can talk in person?

nathanaelbosch: Just to note what we talked about on Friday: `filter_step`, `predict` and `smooth_step` should indeed have a diffusion parameter, whereas `filter`, `smooth`, `smooth_list` and `filtsmooth` should not, so I would leave things as they were. Though if I misunderstood something or if you disagree please re-open this and I'll change things :)

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
nathanaelbosch: Intuitively I would name it `diffusion` or maybe `_diffusion` and set it to `1.0`, since that makes it explicit what's actually happening. I mean it _is_ the diffusion, and we set it to value 1?

Though I understand your motivation for having a different name to make clear that that's a very custom functionality that most people should not care about, and in that line of thought I also see the appeal of setting it to `None`.

So, what do you think?

nathanaelbosch: To brainstorm names a bit: `custom_diffusion`? `internal_diffusion`? `diffusion_coefficient`/`diffcoeff`?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -295,10 +309,17 @@ def smooth_step(
...
             )
 
     def _smooth_step_classic(
-        self, unsmoothed_rv, smoothed_rv, start, stop, intermediate_step=None
+        self,
+        unsmoothed_rv,
+        smoothed_rv,
+        start,
+        stop,
+        intermediate_step=None,
```
nathanaelbosch, line 321: Any preferences on the parameter order in `Kalman`: `filter_step`, `smooth_step` etc?

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: let's go for `_diffusion` (and likewise, `_intermediate_step` and perhaps also `_linearise_at`)? A default to 1 is fine. It keeps the code simple.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: Fine

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: not really. I think the "internal ones" (i.e. intermediate step, diffusion, etc.) should go behind the standard ones (start, stop, etc) but other than that, just pop new ones to the end of the list? I dont think this makes a big difference

tests/test_filtsmooth/test_statespace/test_integrator.py
pnkraemer: Thanks for the comment. Lets be careful with the fact that this sentiment should be reflected everywhere.

pnkraemer:
From my POV all the discussions are finished. What is the state of the PR now? Are there todos left? I have no real overview here...

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
nathanaelbosch: Ok, I renamed all three of them to have leading underscores.

nathanaelbosch:
> From my POV all the discussions are finished. What is the state of the PR now? Are there todos left? I have no real overview here...

The thing left seems to be the IEKS. I'll look into that now.

nathanaelbosch:
The IEKS `filter_step` and `predict` now support a `_diffusion` parameter :+1:. I think that should conclude the PR!

pnkraemer:
Looks good, thanks! Just to refresh my memory: this PR is just about changing the "input" of diffconst, right? Its usage is restricted to ODE solvers (and there, only the forward passes?)? The "bug" about the posteriors is thus still in there I suppose...

nathanaelbosch:
> Looks good, thanks! Just to refresh my memory: this PR is just about changing the "input" of diffconst, right? Its usage is restricted to ODE solvers (and there, only the forward passes?)? The "bug" about the posteriors is thus still in there I suppose...

So if I remember things correctly, the goal here was just to make the methods "aware" of the diffusion. The next step would be to track the diffusion (in the solver, solution object, or KalmanPosterior), and to use the tracked diffusions for interpolation and sampling in the KalmanPosterior. Does that sound right?

pnkraemer:
That sounds great. Thanks for the update. So would the next step be to assemble the KalmanPosterior and ODESolution objects on-the-fly? Either way, we can discuss this offline as it does not have anything to do with this PR here.

pnkraemer (approved):

PR #287, opened 2020-12-18
Abstrahised ODESolution object

pnkraemer:
As discussed, the ODESolution object is only an interface now. That simplifies the implementation for ODE filters quite drastically.


src/probnum/diffeq/__init__.py
```diff
@@ -11,7 +11,14 @@
     threebody,
     vanderpol,
 )
-from .odefiltsmooth import GaussianIVPFilter, ivp2ekf0, ivp2ekf1, ivp2ukf, probsolve_ivp
+from .odefiltsmooth import (
+    GaussianIVPFilter,
+    KalmanODESolution,
```
pnkraemer, line 16: I am naming it `KalmanODESolution` to emphasise that it does not really depend on the ODE filter anymore, but behaves like any Gauss-Markov posterior.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -2,12 +2,13 @@
...
 from probnum.diffeq import odesolver
-from probnum.diffeq.odesolution import ODESolution
 from probnum.random_variables import Normal
 
+from .kalman_odesolution import KalmanODESolution
+
 
 class GaussianIVPFilter(odesolver.ODESolver):
-    """ODE solver that behaves like a Gaussian filter.
+    """ODE solver that uses a Gaussian filter.
```
pnkraemer, line 11: feel free to disagree ;)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -75,13 +76,21 @@ def step(self, t, t_new, current_rv):
 
         return filt_rv, err
 
-    def postprocess(self, times, rvs):
+    def rvlist_to_odesol(self, times, rvs):
```
pnkraemer, line 79: this is now an abstract method in the base class, to enforce that every ODE solver implementation considers creation of ODESolutions.
It partly replaces the previous `postprocess` method (which still exists, but is empty by default).

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -75,13 +76,20 @@ def step(self, t, t_new, current_rv):
...
+        """Create an ODESolution object."""
+
+        kalman_posterior = pnfs.KalmanPosterior(
+            times, rvs, self.gfilt, self.with_smoothing
+        )
+
+        return KalmanODESolution(kalman_posterior)
+
+    def postprocess(self, odesol):
+        """If specified (at initialisation), smooth the filter output."""
```
pnkraemer, line 89: there is no rescaling anymore.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -0,0 +1,120 @@
...
+    >>> initrv = rvs.Constant(0.15)
+    >>> ivp = logistic(timespan=[0., 1.5], initrv=initrv, params=(4, 1))
+    >>> solution = probsolve_ivp(ivp, method="ekf0", step=0.1)
+    >>> # Mean of the discrete-time solution
+    >>> print(solution.y.mean)
+    [[0.15      ]
+     [0.2076198 ]
+     [0.27932997]
+     [0.3649165 ]
+     [0.46054129]
```
pnkraemer, line 45: this doctest has not been touched -- it passes in the same way (so do all the other tests). It is just an internal change, the API is identical to before.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -0,0 +1,120 @@
...
+        # The prior must be an integrator, if not, an error is thrown in 'GaussianIVPFilter'.
+        self.proj_to_y = kalman_posterior.gauss_filter.dynamics_model.proj2coord(
+            coord=0
+        )
+        self.proj_to_dy = kalman_posterior.gauss_filter.dynamics_model.proj2coord(
+            coord=1
+        )
+
+    @property
+    def t(self) -> np.ndarray:
```
pnkraemer, line 83: the docstrings here and below are written in the base class.

src/probnum/diffeq/odesolution.py
```diff
@@ -1,176 +1,70 @@
...
-    def _state_rvs(self):
-        """:obj:`list` of :obj:`RandomVariable`:"""
-        return self._kalman_posterior.state_rvs
-
-    def __call__(self, t):
+    def dy(self) -> pnrv_list._RandomVariableList:
+        """First derivative of the discrete-time solution."""
+        raise NotImplementedError("The first derivative has not been implemented")
+
+    # Not abstract, because providing interpolation could sometimes be tedious.
```
pnkraemer, line 39: this comment. 

I think it can be annoying to enforce dense output. I find it least stressful to only enforce lists of random variables.

src/probnum/filtsmooth/statespace/transition.py
```diff
@@ -246,6 +246,6 @@ def _read_dimension(transition, initrv):
     # relies on evaluating at zero, which is a dangerous endeavour and therefore,
     # this method is not used in Transition.dimension
     transitioned, _ = transition.transition_realization(
-        real=initrv.mean, start=0.0, stop=1.0
+        real=initrv.mean, start=0.0, stop=1.0, step=1.0
```
pnkraemer, line 249: that was a bugfix I encountered while working on the bridges.

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -131,7 +131,7 @@ def setUp(self):
         self.ivp = ode.logistic([0.0, 1.5], initrv)
         self.step = 0.5
         sol = probsolve_ivp(self.ivp, step=self.step, diffconst=1.0, which_prior="ibm1")
-        state_rvs = sol._state_rvs
+        state_rvs = sol.kalman_posterior.state_rvs
```
pnkraemer, line 134: the kalman_posterior is a public attribute now. (1) one is free to change it; it does not break the implementation: (2) it avoids the awkward state_rv wrapping.

tests/test_diffeq/test_odesolver.py
```diff
@@ -2,10 +2,24 @@
...
+
+    @property
+    def t(self):
+        return self._t
+
+    @property
+    def y(self):
+        return self._y
+
+
```
pnkraemer, line 22: this is the minimum implementation of ODE solutions. One could also provide this in the source code, but since we do not have solvers that need this (currently), I find it unnecessary.

tests/test_diffeq/test_odesolver.py
```diff
@@ -21,6 +35,9 @@ def step(self, start, stop, current):
             np.nan,
         )  # return nan as error estimate to ensure that it is not used
 
+    def rvlist_to_odesol(self, times, rvs):
```
pnkraemer, line 38: this method must be implemented now.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -105,16 +113,8 @@ def _odesmooth(self, ode_solution, **kwargs):
...
-
-        smoothed_solution = ODESolution(
-            times=ivp_smoother_posterior.locations,
-            rvs=ivp_smoother_posterior.state_rvs,
-            solver=ode_solution._solver,
-        )
-
-        return smoothed_solution
+        smoothing_posterior = self.gfilt.smooth(ode_solution.kalman_posterior)
+        return KalmanODESolution(smoothing_posterior)
```
pnkraemer, line 117: since sol.kalman_posterior is a public attribute, one could just change this as an attribute, i.e.

`ode_solution.kalman_posterior = self.gfilt.smooth(ode_solution.kalman_posterior); return ode_solution`. Which option do you prefer?

pnkraemer:
I am done with the ODE solution implementation. I commented wherever I thought one would want an explanation behind a decision. Please resolve everything that you do not think requires a discussion.

src/probnum/diffeq/__init__.py
pnkraemer: also, I changed the diffeq-internal imports to relative imports (I think this is better wrt circular imports, etc.)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: No I like it :D

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: The version you implemented seems cleaner to me :) But I don't have a strong preferenece so feel free to change it if you want to.

src/probnum/diffeq/odesolution.py
```diff
@@ -1,176 +1,70 @@
-"""ODESolution object, returned by `probsolve_ivp`
+"""ODESolution interface.
 
+This object is returned by ODESolver.solve().
+
+Provides dense output (by being callable), is sliceable, and collects the time-grid as well as the discrete-time solution.
 Contains the discrete time and function outputs. Provides dense output
```
nathanaelbosch, line 6: Some things seem to appear twice in this docstring, maybe remove the second "provides dense output..."

src/probnum/diffeq/odesolution.py
```diff
@@ -1,176 +1,70 @@
...
-        dy_rvs = [projmat @ rv for rv in self._state_rvs]
-        return _RandomVariableList(dy_rvs)
-
-    @property
-    def _state_rvs(self):
-        """:obj:`list` of :obj:`RandomVariable`:"""
-        return self._kalman_posterior.state_rvs
-
-    def __call__(self, t):
+    def dy(self) -> pnrv_list._RandomVariableList:
```
nathanaelbosch, line 35: It's nice to have types now :)

src/probnum/diffeq/odesolution.py
nathanaelbosch: I agree, I think we should not enforce general ODE solutions to provide dense output.

src/probnum/diffeq/odesolution.py
```diff
@@ -1,176 +1,70 @@
...
+        """First derivative of the discrete-time solution."""
+        raise NotImplementedError("The first derivative has not been implemented")
+
+    # Not abstract, because providing interpolation could sometimes be tedious.
+    def __call__(
+        self, t: float
+    ) -> typing.Union[pnrv.RandomVariable, pnrv_list._RandomVariableList]:
```
nathanaelbosch, line 42: Actually, something seems wrong here: I think the output is a RV if t is a float, and a RVList if t is a list of floats! At least that's what KalmanPosterior is doing right now. I'm not sure if that's actually what we want to have though.

I would say we should either adjust the types here to specify that `t` could also be a list/array of floats, or decide if we want a different behaviour and adjust KalmanPosterior and the KalmanODESolution accordingly.

src/probnum/diffeq/odesolution.py
```diff
@@ -1,176 +1,70 @@
...
-        return np.array([projmat @ sample for sample in samples])
+        return len(self.y)
+
+    def __getitem__(self, idx: int) -> pnrv.RandomVariable:
+        """Access the :math:`i`th element of the discrete-time solution."""
+        return self.y[idx]
+
+    def sample(
+        self,
+        t: typing.Optional[float] = None,
```
nathanaelbosch, line 66: Here I'm wondering again what the function _should_ actually do, and in particular about the role of `t`. What should be sampled here?

If you have a clear idea, it would be great if you could formulate that as a docstring and add it here, otherwise we can discuss what behaviour we want here.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -0,0 +1,120 @@
...
+        return self.kalman_posterior.locations
+
+    @property
+    def y(self) -> pnrv_list._RandomVariableList:
```
nathanaelbosch, line 87: Just an idea: Should we cache this?

nathanaelbosch (changes requested):

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
pnkraemer: good idea

src/probnum/diffeq/odesolution.py
pnkraemer: true. I messed up the missing `Union` for the input.

src/probnum/diffeq/odesolution.py
pnkraemer: I forgot the Union again. Will also explain in the docstring what happens with different inputs.

pnkraemer:
I made all the updates that you wanted (I think). Once the CI passes, I will re-request your review. I hope things are better now.

nathanaelbosch (approved):
Look good to me :)

PR #297, opened 2021-01-12
Square-root Kalman filtering

pnkraemer:
Work in progress. I think I have made all the necessary changes in the transition modules, etc.
My plan is now to implement a square-root KF that applies to (E)KF models and linear priors next.


pnkraemer:
A test for the square-root filter that checks whether the filtering & smoothing outputs on the car tracking model are virtually identical passes (phew!). 


Oh, and these tests are written in **pytest**. It is my first time writing pytests, so my testing code might be bad. I am sure it will improve with feedback and time :)

pnkraemer:
Next up: trying to clean up the PR and make it ready for review without changing too many more things. Some parts are not there yet, though.

JonathanWenger:
Once you merge the newest changes from `master` the test failures on Python 3.6 should disappear I believe.

pnkraemer:
> Once you merge the newest changes from `master` the test failures on Python 3.6 should disappear I believe.

Thanks for the hint! Unfortunately it doesnt... I think it has got to do with the types I added. I will try to put them in quotation marks, perhaps that helps.

pnkraemer:
This PR is up-to-date now. 
The square-root bit was simplified quite drastically by modularising the Kalman filter steps.

Other than implementing (and testing) the square-root implementation, this PR introduces a joint interface for approximate filtering algorithms, and respective interfaces for EKF and UKF. 
It also introduces some more types, but since python 3.6 complains about a few things here, I removed some again as well. 

Side notes:
The sqrt implementation should provide a template for Joseph form implementations, which could happen very soon. 
The functions in `kalman_utils` could be moved around a bit in the future (see #307 ), but this will be easy once the current PR is merged.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -34,6 +36,29 @@ def predict_via_transition(
...
 
 
+def predict_sqrt(
+    dynamics_model,
+    rv,
+    start,
+    stop,
+    _intermediate_step=None,
+    _linearise_at=None,
+    _diffusion=1.0,
```
nathanaelbosch, line 46: I'm just wondering: Should the function really have these unused arguments? (`_intermediate_step` and `_diffusion`)

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -52,13 +77,33 @@ def measure_via_transition(
     )
 
 
+def measure_sqrt(
+    measurement_model,
+    rv,
+    time,
+    _linearise_at=None,
+) -> (pnrv.RandomVariable, typing.Dict):
+    """Compute the prediction in square-root form."""
```
nathanaelbosch, line 86: The docstring does not match ;)

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -218,3 +306,96 @@ def rts_smooth_step_classic(
...
+    new_mean = unsmoothed_rv.mean + G @ (smoothed_rv.mean - predicted_rv.mean)
+    return pnrv.Normal(new_mean, new_cov, cov_cholesky=new_cov_cholesky), {}
+
+
+########################################################################################################################
+# Helper functions
+########################################################################################################################
+
+
+def linear_system_matrices(model, rv, start, stop, _linearise_at):
```
nathanaelbosch, line 357: The function seems a bit weird to me. At least I'm still a bit unsure about how I see the whole "start/stop" notion in these transitions. In discrete models, we pass `stop=None`, which feels surprising at first sight (do we just transition forever?). But I mean I totally understand why the current implementation also makes sense, and I also don't really see a better alternative.

src/probnum/filtsmooth/gaussfiltsmooth/linearizing_transition.py
```diff
@@ -0,0 +1,70 @@
...
+import numpy as np
+
+import probnum.random_variables as pnrv
+import probnum.type as pntype
+from probnum.filtsmooth import statespace
+
+
+# Naming may be an overfit to EKF, but the next more general name that admits UKF
+# would be something like "ApproximateTransition" which says less than LinearizingTransition?
+class LinearizingTransition(statespace.Transition, abc.ABC):
```
nathanaelbosch, line 18: Is it not rather a LinearizedTransition?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -19,30 +21,89 @@ def problem():
...
+def smooth_step():
+    """Square-root smoothing step."""
+    return pnfs.rts_smooth_step_sqrt
+
+
+# Test setup
+
+
+@pytest.fixture
+def kalman(problem, predict, measure, update, smooth_step):
```
nathanaelbosch, line 95: Are all 2^4 combinations of predict, measure, update, and smoothstep are tested here, for each problem?

nathanaelbosch: (just wondering about how things work, no need to change anything in any case^^)

src/probnum/filtsmooth/gaussfiltsmooth/linearizing_transition.py
nathanaelbosch: Alright, coming back here since I understand a little more now. I see now how this is not a LinearizedTransition. But overall I find this part a bit confusing.

Vaguely speaking, maybe there _should_ be something like a "linearized transition" somewhere, explicitly? Or have `linearize` have a return value, instead of changing some internal variable?

But this feels hard to discuss here and I'm also definitely not close enough to the code right now. Maybe we could just have a chat about that design decision tomorrow? :)

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -1,102 +1,151 @@
...
+        stop: typing.Optional[float] = None,
+        step: typing.Optional[float] = None,
+        _linearise_at: typing.Optional[pnrv.RandomVariable] = None,
+        _diffusion: typing.Optional[pntype.FloatArgType] = 1.0,
+        **kwargs
+    ) -> (pnrv.Normal, typing.Dict):
+
+        # Do we want the EKF to linearise first and then transition the real,
+        # or try to call the non-linear transition if possible?
+        # There is no right or wrong, but what is the desired behaviour?
```
nathanaelbosch, line 37: The way I see it, the EKF is only really needed when we transition a RV. For an actual realization, we don't need to approximate anything. Does that make sense? Or did I maybe completely miss the point? :D

pnkraemer: No you did not miss a point as far as I know :) This is precisely my concern.
I think the cases we need to keep in mind here is what happens if (a) the non-linear transition allows transition_realization (in which case we want to call this, I suppose?) and (b) if it does not allow it (do we want an error, or do we want to call the extended kalman?. Currently, the code uses case (a); I am not sure 100% what happens in case (b). We can also leave this in here, and solve it in the future. I dont know what the best solution would be.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
pnkraemer: oops

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
pnkraemer: This start stop malarkey is set into stone via #313 now

src/probnum/filtsmooth/gaussfiltsmooth/linearizing_transition.py
pnkraemer: removed it

src/probnum/filtsmooth/gaussfiltsmooth/linearizing_transition.py
```diff
@@ -0,0 +1,70 @@
+# """Interface for approximate transitions.
```
nathanaelbosch, line 1: Should this file stay?

nathanaelbosch (commented):
Looks good to me! I'm not sure if there were any other open points so if there are any that I missed please let me know. Otherwise we should be able to merge this quickly :)

src/probnum/filtsmooth/gaussfiltsmooth/linearizing_transition.py
pnkraemer: removed it

nathanaelbosch (approved):

PR #304, opened 2021-01-23
Fixed kalman initial data points

### LINKED ISSUE #303: Kalman.filter initial data points
Kalman.filter takes the initrv for granted and does not update it. Instead, its first state is conditioned on the zeroth data point, which is weird.

pnkraemer:
Fixes #303 and introduces minor documentation for Kalman filter initialisation, which was missing (as pointed out by Nina). 

The main difference to before is that the initial random variable in the filter is the location of the first data point! (Which makes a million times more sense?!). The rest is making sure that all the shape tests (which now break ofc) are corrected.


nathanaelbosch (approved):
Looks good! And the PR was nice and small, fast to review ;) I think I understood the goal of the PR but I'm not 100% sure, but I think it's easiest if we just have a very quick chat next week.

PR #305, opened 2021-01-30
Modularise Kalman and fix iterated filtering and smoothing

pnkraemer:
# What is happening

This PR simplifies the Kalman environment in a number of ways, and as a side result, fixes iterated filtering and smoothing, tidies up the preconditioning approaches (matern and ioup have it now, too), and removes a bunch of code (the delta is negative with ~200 lines!). The following is happening here (I tried to sort it a bit).

## Kalman heavy lifting 

At initialisation of `Kalman` objects, there is the option to choose the kind of smoothing step, update, etc.. These are all set to default values (the classical iterations), but it will be straightforward to include Joseph forms, square-root forms (and therefore make #297 much simpler), etc.. This also affects the:

## Preconditioning
Using preconditioning in the Kalman update is now a matter of choosing the respective smoothing step (only smoothing really benefits here). This concentrates the impact of preconditioning to a simple decorator (see `kalman_utils.py`), is fairly orthogonal to predict(), etc., and thus can be used out-of-the-box for IOUP and matern. Preconditioned smoothing has its own unittest now for all three integrators.

## Iterated filtering
IEKF is now simply a matter of wrapping your favourite update scheme (`update_classic`) in the decorator `iterate_update`). IEKS is implemented as an `iterated_filtsmooth` method in Kalman. Both are tested; the latter is tested on the pendulum and the logistic ODE. Both use the same stopping criterion (atol, rtol, maxiter). 

## Tests
Tests for kalman, iterated kalman, stoppingcriterion, kallman_utils, etc. are all written in pytest. This required a few more pylint ignores (redefined-outer-name, for instance, which stems from pytest.fixture) all of which are now in both tox.ini and pyproject.toml

## Other
It felt like I removed a whole lot of code as well as the respective tests. Unfortunatlely this seems to affect my code coverage, which is now only at roughly 86% (desired is 90). I am not sure whether this is a dealbreaker though. 

Remaining comments I will put into the diff.


pyproject.toml
```diff
@@ -59,6 +59,9 @@ line-too-long,
 missing-class-docstring,
 unnecessary-pass,
 unused-variable,
+redefined-outer-name,
+function-redefined,
+import-error
```
pnkraemer, line 64: pytest and pylint seem to be not so good friends...

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -60,12 +60,12 @@ def step(self, t, t_new, current_rv):
...
         # 3.2 Update the measurement covariance (measure again)
-        meas_rv, info = self.gfilt.measure(t_new, pred_rv)
+        meas_rv, info = self.gfilt.measure(rv=pred_rv, time=t_new)
 
         # 4. Update
         zero_data = 0.0
-        filt_rv = self.gfilt.condition_state_on_measurement(
-            pred_rv, meas_rv, zero_data, info["crosscov"]
+        filt_rv = pnfs.condition_state_on_measurement(
+            pred_rv, meas_rv, info["crosscov"], zero_data
```
pnkraemer, line 68: the argument orders are changed slightly now

src/probnum/filtsmooth/__init__.py
```diff
@@ -7,16 +7,22 @@
...
     KalmanPosterior,
     StoppingCriterion,
     UnscentedTransform,
+    condition_state_on_measurement,
+    iterate_update,
+    measure_via_transition,
+    predict_via_transition,
+    rts_smooth_step_classic,
+    rts_with_precon,
+    update_classic,
```
pnkraemer, line 20: these do all the heavy lifting for Kalman

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -24,406 +33,186 @@ class Kalman(BayesFiltSmooth):
...
+        smooth_step=rts_smooth_step_classic,
+    ):
+        self.dynamics_model = dynamics_model
+        self.measurement_model = measurement_model
+        self.initrv = initrv
+        self.predict = lambda *args, **kwargs: predict(dynamics_model, *args, **kwargs)
+        self.measure = lambda *args, **kwargs: measure(
+            measurement_model, *args, **kwargs
+        )
+        self.update = lambda *args, **kwargs: update(measurement_model, *args, **kwargs)
```
pnkraemer, line 63: nail down the inputs that never change.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -24,406 +33,186 @@ class Kalman(BayesFiltSmooth):
...
+        """Compute an iterated smoothing estimate with repeated posterior linearisation.
 
-        Returns
-        -------
-        KalmanPosterior
-            Posterior distribution of the smoothed output
+        If the extended Kalman filter is used, this yields the IEKS. In
+        any case, the result is an approximation to the maximum-a-
+        posteriori estimate.
         """
```
pnkraemer, line 79: this is a function in Kalman now, since there is so little going on other than repeating filtsmooth a few times.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -24,406 +33,186 @@ class Kalman(BayesFiltSmooth):
...
+
+        if stopcrit is None:
+            stopcrit = StoppingCriterion()
```
pnkraemer, line 82: there are default atol, rtol, and maxit choices.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -24,406 +33,186 @@ class Kalman(BayesFiltSmooth):
...
+        old_posterior = self.filtsmooth(
+            dataset=dataset,
+            times=times,
+            _intermediate_step=_intermediate_step,
+            _previous_posterior=None,
+        )
+        new_posterior = old_posterior
+        new_mean = new_posterior.state_rvs.mean
+        old_mean = np.inf * np.ones(new_mean.shape)
+        while stopcrit.do_not_terminate_yet(
```
pnkraemer, line 94: I am not sure about the name of this function, but out of all the options I tried, this was the least confusing. This function basically computes the "internal norm" (you know what I mean), compares it to 1, and throws and error if the maxit is reached.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -24,406 +33,186 @@ class Kalman(BayesFiltSmooth):
...
-                curr_rv,
+
+            # Intermediate prediction
+            predicted_rv, info = self.predict(
+                rv=unsmoothed_rv,
                 start=locations[idx - 1],
                 stop=locations[idx],
-                _intermediate_step=_intermediate_step,
+            )
+            crosscov = info["crosscov"]
```
pnkraemer, line 210: these are necessary, because extracting the smoothing parts reduces down to only the update.

After #301 this will be easy to adress by caching pred_rvs as it should be.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+def predict_via_transition(
+    dynamics_model,
+    start,
+    stop,
+    rv,
+    _intermediate_step=None,
+    _linearise_at=None,
+    _diffusion=1.0,
```
pnkraemer, line 23: here, one can add a square-root implementation (for linear(ised) systems). It needs the same signature, but other than that, anything is possible.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+) -> (pnrv.RandomVariable, typing.Dict):
```
pnkraemer, line 24: these are the only types I am including here. I think it makes sense that all these functions have the same output signature (which I am trying to establish here), but more than that I did not dare as long as we support python 3.6 :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+        rv=rv,
+        time=time,
+        _linearise_at=_linearise_at,
+    )
+    crosscov = info["crosscov"]
+    updated_rv = condition_state_on_measurement(
+        pred_rv=rv, meas_rv=meas_rv, crosscov=crosscov, data=data
+    )
+    return updated_rv, meas_rv, {}
+
```
pnkraemer, line 79: here, the joseph updates and square-root updates can be added. As long as they have the same signature, they are immediately usable inside Kalman.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+    new_mean = pred_rv.mean + crosscov @ np.linalg.solve(
+        meas_rv.cov, data - meas_rv.mean
+    )
+    new_cov = pred_rv.cov - crosscov @ np.linalg.solve(meas_rv.cov, crosscov.T)
+    updated_rv = pnrv.Normal(new_mean, new_cov)
+    return updated_rv
+
+
+def iterate_update(update_fun, stopcrit=None):
+    """Iterated update decorator."""
```
pnkraemer, line 92: this is a decorator that turns any (E/U)KF into an I(E/U)KF.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+    --------
+    >>> import functools as ft
+    >>> rts_smooth_step_classic_with_precon = rts_with_precon(rts_smooth_step_classic)
+    """
+
+    def new_smoothing_function(*args, **kwargs):
+        return _rts_smooth_step_with_precon(smooth_step_fun, *args, **kwargs)
+
+    return new_smoothing_function
+
```
pnkraemer, line 158: preocnditioning is also just a decorator now, therefore it will be compatible with any smoothing update immeditaly (including square-root!)

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+    unsmoothed_rv,
+    predicted_rv,
+    smoothed_rv,
+    crosscov,
+    dynamics_model=None,
+    start=None,
+    stop=None,
+) -> (pnrv.RandomVariable, typing.Dict):
+    """Compute a classical Rauch-Tung-Striebel smoothing step."""
+    smoothing_gain = crosscov @ np.linalg.inv(predicted_rv.cov)
```
pnkraemer, line 207: this is basically the only line where we need preconditioning.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -0,0 +1,215 @@
...
+        unsmoothed_rv,
+        predicted_rv,
+        smoothed_rv,
+        crosscov,
+        dynamics_model=dynamics_model,
+        start=start,
+        stop=stop,
+    )
+    new_rv = precon @ updated_rv
+    return new_rv, {}
```
pnkraemer, line 194: this structure now is not only applicable for IBM, but for any `Integrator`. IOUP and Matern use it as well now.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -92,12 +92,12 @@ def __call__(self, t):
 
     def _predict_to_loc(self, loc):
         """Predict states at location `loc` from the closest, previous state."""
-        prev_idx = (self.locations < loc).sum() - 1
-        prev_loc = self.locations[prev_idx]
-        prev_rv = self.state_rvs[prev_idx]
+        previous_idx = (self.locations < loc).sum() - 1
+        previous_loc = self.locations[previous_idx]
```
pnkraemer, line 96: "prev" and "pred" were too similar IMO

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -144,8 +153,17 @@ def _single_sample_path(self, locations, random_vars):
...
-            curr_rv = self.gauss_filter.smooth_step(
-                unsmoothed_rv, curr_rv, start=locations[idx - 1], stop=locations[idx]
+
+            # Intermediate prediction
+            predicted_rv, info = self.gauss_filter.predict(
+                rv=unsmoothed_rv,
+                start=locations[idx - 1],
+                stop=locations[idx],
+            )
+            crosscov = info["crosscov"]
```
pnkraemer, line 163: again, will be cached soon which will make this step here unnecessary

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -24,6 +24,16 @@ class Integrator:
...
         self.spatialdim = spatialdim
+        self.precon = None  # to be set later
+
+    def setup_precon(self):
+        # This can not be done in the init, due to the order in which
+        # IBM, IOUP, and Matern initialise their superclasses.
+        # Integrator needs to be initialised before LTISDE,
+        # in which case via LTISDE, self.precon is set to None
+        # To make up for this, we call this function here explicitly
+        # after the calls to the superclasses' init.
```
pnkraemer, line 35: this.

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -127,33 +136,26 @@ def transition_rv(self, rv, start, stop, _diffusion=1.0, **kwargs):
...
     def transition_realization(self, real, start, stop, _diffusion=1.0, **kwargs):
         if not isinstance(real, np.ndarray):
             raise TypeError(f"Numpy array expected, {type(real)} received.")
         step = stop - start
         real = self.precon.inverse(step) @ real
-        real, info = self.transition_realization_preconditioned(
+        (
+            real,
+            info,
+        ) = self.equivalent_discretisation_preconditioned.transition_realization(
```
pnkraemer, line 153: no more calls to `transition_preconditioned` because this function does not exist anymore.

tests/test_filtsmooth/test_gaussfiltsmooth/filtsmooth_testcases.py
```diff
@@ -148,7 +149,27 @@ def dh(t, x):
     dynamod = pnfs.statespace.DiscreteGaussian(f, lambda t: q, df)
     measmod = pnfs.statespace.DiscreteGaussian(h, lambda t: r, dh)
     initrv = Normal(initmean, initcov)
-    return dynamod, measmod, initrv, {"dt": delta_t}
+    return dynamod, measmod, initrv, {"dt": delta_t, "tmax": 4}
+
+
+def logistic_ode():
```
pnkraemer, line 155: this test case assures that IEKS works with ODE filters.

tests/test_filtsmooth/test_gaussfiltsmooth/test_iterated_kalman.py
```diff
@@ -1,68 +1,93 @@
...
+
+@pytest.fixture
+def kalman(problem, update):
+    """Create a Kalman object."""
+    dynmod, measmod, initrv, info, *_ = problem
+    return pnfs.Kalman(dynmod, measmod, initrv)
+
+
+def test_rmse_filt_smooth(kalman, problem):
+    """Assert that smoothing beats filtering beats nothing."""
```
pnkraemer, line 65: "iterated smoothing beats smoothing beats filtering" would be more precise

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -1,107 +1,73 @@
...
+    """The usual Kalman update.
+
+    Yields Kalman filter.
     """
+    return pnfs.update_classic
+
+
+@pytest.fixture
+def update():
+    """Iterated classical update.
```
pnkraemer, line 33: will be useful once the EKF and UKF tests also use pytest

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman_utils.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.fixture
+def dynamics_model(ordint, spatialdim):
+    dynamics_model = pnfss.Matern(ordint, spatialdim, lengthscale=1.0)
+    return dynamics_model
+
+
+def test_rts_smooth_step(
+    dynamics_model, unsmoothed_rv, smoothed_rv, predicted_rv, crosscov, dt
+):
+    """Assert that preconditioning does not affect the outcome of the smoothing step."""
```
pnkraemer, line 100: was missing before, now we can check this easily

pnkraemer (commented):
@nathanaelbosch I think you can review now. Sorry for the long post. I hope you agree that the interface is a bit simpler now :) 

The CI is not passing, because of the patch cov diff. See my explanation above...

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
pnkraemer: I can envision that at some point in the future, the smoothing gain is already computed at the forward filtering pass in the prediction and then cached. In this case preconditioning may concentrate even harder to certain lines in some predictions (it is only the smoothing gain computation that breaks necks here), removing the need for different smoothing steps altogether.

pnkraemer:
I rewrote the kalman posterior tests in pytest (and found that one of the changes was for some reason not covered before. well, now it is. maybe this improves the patch coverage sufficiently :)) 

This unveiled some weird sampling seed bug. I will fix it and edit here accordingly.

src/probnum/filtsmooth/statespace/integrator.py
nathanaelbosch: I'm not sure I fully get why this is necessary. But I trust that it is so this is not a reason not to merge, but let's maybe talk about it briefly some day :)

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -127,33 +136,26 @@ def transition_rv(self, rv, start, stop, _diffusion=1.0, **kwargs):
...
 
     def transition_realization(self, real, start, stop, _diffusion=1.0, **kwargs):
         if not isinstance(real, np.ndarray):
             raise TypeError(f"Numpy array expected, {type(real)} received.")
         step = stop - start
         real = self.precon.inverse(step) @ real
-        real, info = self.transition_realization_preconditioned(
+        (
+            real,
+            info,
```
nathanaelbosch, line 152: Ok this seems to be a case where Black does really not lead to nice code :see_no_evil:

nathanaelbosch: Obviously this does is not meant to be a criticism of your code style^^ I don't see how else to do things

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman_utils.py
```diff
@@ -0,0 +1,125 @@
...
+    return 6
+
+
+@pytest.fixture
+def spatialdim():
+    return 1
+
+
+@pytest.fixture
+def spatialdim():
```
nathanaelbosch, line 34: I have 0 pytest experience, but the above fixtures feel a bit weird. I would expect there might be an easier way to achieve that? I saw [here](https://docs.pytest.org/en/stable/parametrize.html#parametrize) that `@pytest.fixture` has a `params` kwarg, and there is also a `@pytest.mark.parametrize`.

But again, 0 pytest experience here, so I'm 100% fine if you want to leave it as it is! :)

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
```diff
@@ -1,156 +1,185 @@
...
+
+
+@pytest.fixture
+def posterior(kalman, problem):
+    """Kalman filtering posterior."""
+    *_, obs, times, states = problem
+    return kalman.filter(obs, times)
+
+
+# This will raise a warning for the extrapolation test...
```
nathanaelbosch, line 40: Is there some way to catch that? Just because ideally, we'd like pytest not to return any warning at all..

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
```diff
@@ -1,156 +1,185 @@
...
+
+
+@pytest.fixture
+def locs(posterior):
+    return posterior.locations[[2, 3]]
+
+
+@pytest.fixture
+def locs():
+    return np.arange(0, 0.5, 0.025)
```
nathanaelbosch, line 152: I'm wondering: Is this also called in the tests? I'm still a bit confused with pytest, and I'm surprised to see `locs` with different function signatures, so I just want to make sure things work as I'd expect them to.

nathanaelbosch (commented):
Really cool stuff! And good job on the tests, too! :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: I see your point. I think it is a valid concern. I think the way to go would be to rename the arguments, not the attributes.
An underscore in the arguments might suggest that we dont want people to change this though, which I think is not the way to go.

How about something like `use_predict=...`, `use_measure=`, ...; then, `self.predict` could remain as is, and it would be clear that this is more of a rule that is passed on that the actual function. (Nb: I think `predict_rule`, `measure_rule`, etc., are not very good names). 

I will rename with `use_*`, and if you agree with this choice then please feel free to resolve. If not, please complain, so we can find a good solution!

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: isnt this a problem with mutability? I agree that this would be easier, but I am a bit timid to do it like this and potentially introduce bugs that we will never find again...

What do you think?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: I changed it to `while not stopcrit.terminate`. Tbf, this is how I started writing it but then I got so confused with the trues, falses, and negations, that I changed things around until they worked and then left it there :D Of course, simplifying the working solution again is a good idea.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -24,406 +33,186 @@ class Kalman(BayesFiltSmooth):
...
-        """Apply Gaussian filtering (no smoothing!) to a data set.
-
-        Parameters
-        ----------
-        dataset : array_like, shape (N, M)
-            Data set that is filtered.
-        times : array_like, shape (N,)
-            Temporal locations of the data points.
-            The zeroth element in times and dataset is the location of the initial random variable.
+    def filter(self, dataset, times, _intermediate_step=None, _previous_posterior=None):
```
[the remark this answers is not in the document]
pnkraemer: Oooops... I restored it.

src/probnum/filtsmooth/statespace/integrator.py
pnkraemer: Short summary: both `integrator` and `SDE` have a `precon` attribute. `SDE`s set it to `None` as a default, but `Integrators` enjoy `NordsieckLikeCoordinates`. Therefore the order of inits matters. 
Unfortunately, due to the values in `driftmat`, etc., `Integrator.__init__` needs to be called before `SDE.__init__`, which messes up the `precon` field. This is what the fix is for.

Please ask me in person if this is not clear yet, I understand that it is a bit twisted...

src/probnum/filtsmooth/statespace/integrator.py
pnkraemer: I agree... but hey, black cannot do everything perfectly! I changed the code here to `out = ...; real, info = out` and black is fine with that now.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman_utils.py
pnkraemer: I am not sure either here... I think one could also parametrise the test functions below, but wouldnt we need to copy the string `pytest.mark.parametrize(["ordint", "spatialdim"], [[0, 1, 6], [1, 4]]` to all the test functions that need this input?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
pnkraemer: In principle, yes, but I am not sure currently how to best tie this to this specific instance of `posterior`. In fact, I am leaning more towards removing the warning than fixing this test (beyond the final data point, calling the posterior is pure extrapolation, which I think is as expected). What do you think?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
pnkraemer: I think so... I see that this looks weird. Shall I put the `posterior` argument into the other fixtures, as well?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
pnkraemer: I removed the warning. Please complain if you think this was not a good idea. If not, please resolve :)

pnkraemer:
I made all the changes that seemed necessary. There are some open discussions. Would you be so kind @nathanaelbosch and go through the discussion points and either elaborate or resolve, at some point in the next days? Thanks for your swift feedback so far!

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman_utils.py
nathanaelbosch: That might be. Since we don't know, maybe we just keep it like that and just refactor later if necessary?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: Sounds good :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: I think I don't really understand the potential mutability problem that you mention. What bugs could that maybe introduce?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
nathanaelbosch: Maybe in this one (`locs`) to make things a bit more clear? Though as long as all of them run correctly, maybe it's also fine as it is.

nathanaelbosch: Oh but I just saw in line 141 that there is a `locs(posterior)` that does not use the `posterior` in the function, so that should be removed (if we keep this function as it is, too).

nathanaelbosch (commented):
I commented on some remaining discussion, but I think we'll be able to merge soon

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: I am talking about https://florimond.dev/blog/articles/2018/08/python-mutable-defaults-are-the-source-of-all-evil/ which may apply here; for instance, `StoppingCriterion()` sets and reset `self.iterations` depending on whether something is accepted or not. 

Though I am not sure whether I am seeing ghosts in this context. Purely from a simplification perspective, I agree that `stopcrit=StoppingCriterion()` is simpler.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman_utils.py
pnkraemer: Okay! I suspect that by the time our pytest skills improve, we may come back to the tests anyway.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
pnkraemer: I put it in, for the sake of a uniform function signature. I dont think it matters too much, though if this turns out to be bad, that will be an easy fix.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
nathanaelbosch: Thanks, that makes sense! :+1:

nathanaelbosch (approved):

PR #310, opened 2021-02-04
Discrete transition variable names

pnkraemer:
## Problem
The variable names in `discrete_transition` have been bugging me for quite a while. 1) they are too similar to the sde names (force, diffmat, dynamics/drift); 2) they are fairly meaningless.

## Solution
I renamed the variables in discrete_transition according to the schemes `state_trans_*` (formerly `dynamics*`), `shift_*` (formerly `force*`), and `proc_noise_cov_*` (formerly `diffmat*`).
This is now sufficiently different to the `SDE` variable names, and finally aligned with how we write about these matrices. 

I know the names are fairly long, but I prefer long over confusing. 

Thoughts?


nathanaelbosch (approved):
I like the name changes :) Also, nice and small PR, great to review, so good job!

Maybe we should merge the other PR first, since cleaning up things in here is probably easier, but feel free to do as you think is best.

Issue #313, opened 2021-02-05
Start stop vs t


pnkraemer:
`start` and `stop` are a bit weird for transitions, in the sense that discrete transitions dont really have a starting value.
Some distinction of this form is important though, because continuous systems need some sort of `stop` value.

A better approach seems to be `t` and `dt`, in the sense of `stop=t+dt`. Alternatively `time` and `incr(ement)`?!

This is somehow related to #307

Issue #307, opened 2021-02-01
Message passing in Transition objects


pnkraemer:
Currently, the `Transition` objects implement Markov transitions in continuous and discrete time.

In functions like `transition_rv`, and `transition_realization`, what happens is that (the message of) a state at time `start` is passed to a state at time `stop`. Examples include calls to the equivalent discretisation of LTI systems, or solving the moment equations for the solution of a linear SDE. 

**What is the reverse message pass here?**
Essentially, the reverse message is passed in `Kalman.condition_state_on_measurement` and `Kalman.smooth_step`, respectively in the implementations thereof (see the changes in #305).  This implies that a `KalmanPosterior` object needs to hold a `Kalman` object, because this is the only way of accessing the backward pass. Since the backward pass is independent of the data (and the measurement model), some refactoring could make this design a bit simpler.

**Instead**
What we could do instead is take more of a message passing perspective on `Transition` objects, in the sense that these objects know how to (i) forward pass both, a RV and a realization (which currently happens in `transition_*`); (ii) the backward pass of a realisation (which could be used for sampling from the posterior) and of a RV (which could be used for smoothing). I am not sure about the naming, but something along the lines of `forward_rv`, `forward_real`, `backward_rv`, `backward_real` would at least be precise. This is open for discussion though. This would essentially entail implementing the smoothing steps in the respective `Transition` objects. I think that is a good idea.

Changing the perspective to more of a message passing notion might require a few renamings (for instance, I suggest `SDETransition` instead of `SDE`, to make sure that this is merely an implementation of a transition and may not make a lot of sense as an SDE itself) and some updates in the docstrings. 

One thing that is not obvious in these changes (which would remove responsibility from `Kalman` and shove it into `Transition`) is how to deal with functions like `Kalman.update` which calls `measure` (essentially `transition_rv`) and `condition_state_on_measurement`.  In for instance square-root Kalman filtering, we want to execute them at the same time. Where would this happen? In `Kalman`? Which information would need to be provided to `Kalman` objects?

pnkraemer:
>  In for instance square-root Kalman filtering, we want to execute them at the same time. Where would this happen? In Kalman? Which information would need to be provided to Kalman objects

I think the cleanest way might be to provide a method `forward_rv_and_backward_real` to transition objects, which either calls both but can also be overwritten. 

-----------------------------------------------------------------------------------------------
P.S: 
@schmidtjonathan this issue is what I was talking about yesterday.

[...]

Issue #306, opened 2021-01-31
KalmanPosterior 


pnkraemer:
The `KalmanPosterior` -- as useful as it is already -- needs a few updates.
0) It should be assembled on the fly (see #301 ) 
1) It needs to store predicted RVs (i.e. the filter posterior) and smoothing gains, not only to make sampling and interpolation cheaper, but also to fix a nasty sampling bug. I think it makes sense to not have the `with_smoothing` flag anymore, but instead populate different lists (e.g. `filt_post_rvs`, `smoo_post_rvs`, etc.) which can then be accessed accordingly. This may also simplify the preconditioning setup even more. 
2) Extrapolation beyond the time points needs no warning 
3) backwards extrapolation (lower than self.locations[0]) should not throw an error, but "predict backwards")
4) It needs a test that asserts that the sample variance matches the marginal variance (this would have unvealed a few of the issues in the past). I tried this in #305 but it turned out that it is very expensive, so I deleted the code again. Ideas are welcome here :)

pnkraemer:
Related to #271

pnkraemer:
I suppose the easiest solution is to give the `KalmanPosterior` a field `filter_posterior` (default to None), which in the case of the smoothing posterior, can hold the filtering posterior (and all its callability, etc.) and refer to this guy when needed (e.g. for sampling).

pnkraemer:
Partially fixed in #317

PR #316, opened 2021-02-10
Refactoring transitions (and some resulting filtering implementations)

pnkraemer:
Takes care of #307 and #313 (and perhaps some other things). In a nutshell: predict and update implementations are now in the transition objects (via forward and backward). Choices between square-root, joseph form, etc., are made on that level (though access to this choice in Kalman, for instance, is open for discussion). 

Even though the line delta appears quite large now, I suspect it may not be too big in the end -- if not negative? So much filtering and ODE code will be obsolete, but has not been touched yet.


pnkraemer:
@nathanaelbosch @schmidtjonathan in case you're interested what the main changes are, look at the documentation of `Transition` and go from there. Decisions like square-root or joseph form implementations are most prominent in `LinearGaussian`. In its equivalent, `LinearSDE`, there is only the framework for this decision (which is where one would also choose the type of smoother -- there is more than one). Implementations thereof are postponed to a different PR. 

More about this tomorrow, which will also be when I assign you guys as reviewers.

src/probnum/kernels/_kernel.py
```diff
@@ -184,6 +184,13 @@ def _check_and_reshape_inputs(
                 "dimension."
             )
 
+            # The below was weirdly necessary to make pylint pass
+            # even though I have not touched anything in the vicinity
+            # of kernels. Please, anyone feel free to remove this
+            # pylint-disable again, provided it passes the CI :)
+            #
+            # pylint: disable=redefined-variable-type
```
schmidtjonathan, line 192: Should be removed to avoid a conflict due to Nina's PR.

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
```diff
@@ -1,401 +1,419 @@
...
+#
+#
+# def triu_to_positive_tril(triu_mat: np.ndarray) -> np.ndarray:
+#     r"""Change an upper triangular matrix into a valid lower Cholesky factor.
+#
+#     Transpose, and change the sign of the diagonals to '+' if necessary.
+#     """
+#     tril_mat = triu_mat.T
+#     with_pos_diag = tril_mat @ np.diag(np.sign(np.diag(tril_mat)))
+#     return with_pos_diag
```
schmidtjonathan, line 419: The entire file should be removed before merging into master, imo.

src/probnum/filtsmooth/statespace/transition.py
```diff
@@ -1,221 +1,355 @@
...
-        start: float,
-        stop: Optional[float] = None,
-        step: Optional[float] = None,
-        _diffusion: Optional[float] = 1.0,
-        _linearise_at: Optional[RandomVariable] = None,
-    ) -> (RandomVariable, Dict):
-        """Transition a random variable from time :math:`t` to time
-        :math:`t+\\Delta t`.
+    def forward_realization(
+        self, real, t, dt=None, compute_gain=False, _diffusion=1.0, _linearise_at=None
```
schmidtjonathan, line 95: I find the naming `real` misleading. Should we name that `realization` instead?

src/probnum/filtsmooth/statespace/sde_utils.py
```diff
@@ -0,0 +1,26 @@
+"""Continuous-time transition utility functions."""
+
+
+import numpy as np
+import scipy.linalg
+
+
+def matrix_fraction_decomposition(driftmat, dispmat, dt):
```
schmidtjonathan, line 8: Do you think a check for the value of `dt` would be reasonable?

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -82,64 +124,120 @@ class LinearSDE(SDE):
 
     def __init__(
         self,
+        dimension,
         driftmatfun: Callable[[FloatArgType], np.ndarray],
         forcevecfun: Callable[[FloatArgType], np.ndarray],
         dispmatfun: Callable[[FloatArgType], np.ndarray],
+        mde_atol=1e-6,
```
schmidtjonathan, line 131: Should we add a parameter description for `mde_atol` / `mde_rtol` to the docstring?

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -82,64 +124,120 @@ class LinearSDE(SDE):
...
-            stop,
-            step,
-            self.driftfun,
-            self.driftmatfun,
-            self.dispmatfun,
             _diffusion=_diffusion,
         )
+        sol = scipy.integrate.solve_ivp(
+            mde,
+            (t, t + dt),
```
schmidtjonathan, line 188: What happens here if `dt = None` ? (Which is the default)

src/probnum/filtsmooth/statespace/sde.py
```diff
@@ -82,64 +124,120 @@ class LinearSDE(SDE):
...
+            y0,
+            method=self.mde_solver,
+            atol=self.mde_atol,
+            rtol=self.mde_rtol,
+        )
+        y_end = sol.y[:, -1]
+        new_mean = y_end[: len(rv.mean)]
```
schmidtjonathan, line 195: Suggestion: define `dim = rv.mean.shape[0]`

src/probnum/filtsmooth/gaussfiltsmooth/iterated_component.py
```diff
@@ -0,0 +1,159 @@
...
+            t,
+            dt=dt,
+            compute_gain=compute_gain,
+            _diffusion=_diffusion,
+            _linearise_at=_linearise_at,
+        )
+
+    # Pass on all the rest to the EKF/UKF component
+
+    def __getattr__(self, attr):
```
nathanaelbosch, line 151: All of this feels a bit hacky :sweat_smile: But I couldn't figure out how to do things differently, I thought class decorators might fix this but idk.

nathanaelbosch: Which means I'm fine with keeping things this way :+1: And in the case that you also don't like this part we could think about alternatives together.

tests/test_diffeq/test_odefiltsmooth/test_kalman_odesolution.py
```diff
@@ -137,19 +139,20 @@ def test_output_shape(self):
...
-                    self.solution[:].mean[[2, 3]],
-                    self.solution[:].cov[[2, 3]],
-                )
-                for sample in five_samples
-            ]
-        ).mean()
-        self.assertLess(chi_squared, 10.0)
-        self.assertLess(0.1, chi_squared)
+    #
+    # def test_sampling_two_locations_multiple_samples(self):
```
nathanaelbosch, line 143: This should probably be removed before merging

tests/test_filtsmooth/test_gaussfiltsmooth/filtsmooth_testcases.py
```diff
@@ -149,8 +151,14 @@ def dh(t, x):
...
     initmean = np.ones(2)
     initcov = var * np.eye(2)
-    dynamod = pnfs.statespace.DiscreteGaussian(f, lambda t: q, df)
-    measmod = pnfs.statespace.DiscreteGaussian(h, lambda t: r, dh)
+    dynamod = pnfs.statespace.DiscreteGaussian(
+        2,
+        2,
+        f,
+        lambda t: q,
+        df,
```
nathanaelbosch, line 159: Does this get nicer if you delete the final comma?

tests/test_filtsmooth/test_gaussfiltsmooth/test_iterated_kalman.py
```diff
@@ -7,65 +7,45 @@
...
-
-@pytest.fixture
-def kalman(problem, update):
+def kalman(problem):
     """Create a Kalman object."""
     dynmod, measmod, initrv, info, *_ = problem
     return pnfs.Kalman(dynmod, measmod, initrv)
 
 
+@pytest.mark.parametrize("problem", [logistic_ode_problem(), pendulum_problem()])
```
nathanaelbosch, line 43: Nice, so this is how parametrizing works in pytest :smile:

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -21,88 +21,17 @@ def problem():
...
-    """Square-root smoothing step."""
-    return pnfs.rts_smooth_step_sqrt
-
-
 # Test setup
 
 
 @pytest.fixture
-def kalman(problem, predict, measure, update, smooth_step):
+def kalman(problem):
```
nathanaelbosch, line 28: Ok I have to ask: Do we still test the different implementations here or not? If yes, where in the code do I see that?

tests/test_filtsmooth/test_statespace/conftest.py
```diff
@@ -0,0 +1,98 @@
+"""Fixtures to be shared across all modules in this directory.
```
nathanaelbosch, line 1: The filename is not very clear to me

nathanaelbosch (changes requested):
Nice work! I added some comments here and there, but overall things look very good. I did not really go through the tests in detail, but I could still do so if that would be helpful.

src/probnum/filtsmooth/statespace/transition.py
```diff
@@ -1,221 +1,355 @@
...
+        raise NotImplementedError
 
+    @abc.abstractmethod
+    def backward_rv(
+        self,
+        rv_obtained,
+        rv,
+        rv_forwarded=None,
+        gain=None,
+        t=None,
```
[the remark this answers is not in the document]
pnkraemer: Only when forward_rv is called internally to compute gain and forwarded rv. If not, this information is contained in the gain and the forwarded rv. I suppose it is a trade-off between (`rv_forwarded, gain`) and (`t, dt`) when it comes to positioning. Would you change the orders?

src/probnum/filtsmooth/statespace/generate_samples.py
```diff
@@ -0,0 +1,45 @@
+import numpy as np
+
+
+def generate_samples(dynmod, measmod, initrv, times, num_steps=5):
```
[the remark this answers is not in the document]
pnkraemer: Good question. I tend to use it quite heavily in examples and tests, so I think would be a bit awkward to have it as `_generate_samples`. 
I also changed the name because `generate` was a bit cryptic IMO if encountered in the context of tests and examples.

src/probnum/filtsmooth/statespace/discrete_transition_utils.py
```diff
@@ -0,0 +1,50 @@
+import typing
+
+import numpy as np
+
+import probnum.random_variables as pnrv
+
```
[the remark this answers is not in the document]
pnkraemer: Good question indeed. I am wondering the same. However, when in doubt I rather make these functions public unless it bites me. Any thoughts about this? (@schmidtjonathan?)

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
pnkraemer: Not in here, but their implementations are tested in statespace

tests/test_filtsmooth/test_statespace/conftest.py
pnkraemer: It is a pytest thing, not my decision :(

src/probnum/filtsmooth/gaussfiltsmooth/iterated_component.py
pnkraemer: Isnt this basically  class decorator? In any case, I just wanted something which (a) essentially inherits everyhing but backward_rv from the component that it passed to `IteratedComponent` (e.g. EKF **or** UKF), and (b) has as little code as possible. If you have any other idea, please let me know. We can also talk about this separately, or (hereby) ask @schmidtjonathan's  opinion ;)

src/probnum/filtsmooth/gaussfiltsmooth/kalman_utils.py
pnkraemer: oooops

src/probnum/filtsmooth/statespace/sde.py
pnkraemer: Good question. I changed the code to raise a `ValueError` in this case.

src/probnum/filtsmooth/statespace/sde_utils.py
pnkraemer: What sort of check do you mean? That it is a float? I dont know, I tend to avoid checks unless there is some really weird behaviour that needs to be caught. 
I guess in line 21 below, if `dt` was not a float, there would be a reasonable error. what do you think? 


Also, I do not expect that MFD is used heavily by an outside user...

src/probnum/filtsmooth/statespace/transition.py
pnkraemer: is it misleading even right next to "forward_realization"? One could rename it, I am just a bit worried about line length.

```
abc, info = transition.forward_realization(realization=sth, t=sth, ...)
``` 

vs
```
abc, info = transition.forward_realization(real=sth, t=sth, ...)
```
do you think there is a lot to be gained here? If so, I will consider changing it!

Perhaps @nathanaelbosch has an opinion here?

tests/test_diffeq/test_odefiltsmooth/test_kalman_odesolution.py
pnkraemer: This is related to the discussion about "what does this test do" in #306 

In any case, I removed it.

tests/test_filtsmooth/test_gaussfiltsmooth/filtsmooth_testcases.py
pnkraemer: :D 

Not sure, but it does if I add keywords to the arguments.

src/probnum/filtsmooth/statespace/transition.py
nathanaelbosch: In this case I don't mind having `real`, it seems clear enough to me in the context of a function called `forward_realization` :) But Jonathan might have the most unbiased view on this.

src/probnum/filtsmooth/statespace/transition.py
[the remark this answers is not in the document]
nathanaelbosch: I don't really have an opinion on this, I just wanted to question it in case this was not that intentional. Since you have a better idea on how the code is used, so feel free to keep the current order :)

src/probnum/filtsmooth/statespace/sde_utils.py
schmidtjonathan: > What sort of check do you mean?
Not really a type check but rather, if someone passes `dt = t_i - t_j` and it accidentally results in a negative `dt`, that might result in weird behavior that is difficult to debug.

However, ... 
> Also, I do not expect that MFD is used heavily by an outside user...
... that's probably true.

src/probnum/filtsmooth/statespace/transition.py
schmidtjonathan: Generally, I don't want to let line length compromise the readability of the code. But if both of you think that it's ok, I am fine with it :) I just immediately thought of real in terms of "real part". As in complex numbers. But I guess that's just me then

src/probnum/filtsmooth/statespace/discrete_transition_utils.py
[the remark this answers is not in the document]
schmidtjonathan: Mh, I personally would leave them public since it is a `utils` module, for which the natural intended use is that it is accessed from outside. If the method is expected to be called only in a narrow context, I think it is misplaced in a `utils` file. However, this is probably not the case, since conditioning on measurements is a quite general concept, right?

(My suggestion: leave as is)

src/probnum/filtsmooth/gaussfiltsmooth/iterated_component.py
schmidtjonathan: I think this should be implemented as a wrapper.
I have an idea how this could be implemented elegantly, but I would suggest opening a separate PR for this.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -135,3 +138,10 @@ def filtering_solution(self):
         return KalmanODESolution(
             kalman_posterior=self.kalman_posterior.filtering_posterior
         )
+
+
+def _project_rv(projmat, rv):
```
schmidtjonathan, line 143: Shouldn't this method be part of the respective `RandomVariable` class?

src/probnum/filtsmooth/statespace/transition.py
pnkraemer: Decision: use `realization`

nathanaelbosch (approved):
Looks good :)

schmidtjonathan (approved):

PR #317, opened 2021-02-12
Kalman posterior

pnkraemer:
This PR introduces a clear separation in `KalmanPosterior` (as well as derived objects, such as `KalmanODESolution`) between filter posterior and smoothing posterior.

EDIT: remaining text deleted, as it is out of date.


src/probnum/_randomvariablelist.py
```diff
@@ -18,28 +18,39 @@ def __init__(self, rv_list: list):
...
-            )
+        # If not empty:
+        if len(rv_list) > 0:
+
+            # First element as a proxy for checking all elements
+            if not isinstance(rv_list[0], pn.RandomVariable):
+                raise TypeError(
+                    "RandomVariableList expects RandomVariable elements, but "
+                    + f"first element has type {type(rv_list[0])}."
+                )
```
pnkraemer, line 29: turns out that _RandomVariableList was not prepared for empty lists... :)

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -78,11 +78,15 @@ def __init__(self, kalman_posterior: pnfs.KalmanPosterior):
 
         # Pre-compute projection matrices.
         # The prior must be an integrator, if not, an error is thrown in 'GaussianIVPFilter'.
-        self.proj_to_y = kalman_posterior.gauss_filter.dynamics_model.proj2coord(
-            coord=0
+        self.proj_to_y = (
+            kalman_posterior.filter_posterior.gauss_filter.dynamics_model.proj2coord(
```
pnkraemer, line 82: the "visible" posterior (to the user) is always the smoothing posterior, and if nothing has been smoothed yet, the filterposterior can be accessed through `kalman_posterior.filter_posterior`.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -28,7 +28,7 @@
 
 def probsolve_ivp(
     ivp,
-    method="ekf0",
+    method="eks0",
```
pnkraemer, line 31: was about time

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -294,12 +294,17 @@ def smooth(self, filter_posterior):
...
+
         rv_list = self.smooth_list(
-            filter_posterior,
+            filter_posterior.state_rvs,
             filter_posterior.locations,
         )
         return KalmanPosterior(
-            filter_posterior.locations, rv_list, self, with_smoothing=True
+            filter_posterior.locations, rv_list, self, filter_posterior=filter_posterior
         )
```
pnkraemer, line 308: I guess this block explains best the change of interface of `KalmanPosterior`.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -60,30 +73,46 @@ def __call__(self, t):
         :obj:`RandomVariable`
             Estimate of the states at time ``t``.
         """
+        # Raise an error if smoothing posterior is called,
+        # but only the filtering posterior is available.
+        if len(self._locations) == 0:
+            raise NotImplementedError("Dense output not available.")
+
+        # Recursive evaluation (t can now be any array, not just length 1)
         if not np.isscalar(t):
```
pnkraemer, line 82: please check if from here on below is clear.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -142,10 +171,32 @@ def sample(self, locations=None, size=()):
...
+        if locations[-1] < self.locations[-1]:
+            t, rv = locations[-1], random_vars[-1]
 
-        num_dim = len(random_vars[-1].sample())
+            # Intermediate prediction
+            predicted_rv, info = self.gauss_filter.predict(
+                rv=rv,
+                start=t,
+                stop=self.locations[-1],
+            )
```
pnkraemer, line 187: this should fix the sampling issues

tests/test_filtsmooth/test_gaussfiltsmooth/filtsmooth_testcases.py
```diff
@@ -233,10 +233,10 @@ def test_filtsmooth_pendulum(self):
         method = pnfs.Kalman(ekf_dyna, ekf_meas, initrv)
 
         # Compute filter/smoother solution
-        filter_posterior = method.filter(obs, tms)
-        filtms = filter_posterior.state_rvs.mean
-        smooth_posterior = method.filtsmooth(obs, tms)
-        smooms = smooth_posterior.state_rvs.mean
+        posterior = method.filter(obs, tms)
+        filtms = posterior.filter_posterior.state_rvs.mean
+        posterior = method.filtsmooth(obs, tms)
```
pnkraemer, line 238: this also kind of explains the change of language in probnum posterior world.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
```diff
@@ -169,16 +169,17 @@ def test_sampling_shapes(samples, posterior, locs, size):
...
+#     # what is computed below in here???
+#     all_chi_squareds = [
+#         chi_squared_statistic(sample, posterior(locs).mean, posterior(locs).cov)
+#         for sample in samples
+#     ]
+#     chi_squared = np.mean(all_chi_squareds)
+#
+#     # This range is quite large, but a failing tests here would at least
+#     # unveil strong deviations...
+#     assert 0.01 < chi_squared < 100.0
```
pnkraemer, line 185: what does this test do? Do you know? I am pretty sure it does at least not do what it claims to do. I'd delete it, but please tell me your opinion.

pnkraemer:
This is what things look like now: (25 samples, 3*stdev, everything evaluated at dense output, plus the actual steps depicted as markers)
Screenshot from 2021-02-12 12-48-23

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -11,8 +13,9 @@
 from probnum.filtsmooth.filtsmoothposterior import FiltSmoothPosterior
 
 
-class KalmanPosterior(FiltSmoothPosterior):
-    """Posterior Distribution after (Extended/Unscented) Kalman Filtering/Smoothing.
+class KalmanPosterior(FiltSmoothPosterior, abc.ABC):
+    """Interface for posterior distribution after (extended/unscented) Kalman
+    filtering/smoothing.
```
pnkraemer, line 18: filtering and smoothing posterior share almost anything, so they have this interface (it is abc!)

The KalmanODESolution for example only sees a `KalmanPosterior` object, and is not aware whether it is a "FilteringODESolution" or a "SmoothingODESolution".

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -60,77 +68,42 @@ def __call__(self, t):
...
-
-    def __len__(self):
-        return len(self.locations)
-
-    def __getitem__(self, idx):
-        return self.state_rvs[idx]
+    @abc.abstractmethod
+    def interpolate(self, t):
+        """Evaluate the posterior at a measurement-free point."""
+        raise NotImplementedError
```
pnkraemer, line 95: this is one of two differences in filtering and smoothing posterior -- dense evaluation (interpolation).

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -60,77 +68,42 @@ def __call__(self, t):
...
 
     def sample(self, locations=None, size=()):
+        # In the present setting, only works for sampling from the smoothing posterior.
```
pnkraemer, line 98: In principle, this is only for the smoothing posterior here, but for example the recursion is something I would like to reuse for filtering posteriors. Hence, the choice of implementing sample() here and _single_sample_path() in the subclasses. 

Thoughts?

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -141,11 +114,88 @@ def sample(self, locations=None, size=()):
             [self.sample(locations=locations, size=size[1:]) for _ in range(size[0])]
         )
 
+    @abc.abstractmethod
     def _single_sample_path(self, locations, random_vars):
+        raise NotImplementedError
+
+    # If these two functions can be combined reliably into one
+    # function, please feel free to make the changes.
```
pnkraemer, line 122: the "<" and "<=" would lead to some awkward corner cases IMO. Do you agree?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -295,11 +295,11 @@ def smooth(self, filter_posterior):
...
 
         rv_list = self.smooth_list(
-            filter_posterior,
+            filter_posterior.state_rvs,
             filter_posterior.locations,
         )
-        return KalmanPosterior(
-            filter_posterior.locations, rv_list, self, with_smoothing=True
+        return SmoothingPosterior(
+            filter_posterior.locations, rv_list, self, filter_posterior=filter_posterior
```
nathanaelbosch, line 302: Should we rename these variables to filtering_posterior, too? I don't mind keeping them the way they are, but I wondered what the best practice here is.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: That's quite cool :)

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -124,11 +124,11 @@ def test_rmse_filt_smooth(kalman, data):
     """Assert that smoothing beats filtering beats nothing."""
     obs, times, truth = data
 
-    filter_posterior = kalman.filter(obs, times)
-    smooth_posterior = kalman.smooth(filter_posterior)
+    posterior = kalman.filter(obs, times)
```
nathanaelbosch, line 127: to be honest, I found the previous test more readable^^ (but no need to change back, that's a _very_ minor thing)

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
```diff
@@ -5,10 +5,11 @@
 import probnum.filtsmooth.statespace as pnfss
 import probnum.random_variables as pnrv
 from probnum._randomvariablelist import _RandomVariableList
-from tests.testing import chi_squared_statistic
 
 from .filtsmooth_testcases import car_tracking
 
+# from tests.testing import chi_squared_statistic
```
nathanaelbosch, line 11: This should probably be removed.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
nathanaelbosch: Did I write that?? If yes I have 0 memories of it.. :neutral_face: 

I understand it roughly as "compute the chi2 statistics and check that it lies in the 99% interval" (probably not 99% but just some very large one).

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -60,77 +68,42 @@ def __call__(self, t):
...
         if t < self.locations[0]:
             raise ValueError(
                 "Invalid location; Can not compute posterior for a location earlier "
                 "than the initial location"
             )
 
+        # Early exit if t is in our grid -- no need to interpolate
         if t in self.locations:
-            idx = (self.locations <= t).sum() - 1
+            idx = self._find_previous_or_same_index(t)
```
nathanaelbosch, line 86: Do we actually need "previous or same" here? Since we have `t in self.locations` we're just looking for the index, to find the RV we're looking for.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: I do agree that merging them is not a good idea. But as commented above, I think we don't actually need "previous or same". The code I had before was just a weird implementation of "find the index where t_i == t" (it's probably more efficient than a "where" because the locations are sorted and it was really just the simplest way of implementing this that came to my mind).

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -60,77 +68,42 @@ def __call__(self, t):
         :obj:`RandomVariable`
             Estimate of the states at time ``t``.
         """
+
+        # Recursive evaluation (t can now be any array, not just length 1)
         if not np.isscalar(t):
-            # recursive evaluation (t can now be any array, not just length 1!)
             return _RandomVariableList([self.__call__(t_pt) for t_pt in np.asarray(t)])
```
nathanaelbosch, line 74: The `np.asarray(t)` looks weird actually, we should be able to just loop over `t`.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: I didn't get the "the recursion is something I would like to reuse for filtering posteriors" part. What do you mean by that? I would generally say that if only the `SmoothingPosterior` is able to sample, then `sample` should probably be there.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -124,3 +124,9 @@ def sample(
 
         samples = self.kalman_posterior.sample(locations=t, size=size)
         return np.array([self.proj_to_y @ sample for sample in samples])
+
+    @property
+    def filter_solution(self):
```
nathanaelbosch, line 129: Where is this used / what is the use-case for this property?

nathanaelbosch (changes requested):

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -134,7 +134,7 @@ def _estimate_local_error(
         local_pred_rv = Normal(pred_rv.mean, calibrated_proc_noise_cov)
         local_meas_rv, _ = self.gfilt.measure(local_pred_rv, t_new)
         error = local_meas_rv.cov.diagonal()
-        return np.sqrt(error)
+        return np.sqrt(np.abs(error))
```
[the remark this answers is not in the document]
pnkraemer: In principle, yes. It lead to fairly flaky tests in my case though, so I decided to accept this absolute value.

By the time square-root filtering is around (#316 ), this will be stabilised/undone anyway.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -149,10 +149,11 @@ def _estimate_diffusion(self, meas_rv):
             value problems.
             Statistics and Computing, 2019.
         """
-        std_like = np.linalg.cholesky(meas_rv.cov)
-        whitened_res = np.linalg.solve(std_like, meas_rv.mean)
-        ssq = whitened_res @ whitened_res / meas_rv.size
-        return ssq
+        # symmetrised = 0.5*(meas_rv.cov + meas_rv.cov.T)
```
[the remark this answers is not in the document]
pnkraemer: because I gave up trying to make cholesky work and accepted that LU is the only way here (until square-root stuff #316 )

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
pnkraemer: extracting the filtering ODE solution out of the kalman posterior. `SmoothingPosterior` has an attribute `filter_posterior`, which is accessed directly. Since `KalmanODESolution` wraps around `KalmanPosterior` objects, there is no un-awkward way of accessing the filter solution (the only way would be to manually project the states out of self.kalman_poisterior.filter_posterior) which is avoided with this simple property.


But seeing this, I will add a useful error if `self.kalmanposterior` has no `filterposterior`, i.e. if it is the FilteringPosterior.

pnkraemer: respectively return `self`: 

```python
    @property
    def filter_solution(self):

        if isinstance(self.kalman_posterior, pnfs.FilteringPosterior):
            return self

        # else: self.kalman_posterior is a SmoothingPosterior object, which has the field filter_posterior.
        return KalmanODESolution(
            kalman_posterior=self.kalman_posterior.filter_posterior
        )
```

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: Good idea

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: True. I remember that I did this for a reason (it was me who wrote this recursion, right?) but since I cannot recall why, and the tests pass just fine, I removed it.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: true. Tbh I just wrapped the code there into a function and did not think about simplifying it, but you're right. Changed it.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: I mean this bit:
```python

        if size == ():
            return np.array(
                self._single_sample_path(locations=locations, random_vars=random_vars)
            )

        return np.array(
            [self.sample(locations=locations, size=size[1:]) for _ in range(size[0])]
        )
```
but I can move it down there. if you prefer this.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
pnkraemer: weird stuff happening here. It should have been
```python

    posterior = kalman.filtsmooth(obs, times)

    filtms = posterior.filtering_posterior.state_rvs.mean
    smooms = posterior.state_rvs.mean
```
which is what I changed it to now.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
pnkraemer: but is this really the chi2 statistic??? 

In any case, I removed it. Wont hurt us I reckon.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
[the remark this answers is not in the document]
pnkraemer: though this reminds me to remove the commented out code (which I just did).

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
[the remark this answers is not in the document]
nathanaelbosch: Sure, if it's removed soon anyways then I don't mind having it in this PR :+1:

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
nathanaelbosch: Since it's not used in the filtering posterior right now I'd say it should be moved into the smoothing posterior.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalmanposterior.py
nathanaelbosch: I think it is. But I might be wrong :sweat_smile:

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
nathanaelbosch: Sounds good, thanks for the explanation :)

nathanaelbosch (changes requested):
I believe the only thing left to do was moving the sampling into the smoothing posterior. Or was there anything else? Otherwise this should be good to merge afterwards :)

pnkraemer:
> moving the sampling into the smoothing posterior

Done :) 

> Or was there anything else?

I dont think so.

nathanaelbosch (approved):
Nice work! :))

PR #318, opened 2021-02-13
Taylor-mode initialisation with JAX

pnkraemer:
This PR introduces first steps towards Taylor-mode AD initialisation for IVP solvers. 

This is a very small PR, which essentially drops the Taylormode function into probnum.diffeq, defines some IVP examples in jax, and has some rudimentary doctests and unittests.

This initialisation is not used anywhere yet. I'd propose we build towards this slowly, because in order for this to be done properly, all `diffeq.ODE` objects, respective `diffeq.IVP` object need to be replaced with `problems.InitialValueProblem` objects (which affects quite a lot of code). I wanted to implemented Taylor-mode based on the "new" problem classes though, and use this as a setup to start nudging all implementations there.

It introduces an optional dependencies `jax` and `jaxlib`, which are caught in the `compute_all_derivatives` function, respectively the IVP implementations. They should be installable via `pip install probnum[jax]`, etc.. 

Thoughts? 

### Issue
Jax does not support windows... since it is only an optional dependency, there should be a way around here, no? Hints welcome :) 

### Solution
The dependencies jax and jaxlib are now only dependencies if the system platform is not windows. There are two sets of tests, one for "with jax" and one for "without jax". I am not very confident in my solution though, so please don't hesitate to criticise :)


setup.cfg
```diff
@@ -60,6 +60,8 @@ test_deps =
      # Optional dependencies of ProbNum
      GPy
      matplotlib
+     jax; platform_system!="Windows"
```
pnkraemer, line 63: jax does not support windows

docs/source/public_api/problems.rst
```diff
@@ -8,20 +8,19 @@ Definitions and collection of problems solved by probabilistic numerical methods
...
-********************
+--------------------
 
 Subpackage offering implementations of standard example problems or convenient
 interfaces to benchmark problems for probabilistic numerical methods. These test
 problems are meant for rapid experimentation and prototyping.
 
 probnum.problems.zoo.diffeq
----------------------------
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
pnkraemer, line 18: the docs failed, because the subsection was not consistent somehow. This way it works, but if there is a better fix, let me know.

src/probnum/diffeq/odefiltsmooth/initialise.py
```diff
@@ -0,0 +1,134 @@
...
+    >>> print(vdp.y0)
+    [2. 0.]
+    >>> print(vdp.dy0_all)
+    None
+    >>> vdp = compute_all_derivatives(vdp, order=3)
+    >>> print(vdp.y0)
+    [2. 0.]
+    >>> print(vdp.dy0_all)
+    [    2.     0.     0.    -2.    -2.    60.    60. -1798.]
+    """
```
pnkraemer, line 75: these doctests are the main functionality tests. Other than that, there is no value test and if you think there should be one, please let me know (though, which one?)

src/probnum/diffeq/odefiltsmooth/initialise.py
```diff
@@ -0,0 +1,134 @@
...
+        tmax=ivp.tmax,
+        y0=ivp.y0,
+        df=ivp.df,
+        ddf=ivp.ddf,
+        solution=ivp.solution,
+        dy0_all=all_initial_derivatives,
+    )
+
+
+def _taylormode(f, z0, t0, order):
```
pnkraemer, line 89: this function is "hidden", because it is only used and IMO should not be part of the API (we are not offering taylormode, but just use it for initial values. 

Also, I am aware that the code in this function is fairly messy. TBH I am not 100 percent sure how it does what it does, (a source is provided), and I expect it to be cleaned up in the future. FOr now, I just want to get this initialisation thing started ASAP to make some real progress in diffeq.

src/probnum/problems/zoo/diffeq/_ivp_examples_jax.py
```diff
@@ -0,0 +1,168 @@
...
+from probnum.problems import InitialValueProblem
+
+__all__ = ["threebody_jax", "vanderpol_jax"]
+# pylint: disable=import-outside-toplevel
+
+
+JAX_ERRORMSG = "IVP instantiation requires jax. Try using `threebody()` instead, or install `jax` via `pip install jax jaxlib`"
+
+
+def threebody_jax(tmax=17.0652165601579625588917206249):
```
pnkraemer, line 12: this is not consistent with the formulation in diffeq.ivp_examples, but I prefer it like this and will update the other variant soon.

src/probnum/problems/zoo/diffeq/_ivp_examples_jax.py
```diff
@@ -0,0 +1,168 @@
...
+
+    try:
+        import jax
+        import jax.numpy as jnp
+        from jax.config import config
+
+        config.update("jax_enable_x64", True)
+
+    except ImportError as err:
+        raise ImportError(JAX_ERRORMSG) from err
```
pnkraemer, line 143: in all jax dependent pieces we import it inside each function like this.

src/probnum/problems/zoo/diffeq/_ivp_examples_jax.py
```diff
@@ -0,0 +1,168 @@
...
+        (mu,) = params
+
+    if y0 is None:
+        y0 = np.array([2.0, 0.0])
+
+    def vanderpol_rhs(Y):
+        return jnp.array([Y[1], mu * (1.0 - Y[0] ** 2) * Y[1] - Y[0]])
+
+    df = jax.jit(jax.jacfwd(vanderpol_rhs))
+    ddf = jax.jit(jax.jacrev(df))
```
pnkraemer, line 157: I mean, why not ;)

tests/test_problems/test_zoo/test_diffeq/test_ivp_examples_jax.py
```diff
@@ -0,0 +1,79 @@
+import pytest
+
```
pnkraemer, line 2: I know this test script may be a bit confusing. Please let me know if it is too bad. Maybe someone with more pytest knowledge will be able to improve it?!

nathanaelbosch (approved):
Looks good to me! However, I might not be the best person to judge about the changes in setup.cfg, and about the general way the import of jax is handled. Maybe it makes sense to get another opinion here? But if you're confident enough in those things then it's also fine with me to merge this :)

pnkraemer:
@JonathanWenger would you have a look at the changes in setups and the respective imports please? The rest you can ignore freely :)

tests/test_problems/test_zoo/test_diffeq/test_ivp_examples_jax.py
JonathanWenger: You should run the tests that do not use JAX even if jax is available in my opinion since the CI toolchain will import jax and then these tests will never be run I think.

JonathanWenger (changes requested):

tests/test_problems/test_zoo/test_diffeq/test_ivp_examples_jax.py
pnkraemer: The tests below check that *if jax is not available, the import error is thrown*. On the other hand, the other tests check that *if jax is available, code is being executed*. So they check the same functions in two separate scenarios, and I dont know how to run them anyway here.

pnkraemer: Though if you have any idea how to handle this better, I'd gladly make changes :)

JonathanWenger: Let me try to explain. Currently the CI imports jax (except on windows) and runs the tests that are not skipped if `JAX_AVAILABLE`. However, in my understanding this  means that tests for code that is run if jax is not available (in this case throwing an import error) is actually not tested by the CI. Judging from the coverage this seems to be the case. Here this is not much of a problem but should you have more sophisticated code that users without jax might want to run, then this should be tested explicitly in my opinion. There is probably two ways to do this, either have different test configurations (with and without jax available) or simply tests that call functions which are called when jax is not available explicitly, sth like `diffeq_zoo.threebody_nojax()`. Since in this PR this simply results in an error I think this is good to go, but we might want to keep this in mind in the future.

JonathanWenger (approved):

tests/test_problems/test_zoo/test_diffeq/test_ivp_examples_jax.py
pnkraemer: Ah okay. Yes, I agree with everything said. Let's keep this in mind for the future!

PR #323, opened 2021-02-17
Fix finite difference stepsize

### LINKED ISSUE #322: Possibly non-deterministic test fails on master
The following test currently fails on master in some builds of the package. This suggests there might be some non-determinism in the test.

```bash
tests/test_diffeq/test_ode/test_ivp_examples.py::TestJacobianEvaluation::test_lorenz_jacobian FAILED [  2%]
```

pnkraemer:
`dt=1e-11` was too small to the point where numerical jacobians were unstable.

`dt=1e-6` performs much better (the realised errors are around `1e-8`, so a tolerance of `1e-4` should to the trick.

The problem with Lorenz was the *relative* tolerance of `1e-1` not being guaranteed, because the state values were rather small.

Fixes #322


tests/test_diffeq/test_ode/test_ivp_examples.py
```diff
@@ -157,21 +157,27 @@ class TestJacobianEvaluation(unittest.TestCase, NumpyAssertions):
...
+
     def test_logistic_jacobian(self):
         rv = Constant(0.1)
         lg1 = ivp_examples.logistic(self.tspan, rv)
         random_direction = 1 + 0.1 * np.random.rand(lg1.dimension)
         random_point = 1 + np.random.rand(lg1.dimension)
         fd_approx = (
             0.5
-            * 1e11
+            * 1.0
```
schmidtjonathan, line 170: The `* 1.0` can be removed, do you want to leave it in for didactical reasons?

schmidtjonathan (approved):
✅

tests/test_diffeq/test_ode/test_ivp_examples.py
pnkraemer: yeah kinda (I think it makes it easier that this is the denominator of the fraction), though I am not too set on this opinion.

PR #326, opened 2021-02-23
Refactor Linear Operator

marvinpfoertner:


src/probnum/linops/_linear_operator.py
```diff
@@ -1,11 +1,19 @@
...
+from typing import Optional, Union
 
 import numpy as np
 import scipy.sparse.linalg
 import scipy.sparse.linalg.interface
 
+import probnum.utils
+from probnum.type import DTypeArgType, ScalarArgType, ShapeArgType
+
+BinaryOperandType = Union[
```
JonathanWenger, line 13: Would it make sense to have this under `probnum.type`?

src/probnum/linops/_linear_operator.py
```diff
@@ -129,19 +114,7 @@ def dot(self, x):
...
-        else:
-            if len(x.shape) == 1 or len(x.shape) == 2 and x.shape[1] == 1:
-                return self.matvec(x)
-            elif len(x.shape) == 2:
-                return self.matmat(x)
-            else:
-                raise ValueError(
-                    "Expected 1-d or 2-d array, matrix or random variable, got %r." % x
-                )
+        return self @ x
```
JonathanWenger, line 117: With this choice are we accepting possibly unexpected behavior when multiplying with a (m, n, d) array, right? In this scenario `np.matmul` treats the array as a stack of 2d arrays while `np.dot` treats it as a stack of vectors resulting in sets of inner products.

I am fine with this choice, but maybe we should raise an error here if anything but a 2d array is passed, like in the original implementation.

src/probnum/linops/_linear_operator.py
```diff
@@ -416,40 +390,53 @@ class ScalarMult(LinearOperator):
...
+        self._scalar = probnum.utils.as_numpy_scalar(scalar, dtype=dtype)
 
-    def _matvec(self, x):
-        return self.scalar * x
+        super().__init__(shape=probnum.utils.as_shape(shape), dtype=self._scalar.dtype)
 
-    def _matmat(self, X):
-        return self.scalar * X
+    @property
+    def scalar(self):
```
JonathanWenger, line 404: Not crucial in this PR, but some type hints for these methods are missing. Feel free to ignore for now.

JonathanWenger (approved):

src/probnum/linops/_linear_operator.py
marvinpfoertner: Yeah, I have that in my TODO and will follow up on type hints (and verify them by tests) in one of the next PRs

src/probnum/linops/_linear_operator.py
marvinpfoertner: Personally, I would opt for a little more locality here. To me it would seem a little out of place in `probnum.type`

src/probnum/linops/_linear_operator.py
marvinpfoertner: I agree with your suggestion. In the next PR, I will remove the inheritance from `scipy.sparse.linalg.LinearOperator`. I would suggest to address this problem there, is that alright?

JonathanWenger: I would be in favor of re-adding the `ValueError` already now, just so that in the meantime no accidental reliance on this behaviour is introduced, which then breaks in the next PR.

marvinpfoertner: Ah sorry, confused myself there 😄 The `ValueError` is still there, I just moved it to `__matmul__`

Issue #272, opened 2020-12-02
API of `probsolve_ivp` 


nathanaelbosch:
We might want to improve some parts of `probsolve_ivp`:
- Default values for `atol` and `rtol`: I believe we should have some sensible default values
- Explicit argument to choose the steprule: I think that would be cleaner than inferring it from `atol`/`rtol` or `step` being set

pnkraemer:
potentially wild suggestion: how about we get the square root and initialisation in order and then remove the constant step option entirely from probsolve_ivp? That would also communicate some nice new confidence in the solver :)

PR #328, opened 2021-02-23
Move `statespace` module out of `filtsmooth` into top level

### LINKED ISSUE #282: (Statespace) models
Currently, transition rules and the like are in `filtsmooth.statespace`. I'd like to pull them out and make them their own thing. In this context, I think it might be interesting to rename it from `statespace` to something like `models` (as in: *probabilistic models*) because all those transitions can essentially be used as likelihoods for all kinds of things. 

Maybe at some point, we will be able to remove the filtsmooth dependency from `diffeq` and only have a `statespace` (`models` resp.) dependency.

Thoughts on the renaming, @nathanaelbosch @JonathanWenger ? I guess no one would complain about moving it up...

schmidtjonathan:
Closes #282 .


docs/source/public_api/filtsmooth.rst
```diff
@@ -12,17 +12,3 @@ observations into account.
...
-
-
-probnum.filtsmooth.statespace
-=============================
-
-Probabilistic State Space Models.
-
-This package implements continuous-discrete and discrete-discrete state
-space models, which are the basis for Bayesian filtering and smoothing,
-but also probabilistic ODE solvers.
```
pnkraemer, line 24: What do you think of the following changes of wording?

implements **components of** continuous-discrete and ...

the basis for BFAS, but also **for** probabilistic ODE solvers

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -31,7 +31,7 @@ def forward_realization(
         compute_gain=False,
         _diffusion=1.0,
         _linearise_at=None,
-    ) -> (pnrv.Normal, typing.Dict):
+    ) -> typing.Tuple[pnrv.Normal, typing.Dict]:
```
pnkraemer, line 34: good catch. Thanks.

pnkraemer (approved):
Looks good already, no complaints on my side!  :)

Thanks for doing this!

schmidtjonathan:
@pnkraemer I had a rebase conflict after updating the branch, so it would be nice, if you could check if everything is still alright, before I merge :)
Thanks!

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -362,31 +362,31 @@ def _string2matern(ivp, which_prior, **kwargs):
...
+        return pnss.Matern(
             3,
             ivp.dimension,
             lengthscale,
             forward_implementation="sqrt",
             backward_implementation="sqrt",
         )
     elif which_prior == "matern92":
-        return pnfs.statespace.Matern(
+        return pnss.Matern(
```
pnkraemer, line 389: I know this is annoying now, and that it is not part of your PR, but this line (and the related ones above) are not covered by tests (for some reason, codecov thinks this is your fault) which leads the the diff coverage being <90% (which is why you wont be able to merge I guess). What do you think about adding a quick test for these settings in ``test_odefiltsmooth.TestLotkaVolterraOtherPriors``? Or, alternatively, remove these options (as they've never been used anyway -- in the future, we will have to consider how to handle the order of the prior in a separate issue; see #272.

pnkraemer (commented):
overall, looks good. There is this diff coverage thing, which is why the CI is failing. See my comment below.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
schmidtjonathan: I will add a test for that and then re-request you as a reviewer 👍

Issue #329, opened 2021-02-24
Cholesky factor in Normal


pnkraemer:
### Problem

Whenever `Normal.cov_cholesky` is called and `cov_cholesky` has not been passed to `Normal.__init__`, it is automatically computed (btw, with a damping factor of 1e-12 that cannot be changed from the outside... is that on purpose?). If it has been passed, it is returned.

I cannot find a way of checking from the outside whether a `cov_cholesky` has been passed to `Normal.__init__` or not (the only difference would be that when `cov_cholesky` has been passed as an argument, then `Normal._compute_cov_cholesky` is a lambda function, otherwise it is a "real" function). Right? Or am I missing something here? 

### Motivation

This is important for changes of variables: if `x ~ Normal(m, C)` , then `P x ~ Normal(P m, PCP^T)`, and if there is a Cholesky factor `C=L L^T`, then we also need to update `L -> PL` (modulo some triangularisation, for instance with a QR-decomposition. In my cases, `P` is usually a diagonal matrix in which case it is fine either way). 

This consideration of the Cholesky factor is not done automatically by `P @ x` (see #319), but it is not possible to do from the outside either without always computing a Cholesky factor (of a potentially really ill-conditioned or even singular matrix). This leads to some unnecessary compute time, whenever the Cholesky factors are not really needed.

### Possible solution(s)

I guess possible solutions would be either:
* Introduce a variable `Normal._cholesky_passed_as_argument` or something of the like, which holds the corresponding value  `True` or `False`. I suppose this would be the smallest increment to the current implementation, but feels kind of dirty to me.
* Explicitly introduce a variable `Normal._cov_cholesky_values` which is None by default, and overwritten if either `Normal(..., cov_cholesky)` is filled or if the property `Normal.cov_cholesky` is called (i.e. doing the caching in the cached property
```python
    @cached_property
    def cov_cholesky(self) -> _ValueType:
        """Cholesky factor :math:`L` of the covariance
        :math:`\\operatorname{Cov}(X) =LL^\\top`."""
        if self._compute_cov_cholesky is None:
            raise NotImplementedError

        return self._compute_cov_cholesky()
```
manually. In this case, one could check `if Normal._cov_cholesky_values is not None: do sth`. 

@JonathanWenger @marvinpfoertner you guys know `Normal` better than I do. What do you say? Any other ideas for solving this?

### Other

[...]

PR #331, opened 2021-02-24
Handling of Cholesky factors in Normal

pnkraemer:
In response to #329 . 

The major changes are:
* If provided, the Cholesky factor is stored explicitly (and not indirectly via the cached property) in `self._cov_cholesky_values`.
* If not provided, it is still computed automatically in the property `cov_cholesky` through calling `self.precompute_cov_cholesky(damping_factor=1e-12)` with a default damping factor (the same constant as before).
* `self.precompute_cov_cholesky` can also be called from the outside with a user-specific damping factor.
* A User can check whether a cholesky factor is provided (either through passing it as an argument or through precomputing it) or not via `self.cov_cholesky_is_precomputed` which returns the corresponding bool. 
* Tests for uni, multi, and matrixvariate distribution capture this behaviour. Operatorvariate had some Cholesky-compute oddities, which is why I did not do this. They can surely be extended in a straightforward manner by someone that knows Operator-variate Normals :)

In general: I hope I understood the behaviour in `Normal.__init__` correctly. If not, please let me know.


src/probnum/random_variables/_normal.py
```diff
@@ -174,9 +176,9 @@ def __init__(
             var = self._dense_var
             entropy = self._dense_entropy
 
-            if cov_cholesky is None:
-                self._compute_cov_cholesky = self.dense_cov_cholesky
-            else:
+            self._compute_cov_cholesky = self.dense_cov_cholesky
```
pnkraemer, line 179: even if cov_cholesky is provided in the ``__init__``, this remains the default behaviour for dense or cov_operator settings (see line 165). Now, even if a Cholesky factor is provided, calling `Normal.precompute_cov_cholesky()` directly computes a new one

src/probnum/random_variables/_normal.py
```diff
@@ -139,6 +139,7 @@ def __init__(
         self._cov = cov
 
         self._compute_cov_cholesky: Callable[[], _ValueType] = None
+        self._cov_cholesky_values = cov_cholesky  # recall: None if not provided
```
pnkraemer, line 142: default is None. If provided, the values of the Cholesky factors are stored here. If not, this variable is changed by `self.precompute_cov_cholesky()`

src/probnum/random_variables/_normal.py
```diff
@@ -257,14 +258,30 @@ def __init__(
...
+    @property
     def cov_cholesky(self) -> _ValueType:
         """Cholesky factor :math:`L` of the covariance
         :math:`\\operatorname{Cov}(X) =LL^\\top`."""
-        if self._compute_cov_cholesky is None:
-            raise NotImplementedError
 
-        return self._compute_cov_cholesky()
+        if not self.cov_cholesky_is_precomputed:
+            self.precompute_cov_cholesky()
```
pnkraemer, line 267: this replaces the `cached_property` decorator.

src/probnum/random_variables/_normal.py
```diff
@@ -257,14 +258,30 @@ def __init__(
...
+        return self._cov_cholesky_values
+
+    def precompute_cov_cholesky(self, damping_factor=COV_CHOLESKY_DAMPING):
```
pnkraemer, line 270: the default damping parameter is the same magic constant as before, only now it is possible to change it by calling precompute_cov_cholesky manually (see tests).

pnkraemer: I personally prefer `precompute_*` over `compute_*`because it suggests more strongly that it is a value that is cached for later, and might also sound a bit more like something that happens in-place.

I would be willing to change the name though, if that is preferred.

src/probnum/random_variables/_normal.py
```diff
@@ -257,14 +258,30 @@ def __init__(
...
+        """Overwrites values of cov_cholesky with new Cholesky factors (careful: in-
+        place operation!)."""
+        self._cov_cholesky_values = self._compute_cov_cholesky(
+            damping_factor=damping_factor
+        )
+
+    @property
+    def cov_cholesky_is_precomputed(self) -> bool:
```
pnkraemer, line 277: easy-access flag to see whether the Cholesky factor is already cached.

`if Normal.cov_cholesky_is_precomputed` reads a bit better than `if Normal._cov_cholesky_values is not None`.

src/probnum/random_variables/_normal.py
```diff
@@ -257,14 +258,30 @@ def __init__(
...
+        """Whether the Cholesky factor of the covariance is readily available, which
+        happens if either cov_cholesky is specified during initialization or if the
+        property has been called."""
+        if self._cov_cholesky_values is None:
+            return False
+        return True
```
pnkraemer, line 287: this could be done in 1 line, but I personally prefer the readibility of this if-clause.

src/probnum/random_variables/_normal.py
```diff
@@ -413,8 +430,10 @@ def _sub_normal(self, other: "Normal") -> "Normal":
         )
 
     # Univariate Gaussians
-    def _univariate_cov_cholesky(self) -> np.floating:
-        return np.sqrt(self._cov)
+    def _univariate_cov_cholesky(
+        self, damping_factor=COV_CHOLESKY_DAMPING
```
pnkraemer, line 437: all of these functions have a default damping parameter now. Similarly to before, only that now it is easier to change it.

tests/test_random_variables/test_normal.py
```diff
@@ -413,6 +415,56 @@ def test_transpose(self):
...
+        """
+        mean, cov = self.params
+        rv = rvs.Normal(mean, cov)
+
+        with self.subTest("No Cholesky precomputed"):
+            self.assertFalse(rv.cov_cholesky_is_precomputed)
+
+        with self.subTest("Without explicit damping"):
+            # damping factor 1e-12 does not mess up this test
+            self.assertAllClose(rv.cov_cholesky, np.sqrt(rv.cov))
```
pnkraemer, line 432: the cov_cholesky is accessible in the usual way.

tests/test_random_variables/test_normal.py
```diff
@@ -413,6 +415,56 @@ def test_transpose(self):
...
+
+        with self.subTest("With explicit damping"):
+            # Sanity check: usual cov_cholesky does not work
+            with self.assertRaises(AssertionError):
+                self.assertAllClose(rv.cov_cholesky, np.sqrt(rv.cov + 10.0))
+
+            rv.precompute_cov_cholesky(damping_factor=10.0)
```
pnkraemer, line 439: it can be computed manually from outside though, with any damping.

tests/test_random_variables/test_normal.py
```diff
@@ -413,6 +415,56 @@ def test_transpose(self):
...
+        with self.subTest("Cholesky precomputed"):
+            self.assertTrue(rv.cov_cholesky_is_precomputed)
+
+        with self.subTest("Returns correct cov_cholesky"):
+            self.assertAllClose(rv.cov_cholesky, cov_cholesky)
+
+        with self.subTest("self.precompute overwrites with true factor"):
+            # Sanity check: usual cov_cholesky does not work
+            rv.precompute_cov_cholesky()
+            self.assertAllClose(rv.cov_cholesky, np.sqrt(rv.cov))
```
pnkraemer: I suppose this "overwrite factor even a Cholesky factor has been provided initially" is an API decision that one can argue about. 
Let me know your thoughts please.

src/probnum/random_variables/_normal.py
marvinpfoertner: Why is this in the plural? Can we have multiple Cholesky decompositions in one Normal random variable?

src/probnum/random_variables/_normal.py
marvinpfoertner: Why is that?

src/probnum/random_variables/_normal.py
marvinpfoertner: I agree

tests/test_random_variables/test_normal.py
marvinpfoertner: Personally, I would not allow overwriting the provided Cholesky factor. What are your thoughts, @JonathanWenger?

JonathanWenger: I agree. Random variables should be as immutable as possible, barring their random seed imo.

pnkraemer: Alright. Then I'd suggest throwing an error (alternative would be a `pass`, which is a bit odd?!).

src/probnum/random_variables/_normal.py
pnkraemer: because it is a matrix... somehow `value` felt wrong. I do not have a strong opinion though, so if you prefer "value", I will change it.

src/probnum/random_variables/_normal.py
pnkraemer: I guess this is resolved in the discussion below?

tests/test_random_variables/test_normal.py
pnkraemer: ummmm which one? :D TypeError? RuntimeError? (a list is in https://docs.python.org/3/library/exceptions.html)

marvinpfoertner: I think I would just throw a plain `Exception` since none of the standard exceptions really fit here

src/probnum/random_variables/_normal.py
marvinpfoertner: Yup, thanks for clarifying!

src/probnum/random_variables/_normal.py
marvinpfoertner: Personally, I would just call it `_cov_cholesky`, is there anything preventing this naming?

pnkraemer: Not strictly, there is just so many methods associated with the Cholesky that I thought it did not hurt to clarify that this one holds the array (if it exists). 

Do you still prefer just `_cov_cholesky`? If so, I will change it.

marvinpfoertner: I would prefer `_cov_cholesky`, but I usually have a strong opinion on naming, so feel free to ignore it 😃 Nevertheless, here's my reasoning: I feel like the internal variable storing the cached value of the public property `cov_cholesky` should be named `_cov_cholesky` to associate the two with one another.

pnkraemer: Done

src/probnum/random_variables/_normal.py
```diff
@@ -174,29 +182,30 @@ def __init__(
...
 
-            if cov_cholesky is None:
-                self._compute_cov_cholesky = self.dense_cov_cholesky
-            else:
-                if not isinstance(cov_cholesky, type(self._cov)):
+            self._compute_cov_cholesky = self.dense_cov_cholesky
+
+            # Ensure that the Cholesky factor has the same type as the covariance,
+            # and, if necessary, promote data types. Check for (in this order): type, shape, dtype.
+            if self._cov_cholesky is not None:
```
pnkraemer, line 189: took me a bit to get this code block, so I thought I'd document what I learned ;)

src/probnum/random_variables/_normal.py
```diff
@@ -174,29 +182,30 @@ def __init__(
...
 
-                if cov_cholesky.dtype != self._cov.dtype:
+                if self._cov_cholesky.dtype != self._cov.dtype:
                     # TODO: Implement casting for linear operators
-                    if not isinstance(cov_cholesky, linops.LinearOperator):
-                        cov_cholesky = cov_cholesky.astype(self._cov.dtype)
-
-                self._compute_cov_cholesky = lambda: cov_cholesky
+                    if not isinstance(self._cov_cholesky, linops.LinearOperator):
+                        self._cov_cholesky = self._cov_cholesky.astype(self._cov.dtype)
```
pnkraemer, line 208: I am not sure how robust it is to always promote the Cholesky data type to that of the covariance. For instance, I'd think it is perfectly okay to specify an integer variance (say 2), which has a non-integer Cholesky (sqrt(2)). If this is promoted, information is lost. 

In any case, I did not feel confident enough to change things here, so instead I included this promotion of types into the docstring (in case people use it and are wondering where information is lost).

src/probnum/random_variables/_normal.py
```diff
@@ -48,6 +48,12 @@ class Normal(_random_variable.ContinuousRandomVariable[_ValueType]):
         Mean of the random variable.
     cov :
         (Co-)variance of the random variable.
+    cov_cholesky :
+        (Lower triangular) Cholesky factor of the covariance matrix. If None, then the Cholesky factor of the covariance matrix
+        is computed when :attr:`Normal.cov_cholesky` is called and then cached. If specified, the value is returned by :attr:`Normal.cov_cholesky`.
+        In this case, its type and data type are compared to the type and data type of the covariance.
+        If the types do not match, an exception is thrown. If the data types do not match,
+        the data type of the Cholesky factor is promoted to the data type of the covariance matrix.
```
pnkraemer, line 56: see the comment below.

src/probnum/random_variables/_normal.py
marvinpfoertner: In the long run, we should probably restrict users from using non-float dtypes here, but for now I'd say it's fine.

marvinpfoertner (approved):

src/probnum/random_variables/_normal.py
pnkraemer: I agree, but the same conversation might come up in terms of `float32` and `float64`. Though, this is neither relevant for this PR nor sufficiently urgent to raise an issue (especially since its documented fairly prominently now).

tests/test_random_variables/test_normal.py
```diff
@@ -584,6 +584,35 @@ def test_cov_cholesky_cov_cholesky_passed(self):
...
                 rv.precompute_cov_cholesky()
 
+    def test_cholesky_cov_incompatible_types(self):
+        """Test the behaviour of Normal.__init__ in the setup where the type of the
+        Cholesky factor and the type of the covariance do not match."""
+        mean, cov = self.params
+        cov_cholesky = np.linalg.cholesky(cov)
+        cov_cholesky_wrong_type = cov_cholesky.tolist()
+        with self.subTest("Different type raises ValueError"):
+            with self.assertRaises(ValueError):
```
marvinpfoertner, line 594: I feel like this should raise a `TypeError`

marvinpfoertner (approved):

marvinpfoertner (approved):
Very nice, this touched up on a lot of legacy problems, thanks! :)

PR #332, opened 2021-02-24
Fix doctest collection

### LINKED ISSUE #325: Doctests not collected by PyTest
Currently, the our doctests in the source tree are not collected by pytest.
This happens both in the CI action and locally.

JonathanWenger:
# In a Nutshell
For the last few commits doctests were not run anymore. This PR fixes the test collection.

# Fix in Detail
This problem occurred since I erroneously had removed the `src` directory from the test path in an attempt to fix an `ImportPathError` which is explained in detail here: https://github.com/tox-dev/tox/issues/373 as part of #263. Essentially 
when using `tox` with `pytest`, then the following scenario arises:

> the problem is - your test imports the installed module thats inside the virtualenv
> afterwards doctest-modules comes along and takes a look at the source modules trying to import them failing with a mismatch

The current recommended fix by the `tox` developers is to use the `testenv` option `usedevelop = True`.

# Other
Additionally I moved the pytest config out of `setup.cfg` in a steady migration towards `pyproject.toml`.

This closes #325

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,6 +5,8 @@
 [tool.pytest.ini_options]
 addopts = "--verbose --doctest-modules"
 norecursedirs = [
+    ".*",
+    "*.egg*",
     "dist",
     "build",
     ".tox"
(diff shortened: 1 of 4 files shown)


nathanaelbosch (approved):
Looks good to me. Thanks for the fix, and for the helpful PR description :)

PR #334, opened 2021-02-25
Fallback for initialisation without jax

pnkraemer:
This PR improves initialisation of ODE solvers in the case where no jax is available (which is default).


src/probnum/diffeq/__init__.py
```diff
@@ -33,9 +24,6 @@
     "fitzhughnagumo",
     "lotkavolterra",
     "seir",
-    "rigidbody",
-    "vanderpol",
-    "threebody",
```
pnkraemer, line 38: I moved those three to `probnum.problems.zoo`, because (i) vanderpol and threebody are used in doctests to show a comparison between initialisation with RK and initialisation with TM; (ii) rigidbody has a jacobian, which helps with setting up the unit test structure in the problem module. Neither of those three was used anywhere else before, so no functionality is lost.

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+        InitialValueProblem object where the attribute `dy0_all` is filled with an
+        :math:`d(\nu + 1)` vector, where :math:`\nu` is the specified order,
+        and :math:`d` is the dimension of the IVP.
+
+
+    Examples
+    --------
+    >>> import sys, pytest
+    >>> if sys.platform.startswith('win'):
+    ...     pytest.skip('this doctest does not work on Windows')
```
pnkraemer, line 55: pytest checks the doctests, but this doctest here does not work on windows (bc. of jax).

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+        y0=ivp.y0,
+        df=ivp.df,
+        ddf=ivp.ddf,
+        solution=ivp.solution,
+        dy0_all=all_initial_derivatives,
+        dy0_errors=errors,
+    )
+
+
+SMALL_VALUE = 1e-28
```
pnkraemer, line 124: this value is added to the marginal stds of the initial derivatives that are known. If we put in zero, there are linalg errors (because a zero-cov RV is conditioned on a dirac likelihood). This value is chosen such that its square-root is a really small damping factor).

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def compute_all_derivatives_via_rk(f, z0, t0, order, df=None, h0=1e-2, method="DOP853"):
+    r"""Compute derivatives of the initial value by fitting an integrated Brownian motion process to a few steps of an approximate ODE solution.
+
+
+    It goes as follows:
```
pnkraemer, line 131: does the below make sense?

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+
+    1. The ODE integration problem is set up on the interval ``[t0, t0 + (2*order+1)*h0]``
```
pnkraemer, line 133: the choice 2q+1 is rather arbitrary. though I vaguely remember from some old Nordsieck and gear paper that q+1 steps are necessary. I found the performance to be improved with a few more steps than that, so 2q+1 is just for good measure.

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+    t_eval = np.arange(t0, t0 + (num_steps + 1) * h0, h0)
+    sol = sci.solve_ivp(
+        f,
+        (t0, t0 + (num_steps + 1) * h0),
+        y0=z0,
+        atol=1e-12,
+        rtol=1e-12,
+        t_eval=t_eval,
+        method=method,
+    )
```
pnkraemer, line 203: this and the code below is a bit messy, but is it still clear what is going on?

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+        return dz_t
+
+    z_shape = z0.shape
+    z_t = jnp.concatenate((jnp.ravel(z0), jnp.array([t0])))
+
+    derivs = []
+
+    derivs.extend(z0)
+    if order == 0:
+        all_derivs = pnss.convert_derivwise_to_coordwise(
```
pnkraemer, line 295: the taylormode implementation uses a different state-derivative ordering than we do in probnum. Therefore the conversion function. I found that it was useful to have at a few places (like, here as well as in some tests), so it is public and tested etc.

src/probnum/filtsmooth/statespace/integrator.py
```diff
@@ -54,6 +106,58 @@ def proj2coord(self, coord: int) -> np.ndarray:
...
+        """
+        dim = (self.ordint + 1) * self.spatialdim
+        projmat = np.zeros((dim, dim))
+        E = np.eye(dim)
+        for q in range(self.ordint + 1):
+
+            projmat[q :: (self.ordint + 1)] = E[
+                q * self.spatialdim : (q + 1) * self.spatialdim
+            ]
+        return projmat
```
pnkraemer, line 137: I find it easier to think of this reshaping in terms of projection matrices (inverses are transposes). it is a tiny bit slower I suppose, but not used in any performance-reliant code anyway.

src/probnum/problems/zoo/diffeq/_ivp_examples.py
```diff
@@ -0,0 +1,186 @@
+import numpy as np
+
+from probnum.problems import InitialValueProblem
+
+__all__ = ["threebody", "vanderpol", "rigidbody"]
+
+
+def threebody(t0=0.0, tmax=17.0652165601579625588917206249, y0=None):
```
pnkraemer, line 8: this interface is "new", compared to the old threebody implementation before. This way one can just call `threebody()` which I found useful already.

tests/test_diffeq/test_odefiltsmooth/_known_initial_derivatives.py
```diff
@@ -0,0 +1,126 @@
+"""Known derivatives of initial values for ODE example problems that are used to test
+the methods in diffeq/odefiltsmooth/initialize.py.
+
+Other than in the ``Integrator`` objects, which stores values coordinate-wise, that is, as ``(y1,
+dy1, ddy1, ..., y2, dy2, ddy2, ...)``,  the values here are stored derivative-wise, that is, as
+``(y1, y2, dy1, dy2, ddy1, ddy2, ...))``. The reason is that in order to
+extract "the first few derivatives", we can conveniently slice the first
+``(d*(q+1))`` entries and reshape accordingly. Otherwise, we would have
+to slice the blocks (:(q+1)), (d*15:d*15+q+1), ...
```
pnkraemer, line 9: does this paragraph make sense?

tests/test_diffeq/test_odefiltsmooth/test_initialize.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@pytest.fixture
+def lv_inits(order):
+    lv_dim = 2
+    vals = LV_INITS[: lv_dim * (order + 1)]
+    return pnss.convert_derivwise_to_coordwise(vals, ordint=order, spatialdim=lv_dim)
+
+
+def test_initialize_with_rk(lv, lv_inits, order):
```
pnkraemer, line 63: even though these kinds of tests are a bit messy currently, they provide the template for any sorts of tests for these implementations. If for instance we do some IEKS initialisation at some point, it can be tested just like this here.

tests/test_problems/test_zoo/test_diffeq/test_ivp_examples.py
```diff
@@ -0,0 +1,45 @@
...
+        increment = step * direction
+        point = ivp.y0 + 0.1 * np.random.rand(len(ivp.y0))
+
+        fd_approx = (
+            ivp.f(time, point + increment) - ivp.f(time, point - increment)
+        ) / (2.0 * step)
+
+        np.testing.assert_allclose(
+            fd_approx, ivp.df(time, point) @ direction, rtol=1e-3, atol=1e-3
+        )
```
pnkraemer, line 45: this is the test for all the IVPs now. To add a new IVP, one adds another instance to the decorator (which is a shortcut for mark parametrise).

pnkraemer (commented):
I think the PR is ready for review now. Please let me know if there is anything wrong. Mostly, I'd like to know whether the logic behind the changes is clear. Thanks :)

nathanaelbosch (changes requested):
Thanks a lot for the PR! Great to have alternatives to the autodiff initialization.

I have a general point I'd like to discuss though, about how to interpret this state initialization.
The way I think about it, it is a procedure to, given a prior x(t) with initial state x(t_0) (e.g. standard normal), and an ivp, condition x on "something", in order to get a better (posterior) x(t_0) (ideally with perfect derivatives and zero cov). 
The code, to me, reflects a different perspective, something like "given an IVP, tell me about the derivatives of the initial value". The fact that the RK version conditions an IBM prior seems like an artifact, only used to extract these derivative values, which are then used to set the set the mean and cov of another prior (could also be ioup or matern) to those numbers.
So first of all, does this make sense? If yes, how do you think about it? If the first description corresponds more closely to the way we think about it, I would ideally like ot reflect this somehow in the code. We can also talk about it in the office tomorrow :)

src/probnum/diffeq/odefiltsmooth/ivp2filter.py
```diff
@@ -107,81 +108,13 @@ def ivp2ukf(ivp, prior, evlvar):
...
-    except NotImplementedError:
-        hess = np.zeros((len(x), len(x), len(x)))
-    if np.isscalar(evl):
-        evl = np.array([evl])
-        jac = np.array([jac])
-        hess = np.array([hess])
-    return (hess @ evl) @ evl + jac @ (jac @ evl)
+def _initial_random_variable(ivp, prior):
+    """Initialize derivatives of the initial value with a Runge-Kutta method."""
+    m0, s0 = compute_all_derivatives_via_rk(
```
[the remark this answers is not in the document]
pnkraemer: yes. At least until we find something better ;)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -177,7 +177,7 @@ def probsolve_ivp(
     >>> from probnum.diffeq import logistic, probsolve_ivp
     >>> from probnum import random_variables as rvs
     >>> import numpy as np
-    >>> initrv = rvs.Constant(0.15)
+    >>> initrv = rvs.Constant(np.array([0.15]))
```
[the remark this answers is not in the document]
pnkraemer: could be because scipy's solve_ivp does not like non-arrays as arguments

src/probnum/problems/_problems.py
```diff
@@ -107,6 +107,7 @@ class InitialValueProblem:
     ddf: typing.Optional[typing.Callable[[float, np.ndarray], np.ndarray]] = None
 
     dy0_all: typing.Optional[np.ndarray] = None
+    dy0_errors: typing.Optional[np.ndarray] = None
```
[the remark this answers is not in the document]
pnkraemer: "measurement error" for the initial derivatives. If they are assumed to be zero (and potentially incorrect), there will be numerical issues (condition a dirac on another dirac). For TM they are zero, but for RK they are not.

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+        proj_to_y,
+        zeros_shift,
+        zeros_cov,
+        proc_noise_cov_cholesky=zeros_cov,
+        forward_implementation="sqrt",
+        backward_implementation="sqrt",
+    )
+
+    num_steps = (
+        2 * order
```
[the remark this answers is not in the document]
pnkraemer: You're right...

src/probnum/diffeq/odefiltsmooth/initialize.py
pnkraemer: In principle, I'd be happy about using the other representation (in which case we wouldnt need the old representation anymore). It is a bit of a project to change it though. 

I actually had this potential future switch in mind, when I decided to make these conversion functions public

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,318 @@
...
+    derivs.extend(y0[:-1])
+    if order == 1:
+        all_derivs = pnss.convert_derivwise_to_coordwise(
+            jnp.array(derivs), ordint=1, spatialdim=len(z0)
+        )
+
+        return all_derivs, jnp.zeros(len(derivs))
+
+    order = order - 2
+    for _ in range(order + 1):
```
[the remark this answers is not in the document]
pnkraemer: do you mean having a case distinction for ``order==0`` and another one for ``order > 0``?

src/probnum/diffeq/odefiltsmooth/initialize.py
pnkraemer: Decision: make these "private" methods of `Integrator`; use wherever necessary, but don't make them too public

pnkraemer:
Decisions from earlier: 

* Make the initialisation functions act on specific priors, and return an RV
* Call them inside ODESolver.initialize()
* Pass the choice between initialisation methods as a string into ``ODESolver.__init__``
* Remove fields `dy0_all` and `dy0_errors` from `InitialValueProblem`.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -17,28 +24,172 @@ class GaussianIVPFilter(ODESolver):
...
+        For more convenient construction, consider :func:`GaussianIVPFilter.from_rk_init`.
     """
 
-    def __init__(self, ivp, gaussfilt, with_smoothing):
-        if not isinstance(gaussfilt.dynamics_model, pnfs.statespace.Integrator):
+    def __init__(
+        self,
+        ivp: IVP,
+        prior: pnfs.statespace.Integrator,
+        measurement_model: pnfs.statespace.DiscreteGaussian,
```
pnkraemer, line 44: No more Kalman initialisation.

This came about, because (according to our discussion) I removed the need to compute an initrv before initialising the filter, which made it awkward to initialise the ODE solver with a Kalman object. Since this is more in the direction of where I want to go with the solver API, I made the change now.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -17,28 +24,172 @@ class GaussianIVPFilter(ODESolver):
...
+            [typing.Callable, np.ndarray, float, pnfs.statespace.Integrator], Normal
+        ],
+    ):
+
+        initrv = Normal(
+            np.zeros(prior.dimension),
+            np.zeros((prior.dimension, prior.dimension)),
+            cov_cholesky=np.zeros((prior.dimension, prior.dimension)),
+        )
+        self.gfilt = pnfs.Kalman(
```
pnkraemer, line 56: internally, not much has changed. Only the API is different.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -17,28 +24,172 @@ class GaussianIVPFilter(ODESolver):
...
         self.with_smoothing = with_smoothing
-        super().__init__(ivp=ivp, order=gaussfilt.dynamics_model.ordint)
+        self.init_implementation = init_implementation
+        super().__init__(ivp=ivp, order=prior.ordint)
+
+    @staticmethod
+    def string_to_measurement_model(
+        measurement_model_string, ivp, prior, measurement_noise_covariance=0.0
+    ):
+        """Construct a measurement model :math:`\\mathcal{N}(g(m), R)` for an ODE.
```
pnkraemer, line 73: this replaces everything that used to be in `ivp2filter`. Once the PR for the probsolve_ivp interface is done, this function will be much leaner.

I decided to move it here as a staticmethod to make absolutely clear that this method belongs to the ODE filter.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -17,28 +24,172 @@ class GaussianIVPFilter(ODESolver):
...
+    @classmethod
+    def from_rk_init(
+        cls,
+        ivp,
+        prior,
+        measurement_model,
+        with_smoothing,
+        init_h0=0.01,
+        init_method="DOP853",
+    ):
```
pnkraemer, line 154: I personally found it cleaner to choose initialisation with these classmethods, while demanding the abstract interface (a function that implements an initialisation) in the __init__. The reason is mostly that `from_rk_init` has a few special parameters, which are not relevant for the other method(s). and  that this way it will be easier to try a new implementation method by writing the function and "plugging it into the ODE solver" without touching the source code.

tests/test_diffeq/test_odefiltsmooth/test_ivp2filter.py
```diff
@@ -1,9 +1,9 @@
...
+"""While previously, this module contained the tests for functions in the
+`diffeq.odefiltsmooth.ivp2filter` module, since this module has become obsolete, we test
+its replacement (`GaussianIVPFilter.string_to_measurement_model`) here.
 
-Covers:
-    * Are the output ExtendedKalman/UnscentedKalman objects?
-    * Does the measurement model do what we think it does
-    * Are the initial values initialised truthfully (y0, f(y0), Jf(y0)f(y0), ...)
+With the next refactoring of `probsolve_ivp` and its tests, please
+refactor this module, too.
```
pnkraemer, line 6: I will update this in the `probsolve_ivp` interface PR, since I touch all the tests in odefiltsmooth there anyway.

tests/test_diffeq/test_odefiltsmooth/test_ivp2filter.py
```diff
@@ -70,36 +58,24 @@ class TestIvp2Ekf1(Ivp2FilterTestCase):
...
-            filtsmooth_object.measurement_model, pnfs.DiscreteEKFComponent
+        measmod = pnd.GaussianIVPFilter.string_to_measurement_model(
+            "ekf1", self.ivp, self.prior, self.evlvar
         )
+        self.assertIsInstance(measmod, pnfs.DiscreteEKFComponent)
 
     def test_ekf1_measmod(self):
-        kalman = pnd.ivp2ekf1(self.ivp, self.prior, self.evlvar)
+        measmod = pnd.GaussianIVPFilter.string_to_measurement_model(
+            "eks1", self.ivp, self.prior, self.evlvar
```
pnkraemer, line 68: I call it once each with "ekf1" once and with "eks1" to cover all the possible branches in the dict. Once the tests are refactored, this will be made more clever.

tests/test_diffeq/test_odefiltsmooth/test_ivp2filter.py
```diff
@@ -70,36 +58,24 @@ class TestIvp2Ekf1(Ivp2FilterTestCase):
...
-                self.ivp.initrv.mean,
-                self.ivp(self.ivp.t0, self.ivp.initrv.mean),
-                self.ivp.jacobian(self.ivp.t0, self.ivp.initrv.mean)
-                @ self.ivp(self.ivp.t0, self.ivp.initrv.mean),
-            ]
-        )
-        self.assertAllClose(filtsmooth_object.initrv.mean, expected_initval.T.flatten())
-
 
 class TestIvpUkf(Ivp2FilterTestCase):
```
pnkraemer, line 80: those test case names remain the same for now I suggest

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -461,28 +442,9 @@ def test_filter_ivp_mat52_ekf(self):
...
-        matrix, apparently."""
-        probsolve_ivp(
-            self.ivp,
-            atol=self.tol,
-            rtol=self.tol,
-            evlvar=0.01,
-            which_prior="matern72",
-            method="uks",
-        )
-
```
pnkraemer, line 475: I removed some of the high order Matern priors with UK observations, because the lack of evlvar parameter makes these numerically unstable.

tests/test_filtsmooth/test_statespace/test_integrator.py
```diff
@@ -288,3 +288,66 @@ def test_forward_realization_values(
         )
         np.testing.assert_allclose(ah_22_ibm @ real, rv.mean)
         np.testing.assert_allclose(diffusion * qh_22_ibm, rv.cov)
+
+
+############################################################################################
+# Tests for the coordinate-representation conversion functions.
+############################################################################################
```
pnkraemer, line 295: they're private, but remain tested

tox.ini
```diff
@@ -63,8 +63,8 @@ deps = pylint == 2.6
 ignore_errors = true
 commands =
          # pylint src --jobs=0
-         pylint src/probnum/diffeq --disable="protected-access,abstract-class-instantiated,too-many-locals,too-few-public-methods,too-many-arguments,unused-argument,missing-module-docstring,missing-function-docstring" --jobs=0
-         pylint src/probnum/filtsmooth --disable="duplicate-code,no-self-use,too-many-locals,too-many-arguments,unused-argument,missing-module-docstring,missing-function-docstring" --jobs=0
+         pylint src/probnum/diffeq --jobs=0
+         pylint src/probnum/filtsmooth --jobs=0
```
pnkraemer, line 67: :partying_face:

pnkraemer:
I implemented all the changes we discussed on Friday (I think). Some of them lead to a bit of refactoring down the road -- the `ivp2filter` module is history now, and the `GaussianIVPFilter` is not initialized with a `Kalman` object anymore now.

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,284 @@
...
+
+
+import numpy as np
+import scipy.integrate as sci
+
+import probnum.filtsmooth as pnfs
+import probnum.random_variables as pnrv
+from probnum.filtsmooth import statespace as pnss
+
+__all__ = [
```
nathanaelbosch, line 12: I know I already asked a long time ago, and this has nothing to do with the PR, but: Why exaclty is this `__all__` needed here? Those are the only two functions in this file anyways, so `from initialize import *` should already have the desired behaviour?

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,284 @@
...
+    h0
+        Maximum step-size to use for computing the approximate ODE solution. The smaller, the more accurate, but also, the smaller, the less stable.
+        The best value here depends on the ODE problem, and probably the chosen method. Optional. Default is ``1e-2``.
+    method
+        Which solver to use. This is communicated as a string that is compatible with ``scipy.integrate.solve_ivp(..., method=method)``.
+        Optional. Default is `DOP853`.
+
+    Returns
+    -------
+    array
```
nathanaelbosch, line 62: If I understood things correctly, it actually returns a Normal now :)

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -0,0 +1,284 @@
...
+    y0
+        Initial value.
+    t0
+        Initial time point.
+    prior
+        Prior distribution used for the ODE solver. For instance an integrated Brownian motion prior (``IBM``).
+
+    Returns
+    -------
+    array
```
nathanaelbosch, line 171: Same here, should be a Normal if I'm not mistaken

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -17,28 +24,172 @@ class GaussianIVPFilter(ODESolver):
...
+        measurement_model: pnfs.statespace.DiscreteGaussian,
+        with_smoothing: bool,
+        init_implementation: typing.Callable[
+            [typing.Callable, np.ndarray, float, pnfs.statespace.Integrator], Normal
+        ],
+    ):
+
+        initrv = Normal(
+            np.zeros(prior.dimension),
+            np.zeros((prior.dimension, prior.dimension)),
```
nathanaelbosch, line 53: Shouldn't this be a standard normal? I'm even surprised that this worked^^

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Looks fine :+1: In case this change was not necessary for this initialization PR, I think it would have been better to include it in the other PR, just to keep this one strictly on-topic. But it's still a good change of course :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: I like the modularity of passing functions :+1: what about `with_rk_init` and `with_taylormode_init`? Personally, that's a bit closer to how I would describe what's happening. On the other hand, "from" seems to be a thing with classmethods. So, no strong opinion, use whichever you prefer :)

nathanaelbosch (changes requested):
Looks good! I made a few comments still and there are a few minor changes, but then we should soon be good to go :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: Since we decided to move the initialisation into `ODESolver.initialize()`, the __init__ had to be thought through again in terms of taking prior and measurement model (not initrv). since the `ivp2filter` functions returned the full filter, I considered it easier to write this function instead of refactoring them in order to return the correct things (which would've been removed in the next step anyway).

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: I see what you're saying here, but AFAIK `from*` is what is used with class methods (everyone that sees `Bla.from_xyz` gets the idea that it is a creation method.

src/probnum/diffeq/odefiltsmooth/initialize.py
pnkraemer: it makes sure that e.g. `pnrv` is not available through importing `initialize`. (And, as a mini side result, it gives some documentation)

src/probnum/diffeq/odefiltsmooth/initialize.py
```diff
@@ -272,11 +271,11 @@ def total_derivative(z_t):
 
     order = order - 2
```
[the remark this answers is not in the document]
pnkraemer: Good idea! Done.

src/probnum/diffeq/odefiltsmooth/initialize.py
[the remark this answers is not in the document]
pnkraemer: Hmmm in principle, I guess so. It is not suuuper easy though, because all the calls to `jet` are different. How about we postpone cleaning up this function?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: It's not really used, but merely a placeholder so that a Kalman object can be initialised. How should I solve this? Put a comment here? Change zeros to NaNs and comment this here?

src/probnum/diffeq/odefiltsmooth/initialize.py
nathanaelbosch: As far as I know, it is available as `import initialize; initialize.pnrv`, but not when importing with `from initialize import *`?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Ok, wondering about this, commenting just as a reminder but I'll come by to discuss

nathanaelbosch (approved):
Great work! Thanks again for the PR, and also for your patience with my comments :)

Issue #335, opened 2021-03-01
Random variable arithmetic and Cholesky factors


pnkraemer:
TL;DR: This is a more thorough version of #319. 

With #331, it is easier to check from the outside whether a Normal RV has a Cholesky factor or not. 
Arithmetic involving `Normal` RVs currently ignores this factor (see e.g. #319), which leads to a loss of information whenever computation relies on Cholesky factors (some do quite heavily, e.g. the square-root Kalman filter). Little by little, I think we should change this. To this end, I suppose we need to do
- [ ] Move `cholesky_update` out of `probnum.statespace` into a more basic module, so it can be used more widely (`probnum.utils`?). At some point, maybe we move it to `marvinpfoertner/cholupdates` (see https://github.com/marvinpfoertner/cholupdates/issues/14) but this is a different issue. 
- [ ] Implement an updated version of **constant-normal arithmetic** (addition and multiplication) that uses Cholesky factors
- [ ] Implement an updated version of **addition of normals** that uses Cholesky factors
- [ ] Implement an updated version of **slicing of multivariate normals**, that uses Cholesky factors
- [ ] Implement an updated version of **reshaping and transposing of normals** that uses Cholesky factors

I'll try to get started with the first one or two points in the near future. I do not know whether I will have time for all of these points, so some help may be needed. 

Also, as soon as it comes to e.g. addition of normals, there are decisions to make (e.g., what should happen if one of the RVs has a Cholesky factor and the other one does not).

PR #337, opened 2021-03-01
Move Cholesky update to a new package `probnum.utils.linalg`

pnkraemer:
First step of #335. 

The idea behind creating the subpackage `probnum.utils.linalg` is that placing functions like `cholesky_update` (which involve a fair amount of computation) right next to functions like `atleast_1d` (which are essentially free) would send the wrong message. 
By forcing a user to import `probnum.utils.linalg` explicitly it might be easier to communicate that this is a different set of functions to what is imported via `probnum.utils` (type checks, etc.).

It can also serve as a place for functions that may be outsourced in the future (see the respective comment in #335)


tests/test_utils/test_linalg/test_cholesky_updates.py
```diff
@@ -0,0 +1,60 @@
...
+    S1 = np.linalg.cholesky(spdmat1)
+    S2 = np.linalg.cholesky(spdmat2)
+    received = utlin.cholesky_update(S1, S2)
+    np.testing.assert_allclose(expected, received)
+
+
+def test_cholesky_optional(spdmat1, even_ndim):
+    """Assert that cholesky_update() transforms a non-square matrix square-root into a
+    correct Cholesky factor."""
+    H = np.random.rand(int(0.5 * even_ndim), even_ndim)
```
marvinpfoertner, line 37: `int(0.5 * even_ndim)` is very error-prone, `even_ndim // 2` is safer

tox.ini
```diff
@@ -71,6 +71,6 @@ commands =
          pylint src/probnum/quad --disable="too-few-public-methods" --jobs=0
          pylint src/probnum/random_variables --disable="missing-function-docstring" --jobs=0
          pylint src/probnum/problems --jobs=0
-         pylint src/probnum/utils --disable="missing-module-docstring,missing-function-docstring" --jobs=0
+         pylint src/probnum/utils --jobs=0
```
marvinpfoertner, line 74: Nice ;)

marvinpfoertner (changes requested):

marvinpfoertner (approved):

PR #338, opened 2021-03-02
Styleguide and developer docs updated

JonathanWenger:
In light of the recent software meeting where some style conventions were discussed, e.g. for imports (`import probnum as pn`). This PR adds them to the styleguide. Also some more clarification on what header structure should be used in tutorials has been added to avoid breaking changes to the documentation sidebar.


STYLEGUIDE.md
```diff
@@ -23,11 +23,20 @@ Use absolute imports over relative imports.
...
 confusion and/or circular imports. This makes imports more readable. When changing the namespace of classes make sure to
 to correct module paths in the documentation by adding `SuperClass.__module__ = "probnum.module"` to the corresponding
 `__init.py__`.
 
+#### Conventions
+If imports are shortened, the following conventions should be used. Full import
+paths are always acceptable.
+
+- `import probnum as pn`
+- `import probnum.random_variables as rvs`
```
pnkraemer, line 38: **Summary:** 
I am not sure about this choice of abbreviation (I am doing it differently), but I am happy that you bring it up this way and I would like to hear your opinion!

**Longer version:**
Thanks for bringing this up! :) I guess you won't be too surprised to see me commenting on this specific change... :) 

I know that `rvs` is how it is done in many ProbNum sub-packages. I also understand that naming is very subjective, nevertheless, I think it is good to discuss it at some point (I remember from the last software meeting that we brought this up, but did not end with a decision; please correct me if I am wrong).

I personally have always used `pnrv`, for the following reasons: it probably makes sense to have consistent abbreviation styles for most sub-packages, not just for random variables. Therefore, if we abbreviate random variables as `rvs`, we should probably lean towards 2-3 letters and only abbreviating the name of the sub-package, without the "probnum" part (maybe something along the lines of `la` for linalg, `de` for diffeq, `fs` for filtsmooth, `sts` for statespace, and so on; details would of course be debatable here). 

Doesn't this potentially bring up the issue of distinguishing probnum-packages from similarly named other packages? One example would be `probnum.type` vs `typing`, where it might be useful to communicate to someone reading the code that `FloatArgType` is from ProbNum, but `Callable` is from `typing`. Another problem would be `probnum.linalg` vs `numpy.linalg` or `scipy.linalg`. I suppose more examples will come up eventually. In any case, if we choose to agree on a default abbreviation (which we don't have to, but which makes sense in my opinion), we should also agree on a default way of distinguishing those potential cases of confusion.

One solution would be to add a `pn` prefix to the ProbNum abbreviations: `pnrv`, `pnde`, `pnfs`, `pnla`, `pnlo`, et cetera, with something like `pntype` or `pntp` for the `type` module. This is what I am currently doing in the code that I write. In my applications, it seemingly solves those namespace confusions, but again, this is very subjective and I understand if you disagree! :) 

I am by no means saying that this is the best solution. Another option would be to import "the other one" with a more precise prefix whenever there is potential confusion (I guess `numpy.linalg` is accessible via `np.linalg` anyway, and `scipy.linalg` could be imported as `scla` or something like this; `typing` could just be imported as `typing`). 

Or, as yet another alternative, at any chance of confusion, we could import the ProbNum sub-package with its full name (I guess `probnum.type` is sufficiently short). Although sufficiently clear, this might send wrong signals to a user, in terms of some packages appearing with its full ProbNum path (`probnum.type`), whereas some other ProbNum packages appear with a different abbreviation (`rvs`).

Please let me know what you think. In any case, this is not something that has to be changed everywhere immediately, but if we set it into stone by mentioning it in the style guide, at least the right objective will be agreed upon :)

JonathanWenger: > **Summary:**
> I am not sure about this choice of abbreviation (I am doing it differently), but I am happy that you bring it up this way and I would like to hear your opinion!
> 
> **Longer version:**
> Thanks for bringing this up! :) I guess you won't be too surprised to see me commenting on this specific change... :)
> 
> I know that `rvs` is how it is done in many ProbNum sub-packages. I also understand that naming is very subjective, nevertheless, I think it is good to discuss it at some point (I remember from the last software meeting that we brought this up, but did not end with a decision; please correct me if I am wrong).

Sorry, I must have misunderstood the conclusion from last week's software meeting. I have removed the import recommendation for now until this has been discussed in more detail. Thanks for giving your detailed opinion. I added my point of view so we can use this post as a basis for a discussion. I think the rest of the PR is non-controversial ;)

> I personally have always used `pnrv`, for the following reasons: it probably makes sense to have consistent abbreviation styles for most sub-packages, not just for random variables. Therefore, if we abbreviate random variables as `rvs`, we should probably lean towards 2-3 letters and only abbreviating the name of the sub-package, without the "probnum" part (maybe something along the lines of `la` for linalg, `de` for diffeq, `fs` for filtsmooth, `sts` for statespace, and so on; details would of course be debatable here).

My understanding from last week's discussion was:
- Users working with ProbNum are free to adopt whatever conventions they want. We don't make recommendations, but in the notebooks we obviously set some kind of precedent.
- Within ProbNum we shorten some imports where convenient (e.g. `probnum.random_variables`).

Now my opinion is, that within ProbNum the scope for `random_variables` is clear, so `rvs` suffices.

> Doesn't this potentially bring up the issue of distinguishing probnum-packages from similarly named other packages? One example would be `probnum.type` vs `typing`, where it might be useful to communicate to someone reading the code that `FloatArgType` is from ProbNum, but `Callable` is from `typing`. Another problem would be `probnum.linalg` vs `numpy.linalg` or `scipy.linalg`. I suppose more examples will come up eventually. In any case, if we choose to agree on a default abbreviation (which we don't have to, but which makes sense in my opinion), we should also agree on a default way of distinguishing those potential cases of confusion.

I think full paths should be used if there is potential for confusion. This is also the recommendation by the [Google styleguide](https://google.github.io/styleguide/pyguide.html).

> One solution would be to add a `pn` prefix to the ProbNum abbreviations: `pnrv`, `pnde`, `pnfs`, `pnla`, `pnlo`, et cetera, with something like `pntype` or `pntp` for the `type` module. This is what I am currently doing in the code that I write. In my applications, it seemingly solves those namespace confusions, but again, this is very subjective and I understand if you disagree! :)
>
> I am by no means saying that this is the best solution. Another option would be to import "the other one" with a more precise prefix whenever there is potential confusion (I guess `numpy.linalg` is accessible via `np.linalg` anyway, and `scipy.linalg` could be imported as `scla` or something like this; `typing` could just be imported as `typing`).
> 
> Or, as yet another alternative, at any chance of confusion, we could import the ProbNum sub-package with its full name (I guess `probnum.type` is sufficiently short). Although sufficiently clear, this might send wrong signals to a user, in terms of some packages appearing with its full ProbNum path (`probnum.type`), whereas some other ProbNum packages appear with a different abbreviation (`rvs`).
> 
> Please let me know what you think. In any case, this is not something that has to be changed everywhere immediately, but if we set it into stone by mentioning it in the style guide, at least the right objective will be agreed upon :)

I put the naming discussion back on the agenda. Once we agree on a style, I'll make a new PR. Let me know if the changes made in this PR are otherwise fine.

PR #342, opened 2021-03-03
Interface for `probsolve_ivp()`

### LINKED ISSUE #272: API of `probsolve_ivp`

pnkraemer:
I guess this resolves #272. 

The API of probsolve_ivp() is made more scipy-esque (by taking f, t0, tmax, y0 as inputs). 
Some other changes to default behaviour are made too, notably adaptive steps as default; higher order prior as default; matern and ioup default hyperparameter choices; method and prior string renamings.

While at it, I also refactored the odefiltsmooth tests in terms of pytest instead of unittest.


src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -40,38 +40,43 @@ class KalmanODESolution(ODESolution):
...
-    >>> initrv = rvs.Constant(0.15)
-    >>> ivp = logistic(timespan=[0., 1.5], initrv=initrv, params=(4, 1))
-    >>> solution = probsolve_ivp(ivp, method="ekf0", step=0.1)
+    >>>
+    >>> def f(t, x):
+    ...     return 4*x*(1-x)
+    >>>
+    >>> y0 = np.array([0.15])
+    >>> t0, tmax = 0., 1.5
+    >>> solution = probsolve_ivp(f, t0, tmax, y0, step=0.1, atol=None, rtol=None)
```
pnkraemer, line 49: this is recovering the previous doctest (atol and rtol at None implies fixed steps). 
I am not sure whether we might want an adaptive step example, but what is the doctest here (the outputs might change marginally depending on some default tolerances which I guess are subject to change in the next weeks)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -21,22 +21,29 @@
...
+    method="EK0",
+    dense_output=True,
+    which_prior="IBM2",
+    atol=1e-2,
+    rtol=1e-2,
     step=None,
-    firststep=None,
-    **kwargs
+    driftspeed=None,
+    lengthscale=None,
```
pnkraemer, line 44: no more kwargs :) (of you're wondering why None as a default instead of, like, 1.0, see below)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -218,91 +254,102 @@ def probsolve_ivp(
...
+        # lazy version of Hairer, Wanner, Norsett, p. 169
+        norm_y0 = np.linalg.norm(ivp.initrv.mean)
+        norm_dy0 = np.linalg.norm(ivp(ivp.t0, ivp.initrv.mean))
+        firststep = 0.01 * norm_y0 / norm_dy0
+    else:
+        firststep = step
+
+    # Set up some default hyperparameters: not great, but much better than 1.0
+    lengthscale = lengthscale if lengthscale is not None else 1.0 / firststep
+    driftspeed = driftspeed if driftspeed is not None else firststep
```
pnkraemer, line 278: turns out that these are much better than choosing 1.0 (see the adaptive stepsize notebook). Of course they are not perfect, but it works surprisingly well.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -218,91 +254,102 @@ def probsolve_ivp(
...
     """
-    stprl = _create_steprule(atol, rtol, step, firststep, ivp)
-    prior = _string2prior(ivp, which_prior, **kwargs)
-    gfilt = _create_filter(ivp, prior, method, **kwargs)
-    with_smoothing = method[-2] == "s" or method[-1] == "s"
-    solver = GaussianIVPFilter(ivp, gfilt, with_smoothing=with_smoothing)
+
+    # Normalize string inputs
+    method = method.upper()
+    which_prior = which_prior.upper()
```
pnkraemer, line 262: I suppose this is controversial

src/probnum/diffeq/steprule.py
```diff
@@ -76,7 +76,7 @@ def __init__(
         firststep,
         atol,
         rtol,
-        limitchange=(0.1, 5.0),
+        limitchange=(0.2, 5.0),
```
pnkraemer, line 79: this asymmetry kind of bothered me (and the outputs seem not worse)... I suppose @NinaEffenberger will find that hilarious ;)

tests/test_diffeq/test_odefiltsmooth/test_ivpfiltsmooth.py
```diff
@@ -0,0 +1,109 @@
+"""Test the ODE filter on a more granular level.
+
+The tests in here go through probsolve_ivp(), but do not directly test
+the interface, but the implementation. Therefore this test module is
+named w.r.t. ivpfiltsmooth.py.
+"""
```
pnkraemer, line 6: the tests "the interface works" are separated from the tests "the numbers are good" now. both are written in pytest.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -218,196 +254,187 @@ def probsolve_ivp(
...
-        return _string2ibm(ivp, which_prior, **kwargs)
+        return _string2ibm(ivp, which_prior)
     elif which_prior in ioup_family:
-        return _string2ioup(ivp, which_prior, **kwargs)
+        return _string2ioup(ivp, which_prior, driftspeed=driftspeed)
     elif which_prior in matern_family:
-        return _string2matern(ivp, which_prior, **kwargs)
-    else:
-        raise RuntimeError("It should have been impossible to reach this point.")
+        return _string2matern(ivp, which_prior, lengthscale=lengthscale)
```
pnkraemer, line 324: this could in principle be replaced with `if which_prior[:3] == "IBM": ...`

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -218,196 +254,187 @@ def probsolve_ivp(
...
             backward_implementation="sqrt",
         )
-    elif which_prior == "ibm4":
+    else:  # which_prior == "IBM4":
         return pnfs.statespace.IBM(
             4,
             ivp.dimension,
             forward_implementation="sqrt",
             backward_implementation="sqrt",
         )
```
pnkraemer, line 357: this could in principle be replaced with  `ordint = which_prior[-1]` which would remove all the ifs. This goes together with the comment above I guess. Thoughts?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
pnkraemer: the same obviously holds for the other priors (which is also true for the comment below)

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
nathanaelbosch: Just to question it: Is there an alternative syntax for choosing fixed steps? It seems inconvenient to set two values to `None`, maybe just checking `step is not None` is enough? Or something more explicit, like `adaptive=false`?

nathanaelbosch, line 49: Also, seeing this example I wonder if I prefer the syntax now or before... I liked our own problem types^^ extracting all the content makes them a bit less useful. But I guess it's a trade-off with being more like scipy.

src/probnum/diffeq/steprule.py
nathanaelbosch: ...ok just so I understand, the only reason for this change is aesthetics? :joy: I think there is an argument for the previous setting though, like allowing the solver to slow down more quickly when encountering a steep slope or something, but then I also saw defaults of (0.2,10.0) in the Julia repos so not sure if that really counts.

Long story short, from my experience I wouldn't expect this to matter a lot so I'm 100% with this change :+1:

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -105,39 +123,48 @@ def probsolve_ivp(
...
+        require Jacobians of the RHS-vector field of the IVP.
+        That is, the argument ``df`` needs to be specified.
+        They are likely to perform better than zeroth order methods in
+        terms of (A-)stability and "meaningful uncertainty estimates".
+        The unscented Kalman filter is supported,
+        but since its square-root implementation is not available yet, it will be
+        less stable (numerically, i.e. in terms of singular matrices)
+        than the extended Kalman filter variations.
+        While we recommend to use correct capitalization for the method string,
+        lower-case letters will be capitalized internally.
```
nathanaelbosch, line 149: Nice write-up! :)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -177,32 +204,41 @@ def probsolve_ivp(
...
+    >>>
+    >>> y0 = np.array([0.15])
+    >>> t0, tmax = 0., 1.5
+    >>> solution = probsolve_ivp(f, t0, tmax, y0, step=0.1, atol=None, rtol=None)
     >>> print(np.round(solution.y.mean, 2))
     [[0.15]
      [0.21]
      [0.28]
-     [0.36]
+     [0.37]
```
nathanaelbosch, line 220: These changes don't cease to surprise me

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
nathanaelbosch: I don't really see where this could lead to errors honestly, so I think it should be fine

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -218,196 +254,187 @@ def probsolve_ivp(
...
-        raise ValueError(errormsg)
-    atol_not_rtol = atol is not None and rtol is None
-    rtol_not_atol = rtol is not None and atol is None
-    if atol_not_rtol or rtol_not_atol:
-        errormsg = "Please specify either both atol and rtol, or neither."
-        raise ValueError(errormsg)
-
-
-def _string2prior(ivp, which_prior, **kwargs):
+def _string2prior(ivp, which_prior, driftspeed, lengthscale):
```
nathanaelbosch, line 314: Alright, so I tried something:
```python
import re
which_prior = "IOUP14"
m = re.match("^(IBM|IOUP|MAT)([0-9]+)$", which_prior)
if m is None:
    raise ValueError("This prior is not supported.")
prior_str, order_str = m.groups()
```
after which we have `prior_str == "IOUP"` and `order_str == "14"`. So all that's left would be an if-else / dict matching over the prior, and some special care for the matern order.

What do you think?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
nathanaelbosch: I'm not a fan of the fact that we match the order here, even though we just want to know the prior. On the other hand, matching the full string is cleaner than matching substrings, so idk....

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
nathanaelbosch: I would strongly prefer extracting the order from the string and then using that, to remove all the duplicate code here. We might need to take some special care to ensure, that people don't supply values >= 10, but if we keep the explicit string matching that we currently have in `_string2prior` then that's not a risk.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -218,196 +254,187 @@ def probsolve_ivp(
...
             lengthscale,
             forward_implementation="sqrt",
             backward_implementation="sqrt",
         )
-    else:
-        raise RuntimeError("It should have been impossible to reach this point.")
 
 
-def _string2filter(_ivp, _prior, _method, **kwargs):
+def _string2filter(_ivp, _prior, _method):
```
nathanaelbosch, line 432: The separation between this function and `_create_filter` is a bit blurry, so I would suggest merging them.

tests/test_diffeq/test_odefiltsmooth/test_ivpfiltsmooth.py
nathanaelbosch: That sounds like a very useful separation :)

tests/test_diffeq/test_odefiltsmooth/test_ivpfiltsmooth.py
```diff
@@ -0,0 +1,109 @@
...
+        method="ek0",
+        which_prior="ibm1",
+        atol=None,
+        rtol=None,
+        step=step,
+    )
+
+
+def test_first_iteration(ivp, sol):
+    """Test whether first few means and covariances coincide with Prop.
```
nathanaelbosch, line 45: Weird docstring split^^

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -1,507 +1,105 @@
...
+                f,
+                t0,
+                tmax,
+                y0,
+                which_prior=which_prior,
+            )
+
+
+def test_wrong_filter_raises_error(ivp):
+    """Priors that are not in the list raise errors."""
```
nathanaelbosch, line 87: priors->filters

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -1,507 +1,105 @@
...
+    t0, tmax = ivp.timespan
+    y0 = ivp.initrv.mean
+
+    # UK1 does not exist
+    with pytest.raises(ValueError):
+        probsolve_ivp(f, t0, tmax, y0, method="UK1")
+
+
+def test_no_step_or_tol_info_raises_error(ivp):
+    """Priors that are not in the list raise errors."""
```
nathanaelbosch, line 98: wrong docstring^^

nathanaelbosch (changes requested):
Thanks for the PR! As announced, I had a few remarks on the string handling, sorry again for that. I also wondered about the change ((ivp)->(f,t0,tmax,y0)) in general. I have to admit that I slightly prefer the ivp variant, but I also understand the reasons for scipy compatibility, so that's more about just stating my opinion than requesting any changes ;)

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
pnkraemer: Let's talk about this later today.

pnkraemer: Yes, I suppose it is related to the convenience function discussion we had yesterday...

src/probnum/diffeq/steprule.py
pnkraemer: What is the default in `ProbNumDiffeq.jl`?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
pnkraemer: Nice, thanks! Tbh, this looks better than I expected^^ I played around with it, and like its robustness!

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -1,507 +1,108 @@
...
+@pytest.mark.parametrize("method", ["EK0", "EK1"])
+@pytest.mark.parametrize(
+    "which_prior",
+    [
+        "IBM1",
+        "IBM2",
+        "IBM3",
+        "IBM5",
+        "IBM8",
+        "IBM11",
```
pnkraemer, line 24: we can get up to 11 already, even with bad initialisation :D 

IOUP and Matern seem capped at 4, though. No idea why.

pnkraemer: (this test only checks whether probsolve_ivp returns something useful for atol=rtol=1e-1. This is not a tough test, but I am happy about order 11 anyway. )

tests/test_diffeq/test_odefiltsmooth/test_odefiltsmooth.py
```diff
@@ -1,507 +1,108 @@
...
+
+def test_wrong_prior_raises_error(ivp):
+    """Priors that are not in the list raise errors."""
+    f = ivp.rhs
+    t0, tmax = ivp.timespan
+    y0 = ivp.initrv.mean
+
+    # Anything that is no {IBM, IOUP, MAT} + Number is wrong
+    # (the Matern number must end on a 2).
+    for which_prior in ["IBM_5", "IOUPX5", "MAT112Y", "MAT33"]:
```
pnkraemer, line 79: any other corner cases that may be checked here? These seemed like the most obvious ones (all of which are naturally captured by the RE -- thanks!)

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
pnkraemer: Decision: include `adaptive=True/False`

src/probnum/diffeq/steprule.py
pnkraemer: Decision: use (0.2, 10.0) (but without any strong reasoning, other than that it seems to be common values)

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
pnkraemer: Decision: keep the "unpacked" interface; think about potential tuple unpacking for the problem classes.

pnkraemer:
Decisions from earlier:
* remove everything from ``probsolve_ivp`` that we wouldn't recommend. That includes Matern & IOUP priors, and the UKF
* Let those parts of documentation (i.e. docstrings and notebooks) that call `probsolve_ivp` use "handcrafted" ODE problems, not the diffeq zoo
* Pass the order of the prior as an int. Include a flag "adaptive".

src/probnum/diffeq/__init__.py
```diff
@@ -49,6 +49,7 @@
     "ODESolution",
     "KalmanODESolution",
     "compute_all_derivatives",
+    "propose_firststep",
```
pnkraemer, line 52: I suppose this comes in handy when defining an ODE solver from scratch.

pnkraemer:
I've removed IOUP, matern, and UKF from the interface function. the notebooks use a handcrafted ODE problem now. probsolve_ivp now uses an `adaptive` flag (which I think makes it much clearer what is going on, thanks for the suggestion!)


Now, there are a few points that more or less overlap with #334. I suggest that we merge #334 first, and then this one here (in #334 I made some remarks "will take care of this in the probsolve ivp interface PR"). 

Therefore I will wait with requesting your review @nathanaelbosch until the other PR is merged :) Thanks!

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -103,48 +103,27 @@ def string_to_measurement_model(
...
-            "UKS": pnfs.DiscreteUKFComponent.from_ode(
-                ivp,
-                prior,
-                evlvar=measurement_noise_covariance,
-            ),
+            # "UK": pnfs.DiscreteUKFComponent.from_ode(
+            #     ivp,
+            #     prior,
+            #     evlvar=measurement_noise_covariance,
+            # ),
```
pnkraemer, line 126: do we want the UKF in this convenience function or not? Pro: it is so easy to do, contra: if we dont include it here, we can refer value checks from probsolve_ivp to this function (and it would be more in line with us telling people: dont use the ukf?)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -176,32 +169,41 @@ def probsolve_ivp(
...
      [0.74]
      [0.81]
-     [0.86]
-     [0.9 ]
-     [0.93]
-     [0.95]
+     [0.87]
+     [0.91]
+     [0.94]
+     [0.96]
```
pnkraemer, line 194: these values changed, because now the initialisation strategy probsolve ivp  uses RK initialization

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -67,76 +65,71 @@ def probsolve_ivp(
...
+        Absolute tolerance  of the adaptive step-size selection scheme.
+        Optional. Default is ``1e-4``.
+    rtol : float
+        Relative tolerance   of the adaptive step-size selection scheme.
+        Optional. Default is ``1e-4``.
+    step :
+        Step size. If atol and rtol are not specified, this step-size is used for a fixed-step ODE solver.
+        If they are specified, this only affects the first step. Optional.
+        Default is None, in which case the first step is chosen as :math:`0.01 \cdot |y_0|/|f(t_0, y_0)|`.
+    algo_order
```
nathanaelbosch, line 97: What about `alg_order`, or even just `order`? I would slightly prefer those two, but feel free to pick whichever you prefer.

nathanaelbosch (approved):
Ok I just had a mini-comment about the naming of `algo_order` (I proposed to use `alg_order` or `order`). But I'm fine with whichever you want to keep, so I'll just approve it already and you can decide to change things or just merge :)

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
pnkraemer: Thanks for the suggestion! If you dont mind, I'd prefer to keep it that way. Even though my opinion is not strong by any means, here are my thoughts:
* `order` crops up way to often already in my mind. I like that `alg(o)_order` is much clearer in this context, even for people that have no idea about ODE filters
* I slightly prefer "algo" over "alg" (and, in order to pretend that I have a real reference for why this is better: https://stackoverflow.com/questions/5372129/short-name-for-algorithm, though this "reference" should be taken with a grain of salt :D)

PR #344, opened 2021-03-10
Introduce base measure sampling in TimeSeriesPosterior and derived classes

pnkraemer:
This PR essentially
* streamlines `FiltSmoothPosterior` and `ODESolutions` under the joint interface `TimeSeriesPosterior` (formerly `FiltSmoothPosterior`) 
* introduces methods `transform_base_measure_realizations...()` in all posteriors as well as in `Transition`s. Now, the randomness in all sampling procedures lives in sample() (which is not new -- now, however, it is 1 out of 2 lines in the sample() functions, the other line delegates to `transforme_base_measure_realizations` which is where the magic happens. 
* Has default call behaviour in all posteriors. Adds the interface "interpolate()" which computes values between gridpoints.
* Updates imports in filtsmooth and diffeq
* Adds types in the modules that were touched anyway (all posterior-ones, essentially).

As a side result of this refactoring, I think I was able to simplify the logic behind dense output and sampling a bit (in terms of sorting where (a) the randomness happens, (b) the forward/backward passes happen, and all the remainders which just point to these spots).

I am aware that -- unfortunately -- the logic in the code has become quite complex. I tried my best to (i) test thoroughly, (ii) name well, and (iii) block-comment-explain lines which I thought might create confusion of sorts. Please let me know if things are not clear!


src/probnum/statespace/transition.py
```diff
@@ -286,37 +288,54 @@ def smooth_list(self, rv_list, locations, _previous_posterior=None):
...
+    #         List of smoothed random variables.
+    #     """
+    #
+    #     base_measure_samples = np.random.randn(len(t), self.input_dim)
+    #     return self.jointly_transform_base_measure_realization_list_backward(
+    #         rv_list,
+    #         t,
+    #         base_measure_samples=base_measure_samples,
+    #         _previous_posterior=_previous_posterior,
+    #     )
```
pnkraemer, line 323: In the spirit of the randomness in sampling schemes happening in `Posterior.sample()` and the "heavy lifting" being implemented here for convenience's sake, I'd propose to remove this function. The rules of how to transform the samples are captured below, and there is IMO no need for this function to be called here. Thoughts?

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -121,18 +108,54 @@ def sample(
         self,
         t: typing.Optional[float] = None,
         size: typing.Optional[probnum.type.ShapeArgType] = (),
+        random_state=None,
     ) -> np.ndarray:
         """Sample from the Gaussian filtering ODE solution by sampling from the Gauss-
         Markov posterior."""
```
pnkraemer, line 114: wrong docs?

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -121,18 +108,54 @@ def sample(
...
         size = probnum.utils.as_shape(size)
+        t = np.asarray(t) if t is not None else None
+
+        if t is None:
+            t_shape = (len(self.locations),)
+        else:
+            t_shape = (len(t) + 1,)
```
pnkraemer, line 121: explain

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -121,18 +108,54 @@ def sample(
...
+        )
+
+    def transform_base_measure_realizations(
+        self, base_measure_realizations, t=None, size=()
+    ):
+        size = probnum.utils.as_shape(size)
 
-        # implement only single samples, rest via recursion
+        # Implement only single samples, rest via recursion
         # We cannot 'steal' the recursion from self.kalman_posterior.sample,
```
pnkraemer, line 142: kalmanposterior.sample() or kalmanposterior.transform_base_measure_realizations()?

src/probnum/diffeq/odesolution.py
```diff
@@ -4,41 +4,23 @@
...
-    # functools.cached_property is only available in Python >=3.8
-    from functools import cached_property
-except ImportError:
-    from cached_property import cached_property
+from probnum import filtsmooth
 
 
-class ODESolution(abc.ABC):
+class ODESolution(filtsmooth.TimeSeriesPosterior):
     """ODE Solution interface."""
```
pnkraemer, line 15: this could actually be initialised, right? So it is not really an interface?

src/probnum/diffeq/odesolution.py
```diff
@@ -84,3 +54,8 @@ def sample(
             Number of samples.
         """
         raise NotImplementedError("Sampling is not implemented.")
+
+    def transform_base_measure_realizations(
+        self, base_measure_realizations, t=None, size=()
+    ):
+        raise NotImplementedError("Sampling not implemented.")
```
pnkraemer, line 61: update

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -94,8 +78,53 @@ def interpolate(self, t):
...
         raise NotImplementedError
 
+    def sample(self, t=None, size=(), random_state=None):
+
+        size = utils.as_shape(size)
+
+        if t is None:
+            t_shape = (len(self.locations),)
+        else:
+            t_shape = (len(t) + 1,)
```
pnkraemer, line 88: explain?

tests/test_diffeq/test_odesolver.py
```diff
@@ -37,7 +29,7 @@ def step(self, start, stop, current):
         )  # return nan as error estimate to ensure that it is not used
 
     def rvlist_to_odesol(self, times, rvs):
-        return MockODESolution(times, rvs)
+        return MockODESolution(locations=times, states=rvs)
```
pnkraemer, line 32: I think this could be replaced by `ODESolution` itself

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -79,66 +71,93 @@ class KalmanODESolution(ODESolution):
...
+        )
+        return self.transform_base_measure_realizations(
+            base_measure_realizations=base_measure_realizations, t=t, size=size
+        )
+
+    def transform_base_measure_realizations(
+        self,
+        base_measure_realizations: np.ndarray,
+        t: Optional[DenseOutputLocationArgType] = None,
+        size: Optional[ShapeArgType] = (),
```
pnkraemer, line 132: it should be possible to remove this argument altogether. Size should be possible to infer from the shape of base_measure_realizations?!

src/probnum/statespace/transition.py
pnkraemer: If you agree, please let me know, and I delete this block before merging

pnkraemer:
See also the decisions from the software meeting (#349 and #348) which are more or less resolved for these parts.

src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -0,0 +1,190 @@
...
+
+    Parameters
+    ----------
+    locations :
+        Locations of the posterior states (represented as random variables).
+    states :
+        Posterior random variables.
+    """
+
+    def __init__(self, locations: np.ndarray, states: np.ndarray) -> None:
```
nathanaelbosch, line 39: Wasn't there a specific array-like type that we always use? I don't completely recall how it was called. But I'm also the least experienced person for the type hints, so np.ndarray might also be fine^^

src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -0,0 +1,190 @@
...
+        t :
+            Location, or time, at which to evaluate the posterior.
+
+        Returns
+        -------
+        random_variables.RandomVariable or _randomvariablelist._RandomVariableList
+            Estimate of the states at time ``t``.
+        """
+        #
+        # # Early exit if t consists of gridpoints:
```
nathanaelbosch, line 76: These comments should probably be removed?

src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -0,0 +1,190 @@
...
+
+DenseOutputValueType = Union[
+    random_variables.RandomVariable, _randomvariablelist._RandomVariableList
+]
+"""Dense evaluation of a TimeSeriesPosterior returns a RandomVariable if evaluated at a single location,
+and a _RandomVariableList if evaluated at an array of locations."""
+
+
+class TimeSeriesPosterior(abc.ABC):
+    """Posterior Distribution over States after time-series algorithms such as
```
nathanaelbosch, line 28: Could you briefly remind me why we need TimeSeriesPosterior, and why just having KalmanPosterior is not sufficient?

nathanaelbosch (changes requested):

src/probnum/_randomvariablelist.py
```diff
@@ -53,6 +53,11 @@ def std(self) -> np.ndarray:
             return np.array([])
         return np.stack([rv.std for rv in self])
 
+    @property
+    def shape(self):
+        first_rv = np.asarray(self[0].mean)
+        return (len(self),) + first_rv.shape
```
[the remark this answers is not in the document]
pnkraemer: Magic, right? ;)

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -79,66 +71,109 @@ class KalmanODESolution(ODESolution):
...
+            remove_final_point = True
+
+        # Infer desired size of the base measure realizations and create them
+        size = utils.as_shape(size)
+        single_rv_shape = self.kalman_posterior.states[0].shape
+        base_measure_realizations = stats.norm.rvs(
+            size=(size + sampling_locs.shape + single_rv_shape),
+            random_state=random_state,
+        )
+
```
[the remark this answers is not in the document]
pnkraemer: That is a good point. The problem is here how to slice the correct dimensions in a way that works with all kinds of samplign shapes (which I wasnt able to do, so I mirrored the KalmanPosterior logic in KalmanODESolution). This could be done with np.einsum I suppose, let me try some things and get back to rou.

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -4,105 +4,164 @@
...
-        t : float
-            Location, or time, at which to evaluate the posterior.
+        t :
+            Location to evaluate at.
 
         Returns
         -------
-        :obj:`RandomVariable`
-            Estimate of the states at time ``t``.
+        random_variables.RandomVariable or _randomvariablelist._RandomVariableList
```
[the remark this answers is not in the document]
pnkraemer: No, I dont think so. Seems like someone mindlessly copied the doscstring here...

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -132,68 +199,85 @@ def interpolate(self, t):
...
-    ----------
-    locations : `array_like`
-        Locations / Times of the discrete-time estimates.
-    state_rvs : :obj:`list` of :obj:`RandomVariable`
-        Estimated states (in the state-space model view) of the discrete-time estimates.
-    transition : :obj:`Transition`
-        Dynamics model used as a prior for the filter.
-    """
+    def interpolate(self, t: DenseOutputLocationArgType) -> DenseOutputValueType:
+        """Predict to the present point.
```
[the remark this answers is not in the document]
pnkraemer: In principle, yes. The details are slightly different (i.e. the FilterinPosterior merely predicts, and the smoothing posterior updates as well), though. What would you prefer?

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -132,68 +199,85 @@ def interpolate(self, t):
...
         return rv
 
-    def sample(self, locations=None, size=()):
+    def sample(
+        self,
+        t: Optional[DenseOutputLocationArgType] = None,
+        size: Optional[ShapeArgType] = (),
+        random_state: Optional[RandomStateArgType] = None,
+    ) -> np.ndarray:
         raise NotImplementedError
```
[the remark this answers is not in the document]
pnkraemer: I dont think this is the full point. The superclass implements `sample()` by delegationg to `transform_base_measure_...` and when trying to sample from the fitlering posterior, this is where the error would be thrown. I thought it would be a bit more human-readable if the error was thrown right at sample(), which is why it is overwritten here.
However, now that we're talking about this, it should have an error message. 

What are your thoughts?

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: TimeSeriesPosterior is intended to be a joint interface for all sorts of posterior processes. This includes `KalmanPosterior`, but also `ODESolution` (, and `ParticleFilterPosterior`).

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: There is numpy.typing.ArrayLike (which I think is what you're referring to), but that type is not compatible with the current numpy version we are using.

Even if we would use the ArrayLike type, I think it would only make sense a line such as `np.asarray(states)` in the first lines (which we may or may not want -- I dont know). If we specify that we want an array, this should be clear either way.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -79,66 +71,109 @@ class KalmanODESolution(ODESolution):
...
-            return np.array([self.sample(t=t, size=size[1:]) for _ in range(size[0])])
+        # Include the final point if a specific grid is demanded
+        # and the rightmost point is left of the rightmost data point.
+        # If this is not done, the samples are not from the full posterior.
+        if t is None:
+            sampling_locs = self.locations
+            remove_final_point = False
+        elif t[-1] >= self.locations[-1]:
+            sampling_locs = t
+            remove_final_point = False
```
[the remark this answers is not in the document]
pnkraemer: If we delegate this entire logic here to `KalmanPosterior`, it will becovered (because this branch is tested for KalmanPosterior, but apparently not for ODE solution). If not, I will write a test for this (should be easy).

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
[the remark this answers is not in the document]
pnkraemer: Ignore the previous message. Here is the proper answer:

This is indeed a potentially controversial point.
We have to decide between (a) letting `KalmanODESolution` have a `transform_base_measure_realizations` which `sample()` points to, and (b) letting `KalmanODESolution.sample()` point to `KalmanPosterior.sample()` (which itself points to a `KalmanPosterior.transform_base_measure...`. 

In the spirit that `sample()` is something that "sits on top of `transform_base...`", I opted for (a). In other words, `KalmanODESolution.sample()` delegates the implementation to `KalmanPosterior` via the `transform_base_measure_realizations()` route. 

On a side note, you're right in that `KalmanODESolution.transform_base_measure_realizations` could probably do without the recursion itself. Let me see what I can do in this regard.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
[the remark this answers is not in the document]
pnkraemer: Done.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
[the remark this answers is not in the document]
pnkraemer: > On a side note, you're right in that KalmanODESolution.transform_base_measure_realizations could probably do without the recursion itself. Let me see what I can do in this regard.

This is taken care of.

pnkraemer:
If you'd like to see how this functionaliy is used, check out the notebook(s) in the docs of https://github.com/probabilistic-numerics/probnum-evaluation/pull/7

src/probnum/filtsmooth/timeseriesposterior.py
nathanaelbosch: I see, thanks!

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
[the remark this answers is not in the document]
nathanaelbosch: My bad, if there are differences then of course we should have it.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
```diff
@@ -79,66 +70,61 @@ class KalmanODESolution(ODESolution):
...
-        # because we need to project the respective states out of each sample.
-        if size != ():
-            return np.array([self.sample(t=t, size=size[1:]) for _ in range(size[0])])
+        samples = self.kalman_posterior.sample(
+            t=t, size=size, random_state=random_state
+        )
+
+        # Project the samples down to the "true" KalmanODESolution dimensions
+        # (which are a subset of the KalmanPosterior dimensions)
+        ode_samples = np.einsum("dq,...q->...d", self.proj_to_y, samples)
```
nathanaelbosch, line 108: Could you elaborate what happens here? I know about einsums, but I don't understand why we use it here (or why it even works). Would `[_project_rv(self.proj_to_y, s) for s in samples]` do the same?

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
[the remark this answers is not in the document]
nathanaelbosch: Ah I see, makes sense. Yes I agree, an error message would be useful

src/probnum/statespace/transition.py
nathanaelbosch: I'm not sure if this is still up to date, but in case it is I still agree :)

nathanaelbosch (approved):
I finally managed to go through again. Apart from a single error message to be added, I think everything should be in order! Or do you remember any other open things, from when we talked last week? I'll approve already, but if yes let me know and I can have another look.

src/probnum/diffeq/odefiltsmooth/kalman_odesolution.py
pnkraemer: yes, it would do the same. Einsum is faster though, because it is vectorised :)

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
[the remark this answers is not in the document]
pnkraemer: Done

pnkraemer:
Error message is added. I dont think there was anything else :) Thanks for going through again! 

On a side note: there is another tiny change in `Preconditioner`: the preconditioner is now assembled with |h| instead of h, which is something that raised weird errors for me in a different project when using ProbNum. I thought we'd get it out of the way here, I hope you dont mind :)

Issue #186, opened 2020-08-27
KalmanPosterior (and therefore ODESolution) doesnt consider sigma-squared


nathanaelbosch:
I don't think we explicitly use it anywhere. Therefore, I would imagine that the used `Q` does not actually use the estimated sigma squared, and therefore the prediction does not actually correspond to the true posterior prediciton. 

@pnkraemer Is this the case, or did I miss something? If this is indeed currently a bug we should discuss how to best fix that.

pnkraemer:
Good question. I think this is indeed a bug but should be fixable by replacing
```python
    def postprocess(self, times, rvs):
        """Rescale covariances with sigma square estimate and (if specified), smooth the estimate."""
        rvs = [
            RandomVariable(
                distribution=Normal(rv.mean(), self.sigma_squared_global * rv.cov())
            )
            for rv in rvs
        ]
        odesol = super().postprocess(times, rvs)
        if self.with_smoothing is True:
            odesol = self._odesmooth(ode_solution=odesol)
        return odesol
```
with 
```python
    def postprocess(self, times, rvs):
        """Rescale covariances with sigma square estimate and (if specified), smooth the estimate."""
        rvs = [
            RandomVariable(
                distribution=Normal(rv.mean(), self.sigma_squared_global * rv.cov())
            )
            for rv in rvs
        ]
        self.prior.diffconst *= self.sigma_squared_global    # changed here!
        odesol = super().postprocess(times, rvs)
        if self.with_smoothing is True:
            odesol = self._odesmooth(ode_solution=odesol)
        return odesol
```
right?

pnkraemer:
Now that I am thinking about it, it might be good to make the fix slightly more complex than this single line. 
There should be a test for the sigma change (take a test problem with an overly small prior sigma and assure that the sigma of the prior is larger after the solve, perhaps checking that the returned final RV is a (positively) scaled version of the last intermediate random variable). Is that too cryptic or do you understand what I mean?

nathanaelbosch:
The fix you proposed seems correct to me, and might be a good solution for the current code. But I agree that we might want to start by designing a test that specifically targets this part of the "correctness" of the kalman posterior. I didn't 100% understand what you meant though.

Also, for time-varying sigmas this fix would not be enough. That's not really relevant right now of course, and I'm not really sure how to best implement them with the current codebase anyways, but I hope that I have some success with them and that I'll therefore want to add them at some point.

pnkraemer:
[...]

PR #352, opened 2021-03-12
Calibration for ODE solvers

pnkraemer:
This PR introduces a choice between constant and dynamic calibration in the probabilistic ODE solver.

Edit: everything below this sentence is out of date. Sigma-squared in the posterior needs to be resolved in order to yield stable smoothing and dense output. 

(

While this provides improved flexibility, it does not resolve issues such as #186. I think resolving these is a different PR (and doing it here would raise some conflicts with #344), and for now I am happy with having a basic choice between calibration styles. 

Question: do you think this setup  generalises more or less well to the other calibration styles you researched on @nathanaelbosch? 

)


src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+
+from probnum import random_variables
+from probnum.type import FloatArgType
+
+DiffusionType = Union[FloatArgType]
+"""Acceptable diffusion types are -- in principle -- floats and arrays of shape (d,).
+In other words, everything that behaves well with `cov1 + diffusion*cov2` and
+`chol1 + sqrt(diffusion) * chol2`.
+For now, let's only use floats, because it is not clear to me how to best implement the square-root behaviour.
+"""
```
pnkraemer, line 21: I think it is useful to define such a type, because we can communicate the similarity between the output types of Diffusion and the inputs of transitions.

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class Diffusion(abc.ABC):
+    r"""Interface for diffusion models :math:`\sigma: \mathbb{R} \rightarrow \mathbb{R}^d` and their calibration."""
+
+    def __repr__(self):
+        raise NotImplementedError
```
pnkraemer, line 28: this has already come in useful for me while debugging :D

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @abc.abstractmethod
+    def __call__(self, t) -> DiffusionType:
+        """Evaluate the diffusion."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def __getitem__(self, idx):
+        """Get the i-th diffusion from the list."""
+        pass
```
pnkraemer, line 38: these are not used yet (though implemented for constant diffusions already, but I think they will be useful when trying to resolve issues such as #186

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def calibrate_locally(self, meas_rv):
```
pnkraemer, line 40: I think it is best to keep this function name rather high level and not talk about quasi-MLEs here -- who knows where we are heading.

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+            b = 1 - a
+            self.diffusion = a * diffusion + b * self.diffusion
+
+    def postprocess_states(self, states, locations):
+        return [
+            random_variables.Normal(rv.mean, self.diffusion * rv.cov) for rv in states
+        ]
+
+
+class PiecewiseConstantDiffusion(Diffusion):
```
pnkraemer, line 98: I opted for naming this `PiecewiseConstantDiffusion` over `DynamicDiffusion` because it resembles the mathematical assumptions more. Nevertheless, I think it remains convenient to use the terms "constant" and "dynamic" in e.g. `probsolve_ivp`.

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+
+        Used for uncertainty calibration in the ODE solver.
+        """
+        std_like = meas_rv.cov_cholesky
+        whitened_res = np.linalg.solve(std_like, meas_rv.mean)
+        ssq = whitened_res @ whitened_res / meas_rv.size
+        return ssq
+
+    @abc.abstractmethod
+    def update_current_information(self, diffusion, t):
```
pnkraemer, line 51: If you dont like the name, please let me know. I think the difficulties here are to show that (1) something happens **in-place** and (2) it can either be the update of an average value or an append.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -60,6 +61,8 @@ def __init__(
             Normal,
         ],
         initrv: typing.Optional[Normal] = None,
+        diffusion=None,
```
pnkraemer, line 64: the diffusion class is passed as a class. Strings are only communicated inside `probsolve_ivp`. I suppose we will add more diffusion soon, and bookkeeping the strings is a bit tedious...

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -60,6 +61,8 @@ def __init__(
...
+        re_predict_with_calibrated_diffusion=True,
```
pnkraemer, line 65: this flag allows switching between the dynamic and pseudo-dynamic diffusions that we talked about before

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -251,6 +222,8 @@ def step(self, t, t_new, current_rv):
     def rvlist_to_odesol(self, times, rvs):
         """Create an ODESolution object."""
 
+        rvs = self.diffusion.postprocess_states(rvs, times)
```
pnkraemer, line 225: this happens in this function, not in postprocess, because it needs access to only the RV list. I found this more convenient this way.

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
```diff
@@ -122,6 +127,15 @@ def probsolve_ivp(
         Whether we want dense output. Optional. Default is ``True``. For the ODE filter,
         dense output requires smoothing, so if ``dense_output`` is False, no smoothing is performed;
         but when it is ``True``, the filter solution is smoothed.
+    diffusion_model : str
+        Which diffusion model to use. The choices are ``'constant'`` and ``'dynamic'``,
+        which implement different styles of
+        online calibration of the underlying diffusion [5]_.
+        Optional. Default is ``'dynamic'``.
+        While we recommend to use correct capitalization for the `diffusion_model` string,
+        upper-case letters will be lower-cased internally.
```
pnkraemer, line 136: sufficiently clear?

src/probnum/diffeq/odefiltsmooth/diffusions.py
nathanaelbosch: I'm a bit conflicted about having both `__call__` and `__getitem__`, since the first suggests very general functions as diffusions while the second only makes sense with our choice of step functions. What about only having `__call__`?

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+        pass
+
+    @abc.abstractmethod
+    def postprocess_states(self, states, locations):
+        """Postprocess a set of ODE solver states after seeing all the data."""
+        pass
+
+
+class ConstantDiffusion(Diffusion):
+    """Piecewise constant diffusion and their calibration."""
```
nathanaelbosch, line 66: Wrong docstring

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+    """Piecewise constant diffusion."""
+
+    def __init__(self):
+        self.diffusions = []
+        self.times = []
+
+    def __repr__(self):
+        return f"PiecewiseConstantDiffusion({self.diffusions})"
+
+    def __call__(self, t) -> DiffusionType:
```
nathanaelbosch, line 108: Why should we not implement this, too? That way, the diffusion would behave like a real step function, and it should be possible since the diffusion object keeps track of `times` already.

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,120 @@
...
+        self._seen_diffusions += 1
+
+        if self.diffusion is None:
+            self.diffusion = diffusion
+        else:
+            a = 1 / self._seen_diffusions
+            b = 1 - a
+            self.diffusion = a * diffusion + b * self.diffusion
+
+    def postprocess_states(self, states, locations):
```
nathanaelbosch, line 92: What about "calibrate_all_states" or something like this?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: Should we really ad this as an option, even though it was mostly just for us to experiment with stuff? If we want to fully support that, we should probably rather implement another diffusion model?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -241,8 +212,8 @@ def step(self, t, t_new, current_rv):
         local_errors = self._estimate_local_error(
             pred_rv,
             t_new,
-            self.sigma_squared_mle * proc_noise_cov,
-            np.sqrt(self.sigma_squared_mle) * proc_noise_cov_cholesky,
+            local_squared_diffusion * proc_noise_cov,
```
nathanaelbosch, line 215: So even in the fixed diffusion model, we use just the diffusion estimate from this step, not the full MSE? That would be a different behaviour to the way ProbNumDiffEq.jl treats the fixed diffusion.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: It is a bit confusing to have such functionality in `rvlist_to_odesol`, it feels like it should rather go into "postprocess".

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
nathanaelbosch: I think you can also remove the statement about lower-case and upper-case letters.

nathanaelbosch (changes requested):

src/probnum/diffeq/odefiltsmooth/diffusions.py
pnkraemer: I think in principle only `__call__` would do most of the work. Though I am note sure whether this is compatible with all the `find_index` stuff in `KalmanPosterior`. However, we can leave it at just `__call__` for now and add a `__getitem__` whenever we see that it is necessary?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: Do you have an idea how to put this into a diffusion model? 

I do not really; in fact, I think this current way is the cleanest way of handling this re-prediction,  because only this way can we use diffusions that are not informed by any ODE solver/ transition behaviour. Do you see what I mean here? 

On a side note, this is also kind of how I think about these things mathematically: re-prediction with a calibrated diffusion is always an option (even though with a constant diffusion this does not really make a lot of sense). What do you think?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: Hmmm okay. Do you think rvlist_to_odesol should happen before or after postprocess?

src/probnum/diffeq/odefiltsmooth/odefiltsmooth.py
pnkraemer: alrighty

src/probnum/diffeq/odefiltsmooth/diffusions.py
pnkraemer: true. I have not done it yet, because it is not needed anywhere yet and because I wanted to keep this PR as small as possible. Do you want me to include this here? I can do it.

src/probnum/diffeq/odefiltsmooth/diffusions.py
pnkraemer: yes, that sounds better than `postprocess_states`. Do you think there is a way of making it even clearer that it uses the global sigma estimate?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: Hmmm yes I see the point. Personally, I think this is a better option in the sense of not breaking when ODEs get more challenging, etc.. But you're right in the sense that we should offer all options. I will ping you offline :)

src/probnum/diffeq/odefiltsmooth/diffusions.py
nathanaelbosch: That sounds good :+1: As you say, if we see that we do need a `__getitem__` we can still add it.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
nathanaelbosch: I think I don't agree with "re-prediction with a calibrated diffusion is always an option". Depending on the model you consider, there should be a right way to do things: The dynamic diffusion needs to be considerered in a prediction, but the constant diffusion can be disregarded if we calibrate at the end. 

I'm also not 100% sure what the "pseudo-dynamic diffusion" means exactly, which is my main reason to why I'm careful to add a support for it. How would we write it down?

src/probnum/diffeq/odefiltsmooth/diffusions.py
nathanaelbosch: If it's fine for you, I think for the sake of completeness it should ideally be implemented as part of this PR. Thanks!

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: I added a (IMO) rather thorough docstring for `step()`. Does that clear things up now? If not, let's talk in person tomorrow.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: Anyway, changed it.

pnkraemer:
Alright, I included all the demanded updates.
Perhaps two things are worth explaining here:
1. I refactored the step (for some reason, it did not really do the dynamic sigma because it did not start with an error-free state). Now, it does all these things. I tried my best to explain the function thoroughly, in both block-comments inside the code as well as in a rather long docstring. This should also clean up the discussion we had in one of the comment sections, I hope. If not, let's talk tomorrow :) 
2. The `re_predict...` variable is still in `IVPSolver`, but it has a leading underscore with the explicit recommendation to leave it at its default value in the docstring. I think this "re-prediction" decision should happen at ODE solver level (because the diffusion doesn't know the dynamics, and measurement model), but I also agree that a user should not be encouraged to play around freely with those decisions. 
3. I also moved up the `DiffusionType` to `probnum.type` and renamed it to `ToleranceDiffusionType`, because now it can describe atol, rtol, errorest, and diffusion (all of which kind of describe the scale of different modes of a system). This now communicates that floats and arrays can be passed around freely as long as they have the proper dimensionality (there is also a warning that vector-valued diffusions are not there yet).

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -25,28 +28,47 @@ class GaussianIVPFilter(ODESolver):
...
-    ivp
+    ivp :
         Initial value problem to be solved.
-    prior
-        Prior distribution.
-    measurement_model
-        ODE measurement model.
-    with_smoothing
+    dynamics_model :
+        dynamics_model distribution.
```
nathanaelbosch, line 34: Maybe just "Dynamics model" or something like that :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -25,28 +28,47 @@ class GaussianIVPFilter(ODESolver):
...
         For more convenient construction, consider :func:`GaussianIVPFilter.construct_with_rk_init` and :func:`GaussianIVPFilter.construct_with_taylormode_init`.
-    initrv
+    initrv :
         Initial random variable to be used by the filter. Optional. Default is `None`, which amounts to choosing a standard-normal initial RV.
         As part of `GaussianIVPFilter.initialise()` (called in `GaussianIVPFilter.solve()`), this variable is improved upon with the help of the
         initialisation algorithm. The influence of this choice on the posterior may vary depending on the initialization strategy, but is almost always weak.
+    diffusion_model :
+        Diffusion model. This determines which kind of calibration is used. We refer to Bosch et al. (2020) [1]_ for a survey.
+    _repeat_after_calibration :
+        Repeat the step after the initial calibration. Improves the result of each step, but slightly increases computational complexity.
```
nathanaelbosch, line 49: I might be getting annoying, but I don't agree with "improves the result of each step" :sweat_smile: This combination of "re-predicting with the current MLE of a constant diffusion" is at the very experimental, and I'm not sure we can claim something like this right now.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -25,28 +28,47 @@ class GaussianIVPFilter(ODESolver):
...
+        Optional. A default value is inferred from the type of diffusion_model model:
+        if it is a `ConstantDiffusion`, this parameter is set to `False`;
+        if it is a `PiecewiseConstantDiffusion`, this parameter is set to `True`.
```
nathanaelbosch, line 52: I'm also wondering: Do you see an argument for using the `PiecewiseConstantDiffusion`, but _not_ predicting again with the correct diffusion? If you do everything in the correct order (predict mean, measure, predict cov, compute measurement cov) then it should also not have any increased computational cost.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -166,175 +162,269 @@ def init_implementation(f, y0, t0, prior, initrv, df=None):
...
 
-        # 4. Update
-        zero_data = np.zeros(meas_rv.mean.shape)
-        filt_rv, _ = self.gfilt.measurement_model.backward_realization(
-            zero_data, pred_rv, rv_forwarded=meas_rv, gain=info["gain"]
+        # Estimate local diffusion_model and error
+        diffusion_error_free = self.diffusion_model.calibrate_locally(
+            meas_rv_error_free
+        )
+        diffusion_full = self.diffusion_model.calibrate_locally(meas_rv)
```
nathanaelbosch, line 312: For a `ConstantDiffusion`, the object here is not actually _the_ diffusion or an estimate of the true diffusion, it's just some local quantity, used later to estimate the actual diffusion. Maybe `local_diffusion_full` and `local_diffusion_error_free` or something like that?

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -166,175 +162,269 @@ def init_implementation(f, y0, t0, prior, initrv, df=None):
...
+        diffusion_for_calibration = self.diffusion_model.update_current_information(
```
nathanaelbosch, line 313: For the `ConstantDiffusion`, in the way as described in Filip's paper, this should just be the identity since there is no calibration, right? If I understand correctly, here you use the current MLE of the global diffusion, and I'm once more not sure we should do that.

nathanaelbosch: Ok I'm mistaken. For the local error estimates and adaptive steps, it makes sense that we need to use something here, e.g. the current MLE of the diffusion :) The naming confused me, since I didn't see a "calibration" happening here in that scenario, but that might be on me since I tend to separately think about calibration and error estimation, even though both are closely linked.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -166,175 +162,269 @@ def init_implementation(f, y0, t0, prior, initrv, df=None):
...
+                rv=current_rv, t=t, dt=dt, _diffusion=diffusion_for_calibration
+            )
+            meas_rv, info = self.measurement_model.forward_rv(
+                rv=pred_rv, t=t_new, compute_gain=True
+            )
+            gain = info["gain"]
+        else:
+            # Combine error-free and noisy-component predictions into a full prediction.
+            # (The measurement has been updated already.)
+            full_pred_cov_cholesky = utils.linalg.cholesky_update(
```
nathanaelbosch, line 332: If this is indeed not tested, it should probably be :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -166,175 +162,269 @@ def init_implementation(f, y0, t0, prior, initrv, df=None):
...
+        zero_data = np.zeros(meas_rv.mean.shape)
+        filt_rv, _ = self.measurement_model.backward_realization(
+            zero_data, pred_rv, rv_forwarded=meas_rv, gain=gain
         )
-        err = np.linalg.norm(local_errors)
 
-        return filt_rv, err
+        proj = self.dynamics_model.proj2coord(coord=self._reference_coordinates)
+        reference_state = np.maximum(proj @ filt_rv.mean, proj @ current_rv.mean)
+        return filt_rv, local_errors, reference_state
```
nathanaelbosch, line 354: Do we have an alternative to "state"? In my mind, "state" describes the d*(q+1)-dimensional object of the full random process, not a projection. But if we don't have a nice alternative,  "state" is fine :)

nathanaelbosch (changes requested):

src/probnum/diffeq/odefiltsmooth/diffusions.py
pnkraemer: `estimate_locally`

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,164 @@
...
+        """Calibrate a set of ODE solver states after seeing all the data."""
+        raise NotImplementedError
+
+
+class ConstantDiffusion(Diffusion):
+    """Constant diffusion and its calibration.
+
+    Parameters
+    ----------
+    use_global_estimate_as_local_estimate :
```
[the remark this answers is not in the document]
pnkraemer: remove this parameter, always use the local sigma

src/probnum/diffeq/odefiltsmooth/diffusions.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+
+    def __repr__(self):
+        return f"ConstantDiffusion({self.diffusion})"
+
+    def __call__(self, t) -> ToleranceDiffusionType:
+        return self.diffusion
+
+    def update_current_information(
+        self, full_diffusion, error_free_diffusion, t
```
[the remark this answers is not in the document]
pnkraemer: decision: do everything in `estimate_locally` (and perhaps rename it to `estimate_locally_and_update_in_place`), pass both meas_rvs to `estimate_locally`, and let that function return the `local` sigma that can then be used for error estimation. `update_current_information` is not necessary anymore.

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: decision: remove this parameter entirely, set this flag "secretly" inside the init by matching the diffusion models.

pnkraemer:
As far as I remember, all the required changes are made. What is left is including the diffusion into the posteriors (dense output and sampling).

src/probnum/statespace/transition.py
```diff
@@ -253,7 +255,11 @@ def smooth_list(self, rv_list, locations, _previous_posterior=None):
             List of random variables to be smoothed.
         locations :
             Locations :math:`t` of the random variables in the time-domain. Used for continuous-time transitions.
-        _previous_posterior
+        _diffusion_list :
+            List of diffusions that correspond to the intervals in the locations.
+            If `locations=(t0, ..., tN)`, then `_diffusion_list=(d1, ..., dN)`, i.e. it contains one element less.
+            Optional.
```
nathanaelbosch, line 261: Is the diffusion list really optional here?

nathanaelbosch: I guess this belongs to `_previous_posterior`

src/probnum/statespace/transition.py
```diff
@@ -306,6 +315,10 @@ def jointly_transform_base_measure_realization_list_backward(
             List of random variables to be jointly sampled from.
         t :
             Locations of the random variables in the list. Assumed to be sorted.
+        _diffusion_list :
+            List of diffusions that correspond to the intervals in the locations.
+            If `locations=(t0, ..., tN)`, then `_diffusion_list=(d1, ..., dN)`, i.e. it contains one element less.
+            Optional.
```
nathanaelbosch, line 321: This line might belong to the `_previous_posterior` argument

src/probnum/statespace/transition.py
```diff
@@ -368,6 +383,10 @@ def jointly_transform_base_measure_realization_list_forward(
             Initial random variable.
         t :
             Locations of the random variables in the list. Assumed to be sorted.
+        _diffusion_list :
+            List of diffusions that correspond to the intervals in the locations.
+            If `locations=(t0, ..., tN)`, then `_diffusion_list=(d1, ..., dN)`, i.e. it contains one element less.
+            Optional.
```
nathanaelbosch, line 389: Same thing again: This might belong to the next parameter :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
```diff
@@ -301,85 +286,138 @@ def step(self, t, t_new, current_rv):
...
+            rv_list,
+            self.dynamics_model,
+            diffusion_model=self.diffusion_model,
         )
 
         if self.with_smoothing is True:
-            rv_list = self.dynamics_model.smooth_list(rv_list, locations)
-            kalman_posterior = pnfs.SmoothingPosterior(
+
+            # if isinstance(self.diffusion_model, statespace.PiecewiseConstantDiffusion):
```
nathanaelbosch, line 404: Should this commented-out code be kept?

nathanaelbosch (changes requested):
I looked through all the files once more and made a few very small comments. Overall the PR looks good to me! Though I have to admit that I'm not 100% sure about everything that's happening, just because it's been such a long time ago. But from what I remember, we reached a consensus on almost everything already anyways :)

There are a few open comments from back in March. Could you go through them and comment or resolve as you think is appropriate?

Apart from that, we should be able to merge this one in soon :)

src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
pnkraemer: it is part of the constant diffusion model tests. These are tested. I don't know why the test coverage does not reflect that. (The lines below are the basic update schemes and they are marked as not-tested either, but if they were wrong NONE of the tests would pass).

pnkraemer:
> I looked through all the files once more and made a few very small comments. Overall the PR looks good to me! Though I have to admit that I'm not 100% sure about everything that's happening, just because it's been such a long time ago. But from what I remember, we reached a consensus on almost everything already anyways :)
Same here. It's been a while, but I remember that we agreed on things. 

> There are a few open comments from back in March. Could you go through them and comment or resolve as you think is appropriate?
Done. One of them I commented (why it should be no concern IMO), the other ones are resolved.

Apart from that, we should be able to merge this one in soon :)

nathanaelbosch (approved):
Looks good to me. Thanks for the PR!

nathanaelbosch (changes requested):
...except for the failing tests :/ Are these an issue or not?

pnkraemer:
> ...except for the failing tests :/ Are these an issue or not?

nothing to worry about. it was a weird seed issue in the particle filtering tests. I fixed it and now it should be back to bulletproof.

nathanaelbosch (approved):
Thanks for fixing the tests!

PR #353, opened 2021-03-13
Basic Particle Filtering

pnkraemer:
This PR introduces fundamental particle filtering into probnum. To this end, it also introduces a `Categorical` random variable.

There is a `ParticleFilter` class that implements a bootstrap PF as well as one with an importance density that is based on an approximate Gaussian filter (EKF, UKF, etc.). There is also a `ParticleFilterPosterior` but it is not really worth talking about.

Other than being a nice to have feature (of course), a PF implementation will also help with not overfitting the `BayesFiltSmooth` interface to Kalman filtering. 

There is also the option of reproducing PF ode solvers now (https://arxiv.org/abs/1810.03440).


pnkraemer:
@schmidtjonathan I will go over this PR again on Monday and then assign you as a reviewer some time next week :)

docs/source/probnum.mplstyle
```diff
@@ -74,7 +74,7 @@ legend.fancybox      : False     ## if True, use a rounded box for the
 #### FIGURE
 ## See http://matplotlib.org/api/figure_api.html#matplotlib.figure.Figure
 figure.figsize       : 8, 4.5     ## figure size in inches
-figure.dpi           : 150        ## figure dots per inch
+figure.dpi           : 100        ## figure dots per inch
```
pnkraemer, line 77: there is no real point in a value > 100 here, since jupyter kind of ignores it anyway.

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
```diff
@@ -0,0 +1,218 @@
...
+    num_particles = len(categ_rv.support)
+    new_support = categ_rv.sample(size=num_particles)
+    new_event_probs = np.ones(categ_rv.probabilities.shape) / len(
+        categ_rv.probabilities
+    )
+    return random_variables.Categorical(
+        support=new_support,
+        probabilities=new_event_probs,
+        random_state=categ_rv.random_state,
+    )
```
pnkraemer, line 36: should this be in `probnum.random_variables`?

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
```diff
@@ -0,0 +1,218 @@
...
+
+    def __init__(
+        self,
+        dynamics_model: Union[statespace.LTISDE, statespace.DiscreteGaussian],
+        measurement_model: statespace.DiscreteGaussian,
+        initrv: random_variables.RandomVariable,
+        num_particles: IntArgType,
+        linearized_measurement_model: Optional[statespace.DiscreteGaussian] = None,
+        with_resampling: bool = True,
+        resampling_percentage_threshold: FloatArgType = 0.1,
```
pnkraemer, line 81: one thing that is not part of `ParticleFilter`s currently is any sort of random state. I'd say we leave it for now and wait until the sampling vs. transforming base measure stuff is resolved?

pnkraemer: Or shall I introduce a `random_state=None` here? your call :)

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
pnkraemer: There should probably be a random state, too.

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
```diff
@@ -0,0 +1,218 @@
...
+        2. Computing a "proposal" random variable.
+        This is either the prior dynamics model or the output of a filter step
+        of an (approximate) Gaussian filter.
+        3. Sample from the proposal random variable. This is the "new" particle.
+        4. Propagate the particle through the measurement model.
+        This is required in order to require the PDF of the resulting RV at
+        the data. If this is small, the weight of the particle will be small.
+        5. Compute weights ("event probabilities") of the new particle.
+        This requires evaluating the PDFs of all three RVs (dynamics, proposal, measurement).
+
```
pnkraemer, line 159: is this stuff clear?

tests/test_random_variables/test_categorical.py
```diff
@@ -0,0 +1,107 @@
+"""Tests for the categorical random variable."""
```
pnkraemer, line 1: since you know pytest better than I do, please dont hesitate to criticise!

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
```diff
@@ -0,0 +1,218 @@
...
+from ._particle_filter_posterior import ParticleFilterPosterior
+
+
+def effective_number_of_events(categ_rv: random_variables.Categorical) -> float:
+    """Effective number of events in the support of a categorical random variable.
+
+    In a particle filter, this is used as the effective number of
+    particles.
+    """
+    return 1.0 / np.sum(categ_rv.probabilities ** 2)
```
pnkraemer, line 20: this is not a property of the categorical RV because I dont think it has any meaning outside of using categorical RVs for importance sampling

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
```diff
@@ -0,0 +1,218 @@
...
+        the data. If this is small, the weight of the particle will be small.
+        5. Compute weights ("event probabilities") of the new particle.
+        This requires evaluating the PDFs of all three RVs (dynamics, proposal, measurement).
+
+        After this is done for all particles, the weights are normalized in order to sum to 1.
+        """
+        new_weights = randvar.probabilities.copy()
+        new_support = randvar.support.copy()
+
+        for idx, (particle, weight) in enumerate(zip(new_support, new_weights)):
```
pnkraemer, line 165: these loops are slow, but at least they are rather easy to read (and I had no idea how to optimize them so that the difference is more than incremental)

pnkraemer (commented):
Please feel free to go at me now. If there are any theoretical questions, you can also ping me offline.

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
pnkraemer: The random state is inferred from the initial random variable.

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
pnkraemer: Moved to `Categorical` anyway.

src/probnum/filtsmooth/gaussfiltsmooth/unscentedkalman.py
```diff
@@ -264,25 +264,7 @@ def from_ode(
...
-            input_dim=prior.dimension,
-            output_dim=prior.spatialdim,
-            state_trans_fun=dyna,
-            proc_noise_cov_mat_fun=diff,
-            proc_noise_cov_cholesky_fun=diff_cholesky,
+        discrete_model = pnss.DiscreteGaussian.from_ode(
+            ode=ode, prior=prior, evlvar=evlvar
         )
-        return cls(disc_model)
+        return cls(discrete_model)
```
schmidtjonathan, line 270: Just a minor thing: Do you think it sensible to outsource this in a similar way in the `EKF`? With an optional `jacobian` kwarg and a EK0 or EK1 check in the `DiscreteEKFComponent.from_ode` method?

pnkraemer: Do you mean in the sense of `from_ode(ivp, prior, evlvar=0., ek0_jacobian=False)` that essentially moves the `ek0_or_ek1` decision up one level?

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
```diff
@@ -0,0 +1,219 @@
...
+        model and compute new weights via the respective PDFs.
+        """
+        proposal_rv = self.dynamics_to_proposal_rv(dynamics_rv, data=data, t=stop)
+        proposal_state = proposal_rv.sample()
+        meas_rv, _ = self.measurement_model.forward_realization(proposal_state, t=stop)
+
+        log_proposal_weight = (
+            meas_rv.logpdf(data)
+            + dynamics_rv.logpdf(proposal_state)
+            - proposal_rv.logpdf(proposal_state)
```
schmidtjonathan, line 193: Am I understanding correctly, that for the bootstrap PF, this just cancels?

src/probnum/random_variables/_categorical.py
```diff
@@ -0,0 +1,105 @@
...
+        self,
+        probabilities: np.ndarray,
+        support: Optional[np.ndarray] = None,
+        random_state: Optional[RandomStateArgType] = None,
+    ):
+        # The set of events is names "support" to be aligned with the method
+        # DiscreteRandomVariable.in_support().
+
+        num_categories = len(probabilities)
+        self._probabilities = np.asarray(probabilities)
```
schmidtjonathan, line 36: We should probably enforce the sum-to-unity constraint for the probabilities, right?
For sampling, `np.random.choice` will throw an error, but `_pmf_categorical` will not complain returning unreasonable values.

tests/test_filtsmooth/test_particlefiltsmooth/test_particle_filter.py
```diff
@@ -0,0 +1,123 @@
...
+# Resampling percentage threshold checks that
+# resampling is performed a) never, b) sometimes, c) always
+all_resampling_configurations = pytest.mark.parametrize(
+    "resampling_percentage_threshold", [-1.0, 0.1, 2.0]
+)
+
+
+@pytest.fixture
+def num_particles():
+    return 4
```
schmidtjonathan, line 34: How realistic is the usecase for huge numbers of particles? Because, spontaneously, I suspect that the algorithm (e.g. the estimator for the number of effective particles might get into numerical trouble), in this case. I have no practical experience, but maybe we would want to check when it starts to get difficult?

tests/test_random_variables/test_categorical.py
schmidtjonathan: Do I ? 😅 I have nothing to criticise

tests/test_random_variables/test_categorical.py
```diff
@@ -0,0 +1,122 @@
...
+    return random_variables.Categorical(
+        probabilities=probabilities, support=support, random_state=random_state
+    )
+
+
+@all_random_states
+@all_supports
+def test_probabilities(categ, probabilities):
+    assert categ.probabilities.shape == (NDIM,)
+    np.testing.assert_allclose(categ.probabilities, probabilities)
```
schmidtjonathan, line 50: Isn't that the same object reference? :D
We could test here whether or not the probabilities sum to one, but that's basically clear by the definition of the `probabilities()`-fixture, so... 🤷

schmidtjonathan (approved):
That's great! I love how neatly the Particle Filter fits into the refactored (`forward`/`backward`) framework.
The code, as well as the tutorial notebook are very clear and understandable. Thanks!

src/probnum/random_variables/_categorical.py
```diff
@@ -0,0 +1,105 @@
...
+
+        parameters = {
+            "support": self._support,
+            "event_probabilities": self._probabilities,
+            "num_categories": num_categories,
+        }
+
+        def _sample_categorical(size=()):
+            mask = np.random.choice(
+                np.arange(len(self.support)), size=size, p=self.probabilities
```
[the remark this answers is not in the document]
pnkraemer: I thought so as well in the first place, but this wont work:
```python
>>> import numpy as np
>>> np.random.choice(np.arange(4))
1
>>> np.random.choice(np.random.rand(4, 4))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "mtrand.pyx", line 905, in numpy.random.mtrand.RandomState.choice
ValueError: a must be 1-dimensional
```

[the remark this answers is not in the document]
schmidtjonathan: Ah, I see! ✅ Nevermind, then 👍

src/probnum/filtsmooth/gaussfiltsmooth/unscentedkalman.py
pnkraemer: Hmmmm I am not so sure behind the logic here. I see what you're trying to do, but there should be a better way. For this PR, I'd propose to leave it as is and refactor this. (I am thinking #361)

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
pnkraemer: I add a sentence about the SIR to the doc

src/probnum/filtsmooth/particlefiltsmooth/_particle_filter.py
pnkraemer: yes. I add an if-else here, to make this point even more obvious.

tests/test_filtsmooth/test_particlefiltsmooth/test_particle_filter.py
pnkraemer: Good question. From a theoretical perspective: the more the better. From a runtime perspective: I am not sure whether we can afford too many particles here. I have hope, though, that we can vectorize the implementation a bit, which would drastically improve the computation.

tests/test_random_variables/test_categorical.py
pnkraemer: this test is mostly about being able to access `categ.probabilities` and that indeed the correct vector comes out. If one refactored `categ.probabilities -> categ.event_probabilities` we want a test to break. This is said test. 

Don't take the values of the vector too seriously though :)

src/probnum/random_variables/_categorical.py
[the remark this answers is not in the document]
pnkraemer: I included this piece of information into the docs. After all, I guess everyone will have that thought (I did, too).

src/probnum/random_variables/_categorical.py
pnkraemer: I am not sure. In principle, I would like to let this implementation not fail prematurely, if that makes sense. Who knows which kind of duck typing one would need in the future :) However, I see the point that perhaps we would like to throw meaningful errors. For the time being, I think that `np.random.choice` would return a sufficiently meaningful error message, but if you want me to change that, I will do so!

schmidtjonathan: My only concern is the `pmf` function. If you have something in mind for which the check would be troublesome, then I'll leave that decision to you :)

Issue #348, opened 2021-03-12
Naming and import conventions


pnkraemer:
The import style is currently not consistent across modules (in terms of when to use abbreviations and when not, and also the choices of abbreviations).

**Decision from the meeting:**
Rename `random_variables` into  ```probnum.randvars``` (follow up with ```probnum.randprocs``` soon).
Adhere to the google style guide more strictly: `from probnum import randvars, statespace, ...` (once `random_variables` is renamed, all the names are sufficiently short).

**Question:**
How about inside-module imports? The google style guide recommends to use full paths only, even if 2 modules are in the same package.
Do we do this, too? I wouldn't be against not following the google style guide here, because I find it easier to read with relative imports on an inside-package level. This is mostly because the classes in the modules are pulled up into a higher-level namespace usually. Thoughts?

PR #355, opened 2021-03-16
Rename random_variables and fix imports

pnkraemer:
This PR renames the `random_variables` package into `randvars` and includes the decisions from #348 in the styleguide. 

The imports of the `randvar` module are unified across the entire package (no more `pnrv` or `as rvs`), even on the notebooks.
The other imports do not follow this rule yet, I suggest that we update those slowly. 

There are two open questions left, which I comment in the source. Please let me know what you're thinking there.


STYLEGUIDE.md
```diff
@@ -35,6 +35,13 @@ If imports are shortened, the following conventions should be used. Full import
 paths are always acceptable.
 
 - `import probnum as pn`
+- `from probnum import randvars, linalg, diffeq, statespace`
+
+Please do not abbreviate import paths unnecessarily. We do **not** use the following imports:
+- `import probnum.random_variables as pnrv` or `import probnum.filtsmooth as pnfs` (correct would be `from probnum import randvars, filtsmooth`)
+- `from probnum import random_variables as rvs` or `import probnum.random_variables as rvs` (the `randvars` name is sufficiently short and does not need to be abbreviated)
+
+All of these rule obey the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html).
```
pnkraemer, line 45: There are two more factors I'd like to introduce here regarding imports, but I am not sure how subjective they are.

1. What about imports like `from probnum.diffeq import probsolve_ivp` in notebooks, for demonstration purposes? Currently this is featured quite heavily (especially in the linalg notebooks, if I remember correctly), but it deviates from the -- now -- fixed import style. I suggest we stick to `from probnum import diffeq` and avoid the other imports in notebooks as well. Thoughts? It does not need to be fixed yet, but I'd like to include it in the styleguide as a direction/
2. How about local relative imports. For instance, should the `_normal.py` import the base class as `from ._random_variables import RandomVariable`, or as `from probnum.randvars._random_variables import RandomVariable`, or as `from probnum import randvars` (which would introduce cyclic imports?) (there is more options, but I hope you get the question). Personally, I prefer importing the class `RandomVariable` directly here. I am not sure about the full path or a relative one in this case. Either way, this will deviate from the styleguide and I would like to write down a rule here.

On another note, shall I explicitly write down here that `typing` and `probnum.type` are excluded from the "full path" thing?

pnkraemer:
I had to make a change in the linear operator notebook. It threw an error (could it be that it has not been run for a while?).
The problem was that in the first cell, `B = A ** 4 + Id` returns a `scipy...._SumLinearOperator` which does not support `.todense()`. By removing the `**4` it worked again, because now it is a `probnum.linops.SumLinearOperator`, which it should be. It seems to me that this did not change the content of the notebook too much, but if so, please let me know a better solution (or we write an issue).

STYLEGUIDE.md
```diff
@@ -35,6 +35,13 @@ If imports are shortened, the following conventions should be used. Full import
...
```
JonathanWenger, line 44: All of these rule**s**...

STYLEGUIDE.md
JonathanWenger: ad 1) In my opinion importing functions in tutorials and docstrings is fine since the styleguide applies to the package itself, not how to use the package. However the NumPy and SciPy docstrings and tutorials do stick to exactly the style you suggest, so I will leave it up to you. If you want to put it in, I am all for it.

ad 2) I remember that there were a lot of issues with cyclic imports and having two instances of classes leading to strange bugs where `RandomVariables` were not instances of `RandomVariables`. Unfortunately I do not remember what the rule is that should be applied here.

ad 3) I think it is a good idea to explicitly write out the exceptions.

JonathanWenger (commented):
Thanks for taking care of this! My comments:

1. According to the styleguide should the imports be
```python
import probnum.randvars
```
instead of ```import probnum.randvars as rvs/pnrv``? there are still a few (in the tests for example) which have this pattern.

2. I think you missed the notebook explaining probabilistic numerical methods. There is still one import referencing `random_variables`.

STYLEGUIDE.md
pnkraemer: Thanks for the feedback! 
1) I will make it all consistent then. I think it is better to do the same everywhere, and this way we are not forcing anyone to use the same style :) 
2) Then I'd say we leave it at the relative imports whenever modules "share a namespace" (you know what I mean). This is how it is currently done and does not seem to break anything. (Other than that, I prefer its readability anyway)
3) Alright!

pnkraemer:
True, I missed going over the tests and over the pn tutorial notebook. Will do so now.

JonathanWenger (approved):

PR #356, opened 2021-03-18
Random Processes

### LINKED ISSUE #44: Create a class RandomProcess
Create a module `randomprocesses` analoguosly to `randomvariables`. This module should contain a super class `RandomProcesses` with functions like `meanfunction` and `covfunction`.

Classes from `statespace` will inherit from `RandomProcesses` as well as future implementations (or wrappers) of Gaussian processes.

JonathanWenger:
# In a Nutshell
This PR adds interfaces for random processes as a way to represent uncertainty about a function and functions with stochastic output.  

# Details
- Interfaces for random and Gaussian processes
- Basic tests
- Random process arithmetic is _not_ part of this PR
- `RandomProcess` is now an abstract class and therefore needs to be subclassed

# Classes

- `RandomProcess`: Objects generalizing functions by adding uncertainty via a covariance function.
- `GaussianProcess`: Standard Gaussian processes without conditioning functionality.
- `GaussMarkovProcess`: Interface for Gaussian processes with the Markov property.

This closes #44.


src/probnum/randprocs/_gauss_markov_process.py
```diff
@@ -0,0 +1,75 @@
...
+        self.x0 = np.asarray(x0).reshape(1, -1)
+
+        super().__init__(
+            input_dim=1,
+            output_dim=1 if self.x0.ndim == 0 else self.x0.shape[0],
+            mean=lambda t: self.__call__(t).mean,
+            cov=lambda t: self.__call__(t).cov,
+        )
+
+    def __call__(self, x: _InputType) -> randvars.Normal:
```
pnkraemer, line 66: While I understand that `x` is the natural input choice for a general random process, it makes me feel uncomfortable to call a Markov-process with `markov_process(x=0.1)` instead of `t`... 
I don't have an idea how to solve this problem though. Either, we could knowingly deviate from this interface in subclasses (and use `t` wherever appropriate), or change `x` to something like `point/location/input/value/...` which could be more general, or suck it up and use `x`. I dont know what to do. What do you think? 
In any case, this relates to the `x0` comment in `GaussMarkovProcess.__init__`.

src/probnum/randprocs/_gauss_markov_process.py
```diff
@@ -0,0 +1,75 @@
...
+        """Closed form solution to the SDE evaluated at ``x`` as defined by the linear
+        transition."""
+        # TODO: currently the SDE Models do not support arbitrary steps as defined by
+        #  the increments in ``x``, only fixed step sizes ``dt```. This has to
+        #  be added first before __call__, mean, etc. work as defined by the
+        #  RandomProcess interface.
+
+        # return self.linear_sde.forward_rv(rv=self.x0, t=self.t0, dt=x[-1] - x[-2])
```
pnkraemer, line 74: At first sight (and sorry for the directness, it is not meant badly!) I dont think this is true.

the code `self.linear_sde.forward_rv(rv=self.x0, t=self.t0, dt=(t-self.t0)` should work in order to evaluate the GM process at a single point `t`. Evaluating it at multiple steps `(t1, t2, ...)` is a different story. It is possible (I can post a link to how), but neither is it readily available (I've never needed it), nor is it trivial to implement.

pnkraemer: Maybe this would also motivate providing the interface for at least a `cov` function (the mean function should be possible directly via forward_rv things). This could then properly subclass the GP interface (because now there would be a real cov), and would help encapsulating a rather weird function.

src/probnum/randprocs/_gaussian_process.py
```diff
@@ -0,0 +1,110 @@
...
+           [0.8824969 , 1.        , 0.8824969 , 0.60653066, 0.32465247],
+           [0.60653066, 0.8824969 , 1.        , 0.8824969 , 0.60653066],
+           [0.32465247, 0.60653066, 0.8824969 , 1.        , 0.8824969 ],
+           [0.13533528, 0.32465247, 0.60653066, 0.8824969 , 1.        ]])
+    """
+
+    def __init__(
+        self,
+        mean: Callable[[_InputType], _OutputType],
+        cov: Union[Callable[[_InputType], _OutputType], kernels.Kernel],
```
pnkraemer, line 71: how much do we want to enforce that a user provides a `Kernel` object as a covariance? If for instance the type would only be `kernels.Kernel` then callable code would still work (with some sort of duck-typing) but not force us into any promises (or, alternatively, we could wrap a callable into a kernel internally).

pnkraemer: If we only require a kernel, it would also remove the need for providing `input_dim` and `output_dim` and checks that they align with the kernel. 
I am not saying this is the way to go, but maybe something to think about (if you have thought about it already, then this will be easy to resolve :) )?

src/probnum/randprocs/_gaussian_process.py
```diff
@@ -0,0 +1,110 @@
...
+        self._meanfun = mean
+        self._covfun = cov
+        super().__init__(
+            input_dim=input_dim,
+            output_dim=output_dim,
+            dtype=np.dtype(np.float_),
+        )
+
+    def __call__(self, x: _InputType) -> randvars.Normal:
+        return randvars.Normal(mean=self.mean(x), cov=self.cov(x))
```
pnkraemer, line 94: is it on purpose that the dtype is automatically inferred from mean or cov?

src/probnum/randprocs/_random_process.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def __repr__(self) -> str:
+        return (
+            f"<{self.__class__.__name__} with "
+            f"input_dim={self.input_dim}, output_dim={self.output_dim}, "
+            f"dtype={self.dtype}>"
+        )
+
+    @abc.abstractmethod
+    def __call__(self, x: _InputType) -> randvars.RandomVariable[_OutputType]:
```
pnkraemer, line 66: See my `x` comment in `GaussMarkovPosterior` :)

src/probnum/randprocs/_random_process.py
```diff
@@ -0,0 +1,253 @@
...
+
+        Returns
+        -------
+        cov
+            *shape=(), (output_dim, output_dim), (n0, n1) or (n0, n1, output_dim,
+            output_dim)* -- Covariance of the process at ``x0`` and ``x1``. If
+            only ``x0`` is given the kernel matrix :math:`K=k(X_0, X_0)` is computed.
+        """  # pylint: disable=trailing-whitespace
+        raise NotImplementedError
+
```
pnkraemer, line 145: I think it would be quite useful to have a method along the lines of `marginal_cov(pointset)` which computes the covariances of all the respective points but not the covariance between points. This will be the biggest usecase for Filtering and ODESolver posteriors and having access to this method would allow implementing a fast way. 

Perhaps that is a pattern that we want to put more into the center? How about a method `self.marginal(x)` that computes the marginal distributions (i.e. kind of like a vectorised call, where the cross-terms are ignored). This will be 999/1000 usecases for GaussMarkov process calls, and since I agree that calling cov() on a set of points assembles the full, large covariance matrix, I'd like to have access to the other entries. In any case, if you agree that this would be fun, it suffices to place the interface here. I will take care of the GM implementation.

pnkraemer: (Oh and perhaps having a fast version of a `marginal`-like function could speed up `var` and `std` below.)

src/probnum/randprocs/_random_process.py
```diff
@@ -0,0 +1,253 @@
...
+            returned.
+        size
+            Size of the sample.
+        random_state :
+            Random state of the random process. If None (or np.random), the global
+            :mod:`numpy.random` state is used. If integer, it is used to seed the local
+            :class:`~numpy.random.RandomState` instance.
+        """
+        if x is None:
+            return lambda x0: self._sample_at_input(
```
pnkraemer, line 222: I think this is a good idea, I am glad that you suggested it months back.
One tiny suggestion: could we maybe replace the `lambda` function with a full one here? Or is there any reason for an anonymous one other than that it is a one-liner?

pnkraemer (commented):
Thanks a lot for this PR! I think it looks great, and hope my specific comments are not too petty.

I have two major(ish) questions/suggestions.
1. I think it would be useful to have a `MarkovProcess` class (that takes an initial random variable and a `Transition`) as an interface for a `GaussMarkovProcess` class. This way, one could define non-Gaussian priors (which are interesting for, e.g., particle filtering). It would be fine for me to add it myself later, but perhaps it is good to set these interfaces up as early as possible.
2. How do you feel about sampling implementations in terms of a push forward of a base measure? While having this as an abstract method may be too restrictive for a casual random process user (if these people exist ;)), at least having an empty function here would force us to establish a unified interface. 
For the `TimeSeriesPosterior` objects, there is a function `transform_base_measure_realizations` which does precisely that. There does not need to be any implementation, but perhaps the interface can be provided in this PR already? What do you think? Who knows, maybe a `transform_base_measure_realizations` would imply another method `sample_from_base_measure(self, size, random_state)` -- I dont know, but if it does, I'd like to decide that early :)

Thanks again!

src/probnum/randprocs/_random_process.py
pnkraemer: I thought about it some more and there is another point that may need to be taken into account:

The sample function as returned here is only valid if called exactly once; if called a second time, the resulting samples will be independent of the first samples (that were generated the first time), which they should not be if one thinks of this function returning a sample path that behaves like a function. I am fine with  this being written in the docstring, but do you perhaps have an idea how to solve this?

JonathanWenger:
> Thanks a lot for this PR! I think it looks great, and hope my specific comments are not too petty.
> 
> I have two major(ish) questions/suggestions.
> 
>     1. I think it would be useful to have a `MarkovProcess` class (that takes an initial random variable and a `Transition`) as an interface for a `GaussMarkovProcess` class. This way, one could define non-Gaussian priors (which are interesting for, e.g., particle filtering). It would be fine for me to add it myself later, but perhaps it is good to set these interfaces up as early as possible.

I ended up removing the `GaussMarkovProcess` interface for now and replaced it with a more general `MarkovProcess` I don't think it is worth at this point to have a separate class just for one difference in the type hint. This also introduced issues with multiple inheritance. I recommend `GaussMarkovProcess` to be added later or simply realized with multiple inheritance.

>     2. How do you feel about sampling implementations in terms of a push forward of a base measure? While having this as an abstract method may be too restrictive for a casual random process user (if these people exist ;)), at least having an empty function here would force us to establish a unified interface.
>        For the `TimeSeriesPosterior` objects, there is a function `transform_base_measure_realizations` which does precisely that. There does not need to be any implementation, but perhaps the interface can be provided in this PR already? What do you think? Who knows, maybe a `transform_base_measure_realizations` would imply another method `sample_from_base_measure(self, size, random_state)` -- I dont know, but if it does, I'd like to decide that early :)

I added an interface for it. Once we have `push_forward` in random variable for GPs this can just call `Normal.push_forward`. Sampling for GPs is realized via the underlying random variable. I think sampling should be handled by the associated random variable (at least for Gaussian processes).
> 
> Thanks again!

src/probnum/randprocs/_gaussian_process.py
JonathanWenger: There is no way to pass dtype no `Normal` random variables with the current implementation.

src/probnum/randprocs/_gauss_markov_process.py
```diff
@@ -0,0 +1,75 @@
...
+
+    Examples
+    --------
+    """
+    # pylint: disable=invalid-name
+
+    def __init__(
+        self,
+        linear_sde: statespace.LinearSDE,
+        x0: randvars.Normal,
```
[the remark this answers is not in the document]
JonathanWenger: I settled on `args` and `initrv` + `initarg` for Markov processes.

src/probnum/randprocs/_gauss_markov_process.py
```diff
@@ -0,0 +1,75 @@
...
+    GaussianProcess : Gaussian processes.
+
+    Examples
+    --------
+    """
+    # pylint: disable=invalid-name
+
+    def __init__(
+        self,
+        linear_sde: statespace.LinearSDE,
```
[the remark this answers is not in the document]
JonathanWenger: Now changed to `Transition` in `MarkovProcess`.

src/probnum/randprocs/_gauss_markov_process.py
JonathanWenger: Settled on `args`. Similar naming is used in JAX.

src/probnum/randprocs/_gaussian_process.py
JonathanWenger: I agree with forcing the user to implement their own `Kernel`. That way we can rely on the interface. Anything else seems unnecessarily complicated.

src/probnum/randprocs/_random_process.py
JonathanWenger: Nice catch! You're right, that is unexpected behavior. When passing a `random_state` everything works as expected. I changed the function such that even when sampling paths, a `random_state` is derived from the global RNG. Now sample paths are deterministic functions.

src/probnum/randprocs/_gauss_markov_process.py
JonathanWenger: Maybe I misunderstand, so please correct me if I am wrong. If I want to evaluate a Gauss-Markov process at time points `(t_0, t_1, ..., t_n)` not necessarily equidistant, then from looking at the interface of `forward_rv` I cannot do this in a single call. However I guess I could just iteratively apply it with new starting point `t_i` and `initrv` the output of forward_rv? The point is that the random process interface lets you evaluate at multiple inputs in a single call. For this PR I would leave the MarkovProcess simply as an interface if that is okay with you. You are much more familiar with your code than I am.

src/probnum/randprocs/_random_process.py
JonathanWenger: That's a good idea. We could even extend this to `RandomVariables`. Calling `.marginal()` on a random variable returns a batch of random variables as discussed in the last software meeting.

src/probnum/randprocs/_random_process.py
pnkraemer: Setting the random state is a good starting point, but I meant something more subtle than that. 

For instance, consider the setting where this function is used to sample from a GP. If this sample is called at a set of inputs, it returns the correct joint samples at the inputs. If however, this function is called again at a different set of inputs, what should happen? (a) the samples should be independent of the previous samples, which happens if they are computed in the same way as the previous samples without taking those into account; (b) the samples should be conditioned on the previous samples (by, roughly speaking, conditioning the GP on hitting those previous samples and sampling from the respective posterior GP now). Do you understand what I mean? I find this hard to express in a comment section, please ping me offline if you want me to elaborate :) 

I think in light of "a sample from a random process being a function", (b) would be desired. This is annoying to implement though (Nina has done it before). The question could be avoided altogether if we would somehow enforce that this sampling function can only be called once, which is why I suggested the solution in my previous comment above. 

In any case, I don't think this problem is a neck breaker for this PR and we can merge anyway as soon as the rest is approved, but let's make an issue about this in that case.

src/probnum/randprocs/_gaussian_process.py
```diff
@@ -0,0 +1,107 @@
...
+    ----------
+    mean :
+        Mean function.
+    cov :
+        Covariance function or kernel.
+
+    See Also
+    --------
+    RandomProcess : Random processes.
+    GaussMarkovProcess : Gaussian processes with the Markov property.
```
pnkraemer, line 33: this does not exist anymore, right?

src/probnum/randprocs/_markov_process.py
```diff
@@ -0,0 +1,91 @@
...
+        initarg: np.ndarray,
+        initrv: randvars.RandomVariable,
+        transition: statespace.Transition,
+    ):
+        self.initarg = initarg
+        self.initrv = initrv
+        self.transition = transition
+
+        super().__init__(
+            input_dim=1 if np.asarray(initarg).ndim == 0 else initarg.shape[0],
```
pnkraemer, line 51: I think the input dim of a Markov process should always be 1

src/probnum/randprocs/_markov_process.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def __call__(self, args: _InputType) -> randvars.RandomVariable:
+        # TODO: currently the statespace.Transition does not support arbitrary steps as
+        #  defined by
+        #  the increments in ``args``, only fixed step sizes ``dt```. This has to
+        #  be added first before __call__, mean, etc. work as defined by the
+        #  RandomProcess interface.
+
+        # return self.transition.forward_rv(rv=self.init_state, t=self.init_arg,
+        # dt=args[-1] - args[-2])
```
pnkraemer, line 64: I think this block comment should be removed. I will implement callability (is that a word?) soon.

src/probnum/randprocs/_markov_process.py
```diff
@@ -0,0 +1,91 @@
...
+        raise NotImplementedError
+
+    def mean(self, args: _InputType) -> _OutputType:
+        return self.__call__(args=args).mean
+
+    def cov(self, args0: _InputType, args1: Optional[_InputType] = None) -> _OutputType:
+        if args1 is None:
+            return self.__call__(args=args0).cov
+        raise NotImplementedError
```
pnkraemer, line 73: Since `__call__` is not available, I suggest to raise a NotImplementedError here as well.

src/probnum/randprocs/_markov_process.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _sample_at_input(
+        self,
+        args: _InputType,
+        size: ShapeArgType = (),
+        random_state: RandomStateArgType = None,
+    ) -> _OutputType:
+        randvar = self.__call__(args=args)
+        randvar.random_state = random_state
+        return randvar.sample(size=size)
```
pnkraemer, line 83: same here :)

src/probnum/randprocs/_random_process.py
```diff
@@ -0,0 +1,298 @@
...
+        Output dimension of the random process.
+    dtype :
+        Data type of the random process evaluated at an input. If ``object`` will be
+        converted to ``numpy.dtype``.
+
+    See Also
+    --------
+    RandomVariable : Random variables.
+    GaussianProcess : Gaussian processes.
+    GaussMarkovProcess : Gaussian processes with the Markov property.
```
pnkraemer, line 38: this does not exist anymore

tests/test_randprocs/test_markov_process.py
```diff
@@ -0,0 +1,12 @@
+"""Tests for Gauss-Markov processes."""
```
pnkraemer, line 1: for markov processes

tests/test_randprocs/test_markov_process.py
```diff
@@ -0,0 +1,12 @@
...
+
+from probnum import randprocs, randvars
+
+
+def test_gauss_markov_process_initrv(gauss_markov_process: randprocs.MarkovProcess):
```
pnkraemer, line 6: test_markov_process_initrv(markov_process: ...)

pnkraemer (approved):
Thanks for the updates! I left some specific comments in the diff, most of them are super minor, the only content-related comment is about the sampling. 

Ad Gauss-Markov process: That works for me. I will patch something together as needed, no need for you now to do anything there! 

Ad push_forward: thanks! Personally, I would've used a more verbose name (i.e. `push_forward_base_measure_realizations` or such), but this is super subjective and no complaint about your choice at all! :) On a side note: I don't think sampling can always be handled via the underlying random variable (well, at least not efficiently), so it is good to have this method here :) 

Ad `args`: This choice works. I will have to get used to this, especially to providing `initarg`, but at least for now, it does the job well. Thanks! Maybe with some more time, we can find another solution. Have you thought about updating the `Kernel` interface accordingly? It does not have to happen in this PR, but eventually, `GaussianProcess` and `Kernel` should have matching signatures I suppose.

src/probnum/randprocs/_markov_process.py
JonathanWenger: I'll keep it this way since it allows a future extension to Markov random fields.

src/probnum/randprocs/_markov_process.py
JonathanWenger: This implementation is valid once `__call__` is implemented and raises a `NotImplementedError` via `__call__` anyway, so I don't see a reason to remove the code. In particular it allows a subclass which only implements `__call__` but nothing else.

src/probnum/randprocs/_markov_process.py
JonathanWenger: see above.

tests/test_randprocs/test_markov_process.py
JonathanWenger: This test explicitly tests Gauss Markov processes, so the naming is valid.

JonathanWenger:
> Thanks for the updates! I left some specific comments in the diff, most of them are super minor, the only content-related comment is about the sampling.

Thanks for the review. I implemented the requested changes. Once you give me a final okay I will merge.

> Ad Gauss-Markov process: That works for me. I will patch something together as needed, no need for you now to do anything there!
> 
> Ad push_forward: thanks! Personally, I would've used a more verbose name (i.e. `push_forward_base_measure_realizations` or such), but this is super subjective and no complaint about your choice at all! :) On a side note: I don't think sampling can always be handled via the underlying random variable (well, at least not efficiently), so it is good to have this method here :)

I used just `push_forward` since `measure` and `sample` is part of the signature. I changed the signature to `push_forward(base_measure, sample)` to make it more explicit and semantically equivalent to `push_forward_base_measure_realizations`. 

> Ad `args`: This choice works. I will have to get used to this, especially to providing `initarg`, but at least for now, it does the job well. Thanks! Maybe with some more time, we can find another solution. Have you thought about updating the `Kernel` interface accordingly? It does not have to happen in this PR, but eventually, `GaussianProcess` and `Kernel` should have matching signatures I suppose.

I'm not particularly happy with it either, but all other choices seemed worse to me. Let's discuss this in the next software meeting and then adapt the interface in RandomProcess and Kernel in the next PR. That change will be super quick.

pnkraemer:
> Once you give me a final okay I will merge.
Okay. 

Let's turn the sampling thing  (https://github.com/probabilistic-numerics/probnum/pull/356#discussion_r608479566) into an issue then?!

src/probnum/randprocs/_random_process.py
JonathanWenger: Okay, I understand now. This is indeed quite tricky. Allowing only a single evaluation of the sample path is equivalent to just offering sampling with input locations. I opted for removing the base implementation for sampling sample paths as functions and raising a `NotImplementedError` instead. That way if we come up with a strategy on how to do this efficiently in the future (or for a specific subclass) we can still implement it. I left the tests in and marked them as failing so that we can easily readd them in the future. Let me know whether that works for you, then I will merge!

pnkraemer: It does, thanks for taking care of this!

PR #359, opened 2021-03-19
Constant normal arithmetic

pnkraemer:
This PR takes care of the second step of #335. That is, constant-normal arithmetic preserves Cholesky factors now, provided the `Normal` RV carries a precomputed Cholesky factor.


pnkraemer:
@marvinpfoertner I removed the "todo: ..." and instead explained *that* the matrix-variate update does not happen, why it does not happen, and reference the respective issue with the next steps (where I added this point).

marvinpfoertner (approved):

PR #360, opened 2021-03-19
Run Jupyter Notebooks in CI

### LINKED ISSUE #295: Build notebooks in CI
Currently the tutorials in jupyter notebook form are not run when changes are made. This can cause breaking changes in the tutorials. 

### Suggestion
Move tutorials build into a separate tox command `tox -e tutorials` and add this to the CI

JonathanWenger:
This PR adds a GitHub action which runs all notebooks using [`nbconvert`](https://nbconvert.readthedocs.io/en/latest/execute_api.html) and fails if any of the cells raise an error. It also fixes any errors in the existing notebooks and brings the notebooks more in line with the styleguide.

This resolves #295 .


pnkraemer (approved):
Wonderful, thanks for taking care of this!

PR #363, opened 2021-03-23
Quad

alpiges:
### contains
- Integration measures: Gaussian, Lebesgue
- Kernel embeddings (i.e. integrals over kernels): ExpQuad for the above measures
- (minimal) Bayesian quadrature method (just gets nodes the "Bayesian Monte Carlo" way: by sampling from the measure)
- BQ interface
- tests

### WIP
- more tests: multivariate integrals

Would be great to get some initial feedback!


pnkraemer:
That is really cool! I am especially a fan of the embeddings -- as I've always been saying: I think it is an absolute BQ goldmine to have them somewhere public! 

Since I have no idea what my role as `assignee` is, I will just ask random questions ;) 

How do you feel about introducing a `QuadratureRule` interface? I know that you probably don't plan on doing anything other than BQ (are there even such options?), but for two reasons that could be a good idea:
1. It could help making sure to stick to a common quadrature rule implementation scheme, with the result that a user that knows classical quadrature feels comfortable with BQ rules (a rather weak reason, but a reason nonetheless)
2. We could use this quadrature rule interface for all sorts of approximate Gaussian filters in `filtsmooth`. There, we need non-Bayesian quadrature rules. While it is also a fun idea to use BQ there (people have been doing this, most notably Hans in one of his early papers), we need to be more general.

Related to 2.: we currently have the unscented transform in `filtsmooth`. Strictly speaking, that is more of a quadrature rule than a filter implementation. Shall we move this to quad eventually? No need to do it now, but what are your thoughts regarding non-Bayesian-but-PN-relevant quadrature rules in the `quad` package?

JonathanWenger:
Quick question, I'll take a closer look later this week. Is the branch up-to-date with the current master? That would explain the drop in test coverage caused by what seems like tests not being run for existing code.

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,236 @@
...
+
+        if self.normalization_constant in [0, np.Inf, -np.Inf]:
+            raise ValueError(
+                "Normalization constant is too small or too large. "
+                "Consider setting normalized = False."
+            )
+
+        # Use scipy's uniform random variable since uniform random variables are not
+        # yet implemented in probnum
+        self.random_variable = scipy.stats.uniform(
```
JonathanWenger, line 167: You can cast this to a ProbNum random variable if necessary using [`probnum.randvars.asrandvar`](https://probnum.readthedocs.io/en/latest/automod/probnum.randvars.asrandvar.html#probnum.randvars.asrandvar)

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,236 @@
...
+        super().__init__(
+            dim=dim, domain=domain, random_state=random_state, name="Lebesgue measure"
+        )
+
+        # Set normalization constant
+        self.normalized = normalized
+        if self.normalized:
+            self.normalization_constant = 1 / np.prod(self.domain[1] - self.domain[0])
+        else:
+            self.normalization_constant = 1
```
JonathanWenger, line 157: You might want this to be a float, so `1.0`.

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,236 @@
...
+        else:
+            if (domain_a_dim > 1 or domain_b_dim > 1) and dim != np.max(
+                [domain_a_dim, domain_b_dim]
+            ):
+                dim_mismatch = True
+
+        if dim_mismatch:
+            raise ValueError(
+                "Domain limits must have the same length or at least "
+                "one of them has to be one-dimensional"
```
JonathanWenger, line 115: Super petty point: current style is that all error messages have a full stop.

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,236 @@
...
+        self,
+        dim: IntArgType,
+        domain: Tuple[Union[np.ndarray, FloatArgType], Union[np.ndarray, FloatArgType]],
+        random_state: Optional[RandomStateArgType] = None,
+        name: Optional[str] = "Custom measure",
+    ):
+
+        self._set_dimension_domain(dim, domain)
+        self.random_state = random_state
+        self.name = name
```
JonathanWenger, line 42: How is `name` used here? Would it be enough to only allow subclassing `IntegrationMeasure` and then `name` is obsolete since it is given through the class name?

JonathanWenger (changes requested):
First BQ PR :partying_face:. Looks great! I made some suggestions in the comments. Feel free to contact me directly with questions as well.

alpiges:
> Quick question, I'll take a closer look later this week. Is the branch up-to-date with the current master? That would explain the drop in test coverage caused by what seems like tests not being run for existing code.

I did (or at least I tried). But that would explain why suddenly files appear in the PR that I haven't touched. I'll make sure to fix this.

alpiges:
@pnkraemer 
> That is really cool! I am especially a fan of the embeddings -- as I've always been saying: I think it is an absolute BQ goldmine to have them somewhere public!
Glad you like them, it's a combinatorial playground!

> Since I have no idea what my role as `assignee` is, I will just ask random questions ;)
Whatever that means, but we want your feedback ;-)

> How do you feel about introducing a `QuadratureRule` interface? I know that you probably don't plan on doing anything other than BQ (are there even such options?), but for two reasons that could be a good idea:
> 
>     1. It could help making sure to stick to a common quadrature rule implementation scheme, with the result that a user that knows classical quadrature feels comfortable with BQ rules (a rather weak reason, but a reason nonetheless)
> 
>     2. We could use this quadrature rule interface for all sorts of approximate Gaussian filters in `filtsmooth`. There, we need non-Bayesian quadrature rules. While it is also a fun idea to use BQ there (people have been doing this, most notably Hans in one of his early papers), we need to be more general.
> 
Sounds reasonable to require the same structure for classical and probabilistic quadrature rules. From that perspective, sure, sounds good.
However, I think we need to separate the methods well enough to not cause confusion about whether a method is now probabilistic or not. How do you handle this with other classical numerical methods?

> 
> Related to 2.: we currently have the unscented transform in `filtsmooth`. Strictly speaking, that is more of a quadrature rule than a filter implementation. Shall we move this to quad eventually? No need to do it now, but what are your thoughts regarding non-Bayesian-but-PN-relevant quadrature rules in the `quad` package?

No strong opinion. If you think that's more of `quad` than `filtsmooth`, sure!

src/probnum/quad/_integration_measures.py
alpiges: Right. I think that is a remnant from when we accepted the choice of measure as `str` instead of `IntegratonMeasure` in `bayesquad`. Maybe we should make the integration measures public then, if the user needs to provide the measure when not integrating against the Lebesgue measure. What do you think?

src/probnum/quad/_integration_measures.py
alpiges: done

src/probnum/quad/_kernel_embeddings.py
```diff
@@ -0,0 +1,186 @@
...
+            self.measure.normalization_constant ** 2
+            * (2 * np.pi * ell ** 2) ** (self.dim / 2)
+            * np.atleast_2d(
+                ell * np.sqrt(2 / np.pi) * (np.exp(-(r ** 2) / (2 * ell ** 2)) - 1)
+                + r * scipy.special.erf(r / (ell * np.sqrt(2)))
+            ).prod()
+        )
+
+
+def get_kernel_embedding(kernel: Kernel, measure: IntegrationMeasure):
```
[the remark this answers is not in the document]
alpiges: I can see that it takes the role of a constructor, but is it a thing to instantiate child classes from parent classes? This feels wrong to me.

src/probnum/quad/_integration_measures.py
alpiges: by just using `asrandvar(scipy.stats.uniform(...))`? What's the benefit?

JonathanWenger: No particular benefit. Reading the comment I thought you were looking for a ProbNum implementation of a uniform random variable.

JonathanWenger:
> @pnkraemer
> 
> > That is really cool! I am especially a fan of the embeddings -- as I've always been saying: I think it is an absolute BQ goldmine to have them somewhere public!
> > Glad you like them, it's a combinatorial playground!
> 
> > Since I have no idea what my role as `assignee` is, I will just ask random questions ;)
> > Whatever that means, but we want your feedback ;-)
> 
> > How do you feel about introducing a `QuadratureRule` interface? I know that you probably don't plan on doing anything other than BQ (are there even such options?), but for two reasons that could be a good idea:
> > ```
> > 1. It could help making sure to stick to a common quadrature rule implementation scheme, with the result that a user that knows classical quadrature feels comfortable with BQ rules (a rather weak reason, but a reason nonetheless)
> > 
> > 2. We could use this quadrature rule interface for all sorts of approximate Gaussian filters in `filtsmooth`. There, we need non-Bayesian quadrature rules. While it is also a fun idea to use BQ there (people have been doing this, most notably Hans in one of his early papers), we need to be more general.
> > ```
> 
> Sounds reasonable to require the same structure for classical and probabilistic quadrature rules. From that perspective, sure, sounds good.
> However, I think we need to separate the methods well enough to not cause confusion about whether a method is now probabilistic or not. How do you handle this with other classical numerical methods?

I recommend sticking with just BQ in this PR. We can always refactor the interface later if we add classical quadrature. That would just bloat this PR unnecessarily. I am also not convinced we should add any classical methods to the package. Generally I am in favor of reusing classical BQ methods from other packages.

> > Related to 2.: we currently have the unscented transform in `filtsmooth`. Strictly speaking, that is more of a quadrature rule than a filter implementation. Shall we move this to quad eventually? No need to do it now, but what are your thoughts regarding non-Bayesian-but-PN-relevant quadrature rules in the `quad` package?
> 
> No strong opinion. If you think that's more of `quad` than `filtsmooth`, sure!

Maybe that's a discussion point for the software meeting.

src/probnum/quad/_integration_measures.py
JonathanWenger: I am fine with having `IntegrationMeasure` public.

src/probnum/quad/_kernel_embeddings.py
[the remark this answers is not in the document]
JonathanWenger: Very true. So i guess this is a factory method of sorts. Nvm my comment then.

tests/test_quad/test_kernel_embeddings.py
```diff
@@ -0,0 +1,90 @@
...
+def test_kvar_float(kernel_embedding):
+    """Test output of kernel variance."""
+    assert isinstance(kernel_embedding.kernel_variance(), np.float)
+
+
+def test_kmean(kernel_embedding, x):
+    """Test kernel mean against scipy.integrate.quad or sampling."""
+    if kernel_embedding.dim != 1:
+        # use MC sampling for ground truth
+        n_mc = 100000
```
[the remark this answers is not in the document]
alpiges: Oh yes, this is bullshit and unnecessary. Uncovered strange behaviour by the Cholesky decomposition though :sweat_smile:

tskarvone:
> > @pnkraemer
> > > That is really cool! I am especially a fan of the embeddings -- as I've always been saying: I think it is an absolute BQ goldmine to have them somewhere public!
> > > Glad you like them, it's a combinatorial playground!
> > 
> > 
> > > Since I have no idea what my role as `assignee` is, I will just ask random questions ;)
> > > Whatever that means, but we want your feedback ;-)
> > 
> > 
> > > How do you feel about introducing a `QuadratureRule` interface? I know that you probably don't plan on doing anything other than BQ (are there even such options?), but for two reasons that could be a good idea:
> > > ```
> > > 1. It could help making sure to stick to a common quadrature rule implementation scheme, with the result that a user that knows classical quadrature feels comfortable with BQ rules (a rather weak reason, but a reason nonetheless)
> > > 
> > > 2. We could use this quadrature rule interface for all sorts of approximate Gaussian filters in `filtsmooth`. There, we need non-Bayesian quadrature rules. While it is also a fun idea to use BQ there (people have been doing this, most notably Hans in one of his early papers), we need to be more general.
> > > ```
> > 
> > 
> > Sounds reasonable to require the same structure for classical and probabilistic quadrature rules. From that perspective, sure, sounds good.
> > However, I think we need to separate the methods well enough to not cause confusion about whether a method is now probabilistic or not. How do you handle this with other classical numerical methods?
> 
> I recommend sticking with just BQ in this PR. We can always refactor the interface later if we add classical quadrature. That would just bloat this PR unnecessarily. I am also not convinced we should add any classical methods to the package. Generally I am in favor of reusing classic quadrature methods from other packages.
> 
> > > Related to 2.: we currently have the unscented transform in `filtsmooth`. Strictly speaking, that is more of a quadrature rule than a filter implementation. Shall we move this to quad eventually? No need to do it now, but what are your thoughts regarding non-Bayesian-but-PN-relevant quadrature rules in the `quad` package?
> > 
> > 
> > No strong opinion. If you think that's more of `quad` than `filtsmooth`, sure!
> 
> Maybe that's a discussion point for the software meeting.

I agree that we shouldn't add an interface for classical quadratures yet. However, we implemented tensor-product versions of two classical Gaussian quadrature rules in order to speed up the tests for kernel embeddings. These are now in `test_kernel_embeddings.py` but it might make more sense to collect all such auxiliary (for now at least) quadrature rules to one logical place.

pnkraemer:
> Maybe that's a discussion point for the software meeting.

> I agree that we shouldn't add an interface for classical quadratures yet. However, we implemented tensor-product versions of two classical Gaussian quadrature rules in order to speed up the tests for kernel embeddings. These are now in test_kernel_embeddings.py but it might make more sense to collect all such auxiliary (for now at least) quadrature rules to one logical place.

Thanks for your feedback. I agree with both of you. My question was more of a mid-/long-term question than a "please do this now" thing (even though longterm, I think it will be a good idea to have an interface `QuadratureRule` or something -- we experienced this for the ODE solvers). Perhaps we can make a `nonprobabilistic` subpackage of `quad` or something like this, where we collect these things. Let's discuss this in the next software meeting.

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,267 @@
...
+        self._set_dimension_domain(dim, domain)
+        self.random_state = random_state
+
+    def __call__(self, points: Union[float, np.floating, np.ndarray]) -> np.ndarray:
+        """Evaluate the density function of the integration measure.
+
+        Parameters
+        ----------
+        points :
+            Input locations (n_points,) or (n_points, dim)
```
JonathanWenger, line 46: Shape hints in ProbNum are specified as `*shape=(n, n)* -- Input locations.` See also the developer guide on documenting code: https://probnum.readthedocs.io/en/latest/development/adding_to_the_api_documentation.html

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Returns
+        -------
+        density_evals :
+            Density evaluated at given locations, (n_points,)
```
JonathanWenger, line 51: Shape hint (see above)

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,267 @@
...
+    random_state :
+        Random state of the random variable corresponding to the integration measure.
+    """
+
+    def __init__(
+        self,
+        domain: Tuple[Union[np.ndarray, FloatArgType], Union[np.ndarray, FloatArgType]],
+        dim: Optional[IntArgType] = None,
+        normalized: Optional[bool] = False,
+        random_state: Optional[RandomStateArgType] = None,
```
JonathanWenger, line 164: The random state is only ever used in the `sample` method. I recommend making it an argument of `sample` such that `LebesgueMeasure` can be made immutable. This change will be made in `RandomVariable` in the future as well as discussed in one of the recent software meetings.

src/probnum/quad/_integration_measures.py
```diff
@@ -0,0 +1,267 @@
...
+    """Gaussian measure on Euclidean space with given mean and covariance.
+
+    If ``mean`` and ``cov`` are scalars but ``dim`` is larger than one, ``mean`` and
+    ``cov`` are extended to a constant vector and diagonal matrix, respectively,
+    of appropriate dimensions.
+
+    Parameters
+    ----------
+    mean :
+        Mean of the Gaussian measure, shape (dim,)
```
JonathanWenger, line 208: Shape hints (see above)

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -0,0 +1,125 @@
...
+from probnum.kernels import ExpQuad, Kernel
+from probnum.randvars import Normal
+
+from .._integration_measures import IntegrationMeasure
+from ..kernel_embeddings import KernelEmbedding
+from ..policies import sample_from_measure
+
+
+class BayesianQuadrature:
+    r"""A base class for Bayesian quadrature
```
JonathanWenger, line 17: Petty docstring comments:
- A base class for Bayesian quadrature.
- .. math:: F = \int_\Omega f(x) d \mu(x).
- The kernel ... full stop
- The policy ... full stop

src/probnum/quad/kernel_embeddings/_kernel_embedding.py
```diff
@@ -0,0 +1,103 @@
...
+        )
+
+    # pylint: disable=invalid-name
+    def kernel_mean(self, x: np.ndarray) -> np.ndarray:
+        """Kernel mean w.r.t. its first argument against the integration measure.
+
+        Parameters
+        ----------
+        x :
+            n_eval locations where to evaluate the kernel mean, shape (n_eval, dim)
```
JonathanWenger, line 55: Shape hint (see above).

src/probnum/quad/kernel_embeddings/_kernel_embedding.py
```diff
@@ -0,0 +1,103 @@
...
+
+        Returns
+        -------
+        k_mean :
+            The kernel integrated w.r.t. its first argument, evaluated at locations x,
```
JonathanWenger, line 60: Shape hint (see above).

src/probnum/quad/policies/sample_measure.py
```diff
@@ -0,0 +1,24 @@
...
+def sample_from_measure(nevals: int, measure: IntegrationMeasure) -> np.ndarray:
+    r"""Acquisition policy: random samples from the integration measure
+
+    Parameters
+    ----------
+    nevals : int
+        Number of function evaluations.
+
+    measure :
+            integration measure :math:`\mu`
```
JonathanWenger, line 17: Unnecessary indent.

JonathanWenger (changes requested):

JonathanWenger:
I am still quite confused where the drop in coverage comes from, but it seems to have originated from this commit: https://codecov.io/gh/probabilistic-numerics/probnum/commit/631c555cb2b822d24ed6134a5810ef6cd45278fb/

Clearly no files have changed that would affect the coverage, so there is either a problem with what Codecov is comparing to or somehow the commit history was changed by that commit (possibly a rebase??).

alpiges:
> Clearly no files have changed that would affect the coverage, so there is either a problem with what Codecov is comparing to or somehow the commit history was changed by that commit (possibly a rebase??).

Uh oh this is possible. I tried a rebase before I realized I can just pull the changes from upstream. But I thought I had aborted it and certainly never moved any commits around. Do you have an idea how to fix this?

src/probnum/quad/_integration_measures.py
alpiges: I agree, but with the current setup of `RandomVariable` I don't know how to pass `random_state` other than when initializing `RandomVariable`. `RandomVariable.sample` does not accept a `random_state`. I think this needs action in `RandomVariable` first.

alpiges:
Closing due to codecov issues. All comments have been applied. Everything has been moved to a new PR.

PR #365, opened 2021-03-24
Dense output sampling

pnkraemer:
This PR deals with more issues in sampling procedures. When the sampling grid is less dense than the evaluation grid, ugly stuff happens:

Screenshot from 2021-03-24 11-10-45

While at it (and because it is very related to the fix), dense output computation is vectorised quite drastically. There is also explicit treatment of extrapolation (not sure how well this was handled previously) and explicit exceptions when trying to sample or call left of the domain (the latter seems possible in principle, but I am not sure how reliable transition.forward_rv handles negative increments...)


pnkraemer:
Now things seems to be fixed again :) 
Screenshot from 2021-03-24 12-32-40

src/probnum/statespace/generate_samples.py
```diff
@@ -15,10 +18,6 @@ def generate_samples(dynmod, measmod, initrv, times):
         Random variable according to initial distribution
     times : np.ndarray, shape (n,)
         Timesteps on which the states are to be sampled.
```
schmidtjonathan, line 20: Do we want to add `random_state` to the parameter doc string?

schmidtjonathan (approved):
Thanks for taking care of this!

src/probnum/statespace/generate_samples.py
pnkraemer: Done

src/probnum/statespace/transition.py
```diff
@@ -350,6 +350,64 @@ def jointly_transform_base_measure_realization_list_backward(
...
+            curr_rv, _ = self.forward_realization(
+                curr_sample,
+                t=t[idx - 1],
+                dt=dt,
+                _linearise_at=_linearise_prediction_step_at,
+            )
+            curr_sample = (
+                curr_rv.mean
+                + curr_rv.cov_cholesky
+                @ base_measure_realizations[idx - 1].reshape((-1,))
```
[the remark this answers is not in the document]
pnkraemer: Done

PR #373, opened 2021-04-06
Benchmarks for filtering and smoothing

### LINKED ISSUE #354: Benchmarks for filtering
We need 'asv` benchmarks for filtering. Nothing too special, but the following would be nice:

* filtsmooth on a pendulum problem with e.g. square-root EKF (I think square-root should be default behaviour soon, because it is so much more stable)
* sampling from the Kalman posterior (this is slow currently and needs to be sped up!)
* bootstrap and gaussian-proposal particle filtering on a pendulum (and doing something with the posterior, e.g. resampling?) once #353 is merged

schmidtjonathan:
Closes #354


benchmarks/filtsmooth.py
```diff
@@ -0,0 +1,160 @@
+"""Benchmarks for Gaussian filtering."""
+import numpy as np
+
+import probnum.filtsmooth as pnfs
+import probnum.statespace as pnss
+from probnum.randvars import Normal
```
pnkraemer, line 6: can we use the correct imports here, please? Thanks :)

benchmarks/filtsmooth.py
```diff
@@ -0,0 +1,160 @@
...
+        + np.diag(np.array([delta_t ** 2 / 2]), 1)
+        + np.diag(np.array([delta_t ** 2 / 2]), -1)
+    )
+    r = var * np.eye(1)
+    initmean = np.ones(2)
+    initcov = var * np.eye(2)
+    dynamod = pnss.DiscreteGaussian(2, 2, f, lambda t: q, df)
+    measmod = pnss.DiscreteGaussian(2, 1, h, lambda t: r, dh)
+    initrv = Normal(initmean, initcov)
+    return dynamod, measmod, initrv, {"dt": delta_t, "tmax": 4}
```
pnkraemer, line 49: why is this copied here and not referenced?

benchmarks/filtsmooth.py
```diff
@@ -0,0 +1,160 @@
...
+
+class Filtering:
+    """Benchmark Kalman filter for different linearization techniques."""
+
+    param_names = ["linearization"]
+    params = [["ekf", "ukf"]]
+
+    def setup(self, linearization):
+        dynmod, measmod, initrv, info = load_pendulum()
+        _lin_method = {
```
pnkraemer, line 60: I am curious: is there any reason for the leading underscore? And I, personally, like to split the dict containing options from the actual choosing, i.e. `choose_lin_method = {...}; _lin_method = choose_lin_method[linearization]`. This is super minor, but perhaps your method is better... :)

benchmarks/filtsmooth.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def time_filter(self, linearization):
+        self.kalman_filter.filter(dataset=self.observations, times=self.locations)
+
+    def peakmem_filter(self, linearization):
+        self.kalman_filter.filter(dataset=self.observations, times=self.locations)
+
+
+class Smoothing:
+    """Benchmark Kalman filter for different linearization techniques."""
```
pnkraemer, line 87: benchmark kalman smoother?

benchmarks/filtsmooth.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+        self.smoothing_posterior = self.kalman_filter.smooth(
+            filter_posterior=filtering_posterior
+        )
+
+    def time_sample(self, linearization, num_samples):
+        self.smoothing_posterior.sample(size=num_samples)
+
+    def peakmem_sample(self, linearization, num_samples):
+        self.smoothing_posterior.sample(size=num_samples)
```
pnkraemer, line 160: Do you think it would make the sampling benchmark more meaningful if the sampling grid was more dense than the filtering grid?

pnkraemer (commented):
Thanks for taking care of this! I have only minor comments and one request:

can we also benchmark continuous-time interpolation, aka dense output? It should work simialrly to the sampling benchmark, only that here, the dense output grid should be more dense than the sampling grid.

benchmarks/filtsmooth.py
schmidtjonathan: Because it is not part of our problem zoo. Shall I use a different problem?

benchmarks/filtsmooth.py
schmidtjonathan: No, there is no reason, other than to mark this variable as not relevant _as_ a variable. It merely serves as a `switch ... case ...` substitute.

schmidtjonathan:
@pnkraemer I now gathered the filtering inter- / extrapolation , smoothing inter-/ extrapolation and the dense sampling in one `DenseGridOperation` benchmark (feel free to criticize the naming 😄 ).
Is that ok?

benchmarks/filtsmooth.py
pnkraemer: so as soon as we move this to the zoo (which would happen soon, right?), it would be referenced here?

benchmarks/filtsmooth.py
pnkraemer: hmmm that would make it a relevant variable, no? :D I think that this is very subjective, though, so feel free to ignore this comment.

benchmarks/filtsmooth.py
schmidtjonathan: I think so, yes. So far, we only use the problem in the `test`s. Do you want me to add this to the problem zoo within this PR? Would be no problem 👍

benchmarks/filtsmooth.py
schmidtjonathan: I find it ugly as well, and I see no point in keeping it this way. Seemed right at the time, but I'll get rid of it 😄

pnkraemer:
> EDIT: I fear that the benchmarks became a bit lengthy by adding 50% of the time interval as an extrapolation grid and dividing the step size by 4.
I dont think it is necessary to go crazy here, 10-15% as extrapolation and diving the step by 2 would suffice IMO (another option would be to draw a number of random locations in between t0 and tmax + 10% -- this would cap the complexity as well) :).

benchmarks/filtsmooth.py
pnkraemer: we dont have a filtsmooth zoo, right? In this case it would be great if you took care of it in a different PR? What do you think?

schmidtjonathan: Yep, sure, will do :)

schmidtjonathan:
> > EDIT: I fear that the benchmarks became a bit lengthy by adding 50% of the time interval as an extrapolation grid and dividing the step size by 4.
> 
> I dont think it is necessary to go crazy here, 10-15% as extrapolation and diving the step by 2 would suffice IMO (another option would be to draw a number of random locations in between t0 and tmax + 10% -- this would cap the complexity as well) :).

Do you think that sorting the randomly drawn grid might not be sufficiently stable?

pnkraemer:
In which sense not sufficiently stable? Sorting sounds like a good idea!

schmidtjonathan:
> In which sense not sufficiently stable? Sorting sounds like a good idea!

The (comparison-based) sorting algorithm worries me a bit in the context of random numbers. Might be unreasonable though.

pnkraemer:
Hmmm how about the following? Let `posterior.locations` be (0., 1., ..., 10.). Then, we can go in 10 steps: (0, 1.1, 2.2, ..., 11) all of which lie between grid points, but the complexity is the same and deterministic?

schmidtjonathan:
@pnkraemer I just had a chat with @marvinpfoertner , and the conclusion was that it is safe. But I added a `np.unique` such that we don't get a `dt=0`. So imo, it's ready for review.

pnkraemer (approved):

Issue #258, opened 2020-11-20
Square-root implementation of continuous-time filtering and smoothing


pnkraemer:
As of now, `statespace.linear_sde_statistics` uses RK4 for numerically integrating the covariance equation in `linear_sde_statistics()` (a so-called continuous matrix-riccati equation).

We can probably do much better by using a method that preserves symmetry and positive definiteness at each step. If we cannot guarantee these qualities for the result, the resulting covariance will suffer badly from round-off errors and smoothing iterations may break (we've been through this in a different place before...).
One such method is https://webspace.science.uu.nl/~frank011/Articles/FrZh14.pdf, using a Möbius transformation, though there may be others.

It is not urgent at the moment but will be something to keep in mind. Maybe it is sufficient to switch to either a simple implicit method or a simple symplectic method. I dont know, but I guess we will run into this problem in the near future.

pnkraemer:
https://ieeexplore.ieee.org/document/4045974 this is the solution (square-root implementation for solving the riccati equation)

pnkraemer:
The respective method is now `LinearSDE._solve_mde_*`.

PR #374, opened 2021-04-07
square-root implementation of continuous-time filtering

jzenn:
This adds a square-root implementation of continuous-time filtering to `LinearSDE` as described in Issue #258.

Some comments:

1. ~~I use `scipy.linalg.lu_factor` and `scipy.linalg.lu_solve` instead of two times `np.linalg.solve` or `np.linalg.inv` one time.~~
2. The covariance matrix of a random variable needs to be positive definite for the sqrt-algorithm to work. Currently there is no error-handling for a non-positive definite matrix.
3. There is no docstring for `forward_implementation` which I adapted from `DiscreteLinearGaussian`.
4. To avoid duplication I refactored `scipy.integrate.solve_ivp` into `_solve_mde_forward`.


pnkraemer:
Hmmm, somehow this PR slipped past me, but it looks pretty ready! 

Let me know when you want a review please :)

jzenn:
You can review the PR if you like :)

src/probnum/statespace/sde.py
```diff
@@ -298,6 +353,45 @@ def f(t, y):
...
+
+            # Vectorize outcome
+            new_cov_sqrt_flat = new_cov_sqrt.flatten()
+            y_new = np.hstack((new_mean, new_cov_sqrt_flat))
+            return y_new
+
+        initcov_sqrt_flat = initrv.cov_cholesky.flatten()
+        y0 = np.hstack((initrv.mean, initcov_sqrt_flat))
+
+        return f, y0
```
pnkraemer, line 393: Do you think it would be possible to document this a bit more extensively than linking the paper? It is paywall-protected, so not everyone can see the equation easily. One way would be to write down the actual equation (in a maths way) that is solved here. That would help me when reading the code :)

src/probnum/statespace/sde.py
```diff
@@ -298,6 +353,45 @@ def f(t, y):
...
+            F = self.driftmatfun(t)
+            u = self.forcevecfun(t)
+            L = self.dispmatfun(t)
+
+            new_mean = F @ mean + u
+            cov_sqrt_lu, cov_sqrt_piv = scipy.linalg.lu_factor(cov_sqrt)
+            F_bar = scipy.linalg.lu_solve((cov_sqrt_lu, cov_sqrt_piv), F @ cov_sqrt)
+            L_bar = np.sqrt(_diffusion) * scipy.linalg.lu_solve(
+                (cov_sqrt_lu, cov_sqrt_piv), L
+            )
```
pnkraemer, line 380: isnt this sqrt a cholesky factor? if so, cho_solve would not only be faster but also more stable numerically

tests/test_statespace/test_sde.py
```diff
@@ -232,6 +244,28 @@ def test_solve_mde_forward_values(ltisde_as_linearsde, ltisde, v_const, diffusio
...
+
+    out_linear_2, _ = ltisde_as_linearsde.forward_rv(
+        out_linear, t=0.1, dt=0.1, _diffusion=diffusion
+    )
+    out_linear_2_sqrt, _ = ltisde_as_linearsde_sqrt_forward_implementation.forward_rv(
+        out_linear, t=0.1, dt=0.1, _diffusion=diffusion
+    )
+
+    np.testing.assert_allclose(out_linear_2_sqrt.mean, out_linear_2.mean)
+    np.testing.assert_allclose(out_linear_2_sqrt.cov, out_linear_2.cov)
```
pnkraemer, line 266: Please correct me if I am wrong. The sqrt-forward implementation solves a different ODE than the non-square root implementation, both use an adaptive solver. So is it actually meaningful to check that the outputs are identical up to numerical precision? It is already tested whether they coincide with the LTI transition, right? 

I am not sure whether I am asking a question or proposing anything. But do you think that it would be a meaningful layer of testing to do the present test here on a non-lti sde?

src/probnum/statespace/sde.py
```diff
@@ -231,9 +235,58 @@ def _solve_mde_forward(self, rv, t, dt, _diffusion=1.0):
...
+            rv,
+            _diffusion=_diffusion,
+        )
+
+        sol = self._solve_mde_forward(mde, y0, t, dt)
+
+        dim = rv.mean.shape[0]
+        y_end = sol.y[:, -1]
+        new_mean = y_end[:dim]
+        new_cov_sqrt = y_end[dim:].reshape((dim, dim))
```
pnkraemer, line 251: this block of code exists for both forward implementations, right? Do you think it is worth trying to factor this into a common implementation, or do you reckon it is wasted efforts?

pnkraemer (commented):
Thanks a lot for addressing this! I left some comments in the code diff, nothing major. :)

src/probnum/statespace/sde.py
jzenn: I thought about this but there would be a naming issue since once we get the square-root covariance and once the covariance (in 251). If that's ok for you I can change it.

src/probnum/statespace/sde.py
jzenn: I can do that.

src/probnum/statespace/sde.py
pnkraemer: Hmmm I dont know. I am a bit worried about all the flattening and slicing that is happening here, so in case some error comes up, it would be good to only have to fix it in one place. I see the naming though, and perhaps it is actually more readable if it is *not* factored out. As long as it only exists, like, twice, I am fine with a bit of duplication. 

Long story short, I dont know -- your call!

jzenn: Would something like `new_cov_or_cov_sqrt` resolve the naming issue?

pnkraemer: I think it would help, especially with an accompanying comment that explains why the `or`. Not sure whether it entirely resolves the issue though...

pnkraemer:
Can you also add the `forward_implementation` argument to `ContinuousEKFComponent` please? By implementing forward and backward transitions for `LinearSDE` objects, they are immediately available for continuous-time extended Kalman filtering (ContinuousUKF does not work quite the same with the current code, unfortunately). 

A guide is the respective choice of implementation in `DiscreteEKFComponent`. 

Thanks! :)

src/probnum/statespace/sde.py
jzenn: I initially thought that we cannot rely on `cov_sqrt` being positive definite but you're right it is initialised using `initrv.cov_cholesky`.

tests/test_statespace/test_sde.py
jzenn: This should test the forward values being the same for both implementations. I thought for the simple LTI case they should coincide. Nevertheless, I think testing on a non-LTI SDE might be a good idea.

jzenn:
> Can you also add the `forward_implementation` argument to `ContinuousEKFComponent` please? By implementing forward and backward transitions for `LinearSDE` objects, they are immediately available for continuous-time extended Kalman filtering (ContinuousUKF does not work quite the same with the current code, unfortunately).

I haven't thought about this. I will add `forward_implementation`.

tests/test_statespace/test_sde.py
jzenn: > It is already tested whether they coincide with the LTI transition, right?
If I understand you correctly, it is already tested whether the classic implementation coincides with the LTI transition but not whether the square-root implementation coincides with the LTI transition.

pnkraemer: Ah I see. In this case nevermind my comment :) Thanks!

src/probnum/statespace/sde.py
jzenn: I just looked into this again realizing that I did not use a Cholesky decomposition since `cov_sqrt` is not symmetric. Sorry for the confusion!

jzenn:
I just realized that `driftmatfun` is called `G(t)` in the docstring while the `_setup_*` methods call it `F(t)`. I will change that to be consistent.

src/probnum/statespace/sde.py
```diff
@@ -298,6 +350,86 @@ def f(t, y):
...
+
+            \bar G(t) = S^{-1}(t)G(t)S(t),
+            \bar L(t) = S^{-1}L(t)
+
+        and
+
+            \dot S(t) = S(t)[M(t)]_{lt}
+
+        where lt denotes the lower-triangular operator defined by
+
```
pnkraemer, line 393: wow, thanks a lot for this! Do you think it would help to put these maths environments into a `.. math:: \dot S(...` environment? This would make sphinx render this docstring well (which currently is not needed, because this function is not visible). I dont know whether it would be worth the effort. What do you reckon?

src/probnum/statespace/sde.py
```diff
@@ -298,6 +350,86 @@ def f(t, y):
...
+            cov_sqrt = cov_sqrt_flat.reshape((dim, dim))
+
+            # Apply iteration
+            G = self.driftmatfun(t)
+            u = self.forcevecfun(t)
+            L = self.dispmatfun(t)
+
+            new_mean = G @ mean + u
+            cov_sqrt_lu, cov_sqrt_piv = scipy.linalg.lu_factor(cov_sqrt)
+            G_bar = scipy.linalg.lu_solve((cov_sqrt_lu, cov_sqrt_piv), G @ cov_sqrt)
```
pnkraemer, line 415: the maths above would suggest to me that cov_sqrt ("S") is triangular (because of this choice of $M$, or am I misunderstanding?

src/probnum/statespace/sde.py
```diff
@@ -199,12 +210,57 @@ def backward_rv(
...
+        # Useful for backward transitions
+        # Aka continuous time smoothing.
+        sol_mean = lambda t: sol.sol(t)[:dim]
+        sol_cov_sqrt = lambda t: sol.sol(t)[dim:].reshape((dim, dim))
+        sol_cov = (
+            lambda t: sol.sol(t)[dim:].reshape((dim, dim))
+            @ sol.sol(t)[dim:].reshape((dim, dim)).T
+        )
+
+        return randvars.Normal(mean=new_mean, cov=new_cov), {
```
pnkraemer, line 255: I think one can also initialise the `cov_cholesky` field of `Normal` with `new_cov_sqrt` (after all, that was the point here, right?). Or am I misunderstanding?

pnkraemer (commented):
Thanks a lot for your efforts johannes! I had some questions regarding the cov_sqrt (and whether it is a Cholesky factor, because if it is, we may want to use this at some places). I left those in the code diff :)

src/probnum/statespace/sde.py
jzenn: @pnkraemer here

src/probnum/statespace/sde.py
jzenn: Yes you're right. I left a comment above in the last review explaining this.

src/probnum/statespace/sde.py
jzenn: I can do that

src/probnum/statespace/sde.py
jzenn: Yes, I totally missed this.

src/probnum/statespace/sde.py
jzenn: I will change that to `scipy.linalg.solve_triangular`.

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
```diff
@@ -136,6 +136,7 @@ def __init__(
         mde_atol=1e-5,
         mde_rtol=1e-5,
         mde_solver="LSODA",
+        forward_implementation="classic",
```
pnkraemer, line 139: thanks for doing this! Is this argument in the docstring?

src/probnum/statespace/sde.py
```diff
@@ -298,6 +350,97 @@ def f(t, y):
 
         return f, y0
 
+    def _setup_vectorized_mde_forward_sqrt(self, initrv, _diffusion=1.0):
+        r"""Set up forward moment differential equations (MDEs) using a square-root
+        implementation. (https://ieeexplore.ieee.org/document/4045974)
+
+        The covariance :math:`P(t)` obeys the Ricatti equation
```
pnkraemer, line 357: riccati, not ricatti (right?)

src/probnum/statespace/sde.py
```diff
@@ -199,12 +210,57 @@ def backward_rv(
...
+            _diffusion=_diffusion,
+        )
+
+        sol, new_mean, new_cov_sqrt = self._solve_mde_forward(mde, y0, t, dt, dim)
+        new_cov = new_cov_sqrt @ new_cov_sqrt.T
+
+        # Useful for backward transitions
+        # Aka continuous time smoothing.
+        sol_mean = lambda t: sol.sol(t)[:dim]
+        sol_cov_sqrt = lambda t: sol.sol(t)[dim:].reshape((dim, dim))
```
pnkraemer, line 249: would be great (if this is a cholesky factor) to call this `sol_cov_cholesky` (in line with how we refer to this variable in e.g. `Normal`).

pnkraemer: this applies to all the code in this module :)

src/probnum/statespace/sde.py
```diff
@@ -215,25 +271,19 @@ def _solve_mde_forward(self, rv, t, dt, _diffusion=1.0):
             rtol=self.mde_rtol,
             dense_output=True,
         )
-        dim = rv.mean.shape[0]
         y_end = sol.y[:, -1]
         new_mean = y_end[:dim]
-        new_cov = y_end[dim:].reshape((dim, dim))
+        # If forward_sqrt is used, new_cov_or_cov_sqrt is the square-root of the covariance
+        # If forward_classic is used, new_cov_or_cov_sqrt is the covariance
```
pnkraemer, line 277: thanks for these explanations :)

src/probnum/statespace/sde.py
```diff
@@ -298,6 +350,97 @@ def f(t, y):
...
+
+        Compute an ODE vector field that represents the MDEs and is
+        compatible with scipy.solve_ivp.
+        """
+        dim = len(initrv.mean)
+
+        def f(t, y):
+            # Undo vectorization
+            mean, cov_sqrt_flat = y[:dim], y[dim:]
+            cov_sqrt = cov_sqrt_flat.reshape((dim, dim))
```
pnkraemer, line 418: `cov_cholesky` would be better to have here, because it conveys more information than `cov_sqrt`

pnkraemer (commented):
Thank you so much! I am sorry that there have been so many iterations with this PR, and that it has been pending for a while now. Once my final comments (all are super minor) are resolved, let's merge it!

src/probnum/statespace/sde.py
jzenn: Yes, you're right :)

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
jzenn: No, it also is not for `LinearSDE` and `DiscreteLinearGaussian` (see also 3. in https://github.com/probabilistic-numerics/probnum/pull/374#issue-610816053).

src/probnum/statespace/sde.py
jzenn: >if this is a cholesky factor

The paper does not explicitly state anything regarding `cov_sqrt` being a Cholesky factor. I think `cov_sqrt` should be a Cholesky factor but I am not certain.

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
pnkraemer: Hm okay, I can live with that :)

src/probnum/statespace/sde.py
pnkraemer: I think it should (because of the way it is truncated after each step), but you know more about this than I do :) One simple way of finding out would be to run the tests with a simple `assert np.testing.assert_allclose(cov_sqrt, np.tril(cov_sqrt))` and if it passes, it should be a cholesky factor. If so, then renaming these variables would be wonderful, and if not, leave it as is (I am truly sorry to be so petty here with respect to the naming, but this information whether it is lower triangular or not is fairly crucial IMO -- after this we can merge, I promise!! :) )

jzenn: >One simple way of finding out would be to run the tests with a simple assert np.testing.assert_allclose(cov_sqrt, np.tril(cov_sqrt)) and if it passes, it should be a cholesky factor

Yes, this is true by definition. However, I don't see that `cov_sqrt_{ii} > 0` which does need to be true for a Cholesky factor, right?

src/probnum/filtsmooth/gaussfiltsmooth/extendedkalman.py
jzenn: I can introduce the argument in the docstring if you'd feel more comfortable with that :)

src/probnum/statespace/sde.py
pnkraemer: Ah I see. You can transform the factor into a valid cholesky factor though: `S = L1 L1t = L2 Q Qt L2t` and if `Qt` is the (orthogonal) matrix consisting of 1 and -1 on the diagonals, then (in this context here) we can rather loosely switch back between L1 and L2 in order to transform the lower-triangular matrix-sqrt into a Cholesky factor (its diagonals need to be != 0, but that should be no problem here). This is what is done in `probnum.utils.linalg.cholesky_update`, and might be applicable here. What do you think?

jzenn: Oh I see. I haven't thought on that end. My tests suggested that the diagonal entries are indeed positive but I did not see a way to proof this following the algorithm of the paper. I will look into `probnum.utils.linalg.cholesky_update` and report back if that is ok.

jzenn:
One more question regarding `_cholesky` and `_sqrt`: Would you recommend renaming the methods as well (and the class parameter)?

pnkraemer:
no need, the `_sqrt`, `_classic` etc. endings are used everywhere around here, and are arguably easier to parse (the algorithm is called square-root kalman filter) :)

src/probnum/statespace/sde.py
pnkraemer: I see that you re-requested my review -- may I ask what the result here was? :)

jzenn: I am using `probnum.utils.linalg.cholesky_update` to ensure that `cov_cholesky` indeed is a Cholesky factor as you suggested. However, I did not come up with an example where the algorithm produces a triangular matrix with non-positive diagonal entries.

jzenn:
Is there anything else I can do here @pnkraemer?

src/probnum/statespace/sde.py
```diff
@@ -298,6 +353,101 @@ def f(t, y):
...
+            new_mean = G @ mean + u
+            G_bar = scipy.linalg.solve_triangular(
+                cov_cholesky, G @ cov_cholesky, lower=True
+            )
+            L_bar = np.sqrt(_diffusion) * scipy.linalg.solve_triangular(
+                cov_cholesky, L, lower=True
+            )
+            M = G_bar + G_bar.T + L_bar @ L_bar.T
+
+            new_cov_cholesky = cholesky_update(
```
pnkraemer, line 437: I think you meant `utils.linalg.tril_to_positive_tril` here?!

pnkraemer (commented):
Hey, sorry for the delay again. I think the call to Cholesky update is mixed up (it should rather be `tril_to_positive_tril`, and the branch is not up to date anymore. I promise to approve when these two things are resolved!

src/probnum/statespace/sde.py
jzenn: I think this should result in the same matrices, right? But I see that using `utils.linalg.tril_to_positive_tril` might be more descriptive of this use case (and we'd save the call to `np.linalg.qr`).

pnkraemer: yes, the call to qr is the things that should be avoided here. QR decompositions are expensive :)

pnkraemer (approved):
Thanks a lot for your patience johannes!

PR #376, opened 2021-04-14
Perturbation functions + testing

NinaEffenberger:
perturbation functions for the perturbed solvers as proposed by Abdulle and Garegnani (2020)


src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,50 @@
+"""perturbation functions to perturb the stepsize."""
```
pnkraemer, line 1: Super minor: """Perturbation ...""" (capitalisation)

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,50 @@
...
+import numpy as np
+
+
+def perturb_uniform(step, order, noise_scale, seed=1):
```
pnkraemer, line 5: 1. Instead of `seed=1`, we use `random_state=None` in probnum. This random state is then passed directly to a scipy sampling function, in this case `scipy.stats.uniform.rvs(random_state=random_state)`. Would be great if you could update this :)
2. can you maybe change "order" to "solver_order" or something? there are so many "order"s floating around probnum.diffeq, that it surely helps to be a bit more verbose here.

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,50 @@
...
+    order : int
+        order of the solver
+    noise_scale : float
+        scales the perturbation
+    seed : int
+        seed for pseudo-random number generator.
+    """
+    np.random.seed(seed=seed)
+    if step < 1:
+        noisy_step = np.random.uniform(
```
pnkraemer, line 22: see my random state and sampling comment above. This is a more flexible way than setting np.random.seed to a fixed value.

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,50 @@
...
+            step - noise_scale * step ** (order + 0.5),
+            step + noise_scale * step ** (order + 0.5),
+        )
+    else:
+        print("Error: Stepsize too large (>=1), not possible")
```
pnkraemer, line 27: Instead of printing "error", I think it would be better to either raise a "ValueError" or at least throw a "real" warning (https://docs.python.org/3/library/warnings.html). Since the implementation actually breaks here, I would prefer the ValueError though.

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,50 @@
+"""perturbation functions to perturb the stepsize."""
+import numpy as np
+
+
+def perturb_uniform(step, order, noise_scale, seed=1):
+    """perturbs the step with uniformly distributed noise scaled by noise-scale.
+    proposed by Abdulle and Garegnani(2020)
```
pnkraemer, line 7: docstrings are usually written in imperative: "Perturb the step with uniformly...". 

Regarding the reference: there is an official sphinx environment thingy for "Reference", which may be useful here? The docstrings for "probsolve_ivp" or "problinsolve" would be a good starting point to learn how (well, at least this is where I look it up because I forget it all the time)

tests/test_diffeq/test_perturbedsolvers/test_perturbation_functions.py
```diff
@@ -0,0 +1,56 @@
...
+from probnum.diffeq.perturbedsolvers import _perturbation_functions
+
+random_state = np.random.mtrand.RandomState(seed=1)
+
+
+def test_perturb_uniform():
+    random_state = np.random.mtrand.RandomState(seed=1)
+    step = 0.2
+    noise_scale = 1
+    order = 4
```
pnkraemer, line 12: the parameters "step, noise_scale, order" would make good fixtures (https://docs.pytest.org/en/stable/fixture.html). If you dont know what I mean, let me know offline :)

tests/test_diffeq/test_perturbedsolvers/test_perturbation_functions.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+
+from probnum.diffeq.perturbedsolvers import _perturbation_functions
+
+random_state = np.random.mtrand.RandomState(seed=1)
+
+
+def test_perturb_uniform():
```
pnkraemer, line 8: it seems to me that the test checks whether the output is identical once the seed is fixed, right? I am not sure whether this is the best possible test here (it might actually be too close to the implementation of the test function, but this is only a hunch, not a clear statement)... 

Can we maybe also have some sort of test that checks the assumptions of the Abdulle/Garegnani paper (i.e. E(h_k) = h, Var())? That would make a good test across perturbation functions as well.

pnkraemer: Oh i see that you do this for the mean below. One for the variance would be good, too (because this is also an important part of the assumption!)

tests/test_diffeq/test_perturbedsolvers/test_perturbation_functions.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def test_perturb_lognormal_avrg():
+    num_samples = 100
+    all_steps = 0
+    unperturbed_step = 0.2
+    for _ in range(num_samples):
+        stp = _perturbation_functions.perturb_lognormal(unperturbed_step, 4, 1)
+        all_steps += stp
+    mean_suggested_step = all_steps / num_samples
```
pnkraemer, line 53: this would be much easier and cheaper to test if the perturb_lognormal had a "size" argument that would return multiple samples. Do you think that would be a good idea? This size argument could have a default as `size=()` but makes testing easier (and exposes all parameters that would be passed to the scipy sampler anyway)

tests/test_diffeq/test_perturbedsolvers/test_perturbation_functions.py
```diff
@@ -0,0 +1,56 @@
...
+    np.testing.assert_allclose(
+        mean_suggested_step, unperturbed_step, atol=1e-4, rtol=1e-4
```
pnkraemer, line 55: how flaky is this test? It seems to me as if 100 samples dont always give 1e-4 error... but you know this test much better than me :)

pnkraemer (commented):
Hey, thanks for getting started here! Overall it looks good. I left a few comments with explanations how to make the code more similar to existing Probnum code, all of which should be easy fixes. If anything is unclear, ping me offline :)

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,58 @@
+"""perturbation functions to perturb the stepsize.
```
pnkraemer, line 1: "Perturbation"...

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,58 @@
...
+----------
+.. [1] https://arxiv.org/abs/1801.01340
+"""
+import numpy as np
+import scipy
+
+
+def perturb_uniform(step, solver_order, noise_scale, random_state=None, size=()):
+    """Perturb the step with uniformly distributed noise scaled by noise-scale [1]_.
+    proposed by Abdulle and Garegnani(2020)
```
pnkraemer, line 13: Can you use a line break before "proposed..." please? This is usual docstring style (see for instance the docstring style used by docformatter https://pypi.org/project/docformatter/).

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,58 @@
...
+
+    Parameters
+    ----------
+    step : float
+        unperturbed step propesed by the steprule
+    solver_order : int
+        order of the solver
+    noise_scale : float
+        scales the perturbation
+    seed : int
```
pnkraemer, line 23: this is not an integer, but a `RandomStateArgType` (see `probnum.type`).

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,58 @@
...
+        seed for pseudo-random number generator.
+    """
+    scipy.stats.uniform.rvs(random_state=random_state)
```
pnkraemer, line 26: What I meant with me comment in the previous review, the code in this function should be along the lines of
1) `if step >= 1.: raise ValueError("Stepsize too large (>= 1)")`, which raises the error correctly
2) `uniform_rv_samples = scipy.stats.uniform.rvs(random_state=random_state, size=size)`  which computs standard (0,1) uniform samples
3) `scaled_uniform_rv_samples = left_boundary + (right_boundary - left_boundary)*uniform_rv_samples`, where `left_boundary=step - noise_scale * step...` and `right_boundary=step + noise_scale...`. This moves the uniform samples to the right domain

This is different to your implementation in that it properly handles `size`, `random_state`, and `step >= 1.`. Sorry if my comment in the previous review was not sufficiently accurate, I should've been more clear with the `scipy.stats.uniform` mention. If there are any questions, please let me know, okay? :)

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
```diff
@@ -0,0 +1,58 @@
...
+    mean = np.log(step) - np.log(
+        np.sqrt(1 + noise_scale * (step ** (2 * solver_order)))
+    )
+    cov = np.log(1 + noise_scale * (step ** (2 * solver_order)))
+    noisy_step = np.exp(np.random.normal(mean, cov, size))
+    return noisy_step
```
pnkraemer, line 58: Same comment as the one above (in terms of handling size, random_state -- again, sorry that I was unclear last time).
For instance, I'd suggest the implementation
```
shift =0.5 *  np.log(1 + noise_scale * (step ** (2*solver_order)))
mean = np.log(step) - shift
cov = 2*shift
samples = scipy.stats.multivariate_normal.rvs(mean=mean, cov=cov, size=size, random_state=random_state)
return samples
```
which handles `random_state` and `size` properly, and computs the sqrt/log thing only once :) Again, please let me know if there are any questions! :)

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
pnkraemer: it might also be good to compute the shift `noise_scale * step**(solver_order + 0.5)` only once (for simplicity and symmetry, and to a probably neglectible amount: speed).

tests/test_diffeq/test_perturbedsolvers/test_perturbation_functions.py
```diff
@@ -0,0 +1,60 @@
...
+    mean_suggested_step = np.sum(suggested_steps) / num_samples
+    np.testing.assert_allclose(mean_suggested_step, step, atol=1e-4, rtol=1e-4)
+
+
+def test_perturb_lognormal_mean(step, solver_order, noise_scale, num_samples):
+    suggested_steps = _perturbation_functions.perturb_lognormal(
+        step, solver_order, noise_scale, random_state=1, size=num_samples
+    )
+    mean_suggested_step = np.sum(suggested_steps) / num_samples
+    np.testing.assert_allclose(mean_suggested_step, step, atol=1e-6, rtol=1e-6)
```
pnkraemer, line 42: these tests are identical, so you could write the test once (and make the `perturb_x` functions into inputs that are then parameters of the test). This would make it easier to run the tests for more perturbation functions :)

tests/test_diffeq/test_perturbedsolvers/test_perturbation_functions.py
```diff
@@ -0,0 +1,60 @@
...
+    np.testing.assert_allclose(expected_var, var, atol=1e-4, rtol=1e-4)
+
+
+def test_perturb_lognormal_var(step, solver_order, noise_scale, num_samples):
+    expected_var = (1 / 3) * step ** (2 * solver_order + 1)
+    suggested_steps = _perturbation_functions.perturb_uniform(
+        step, solver_order, noise_scale, random_state=1, size=num_samples
+    )
+    var = ((suggested_steps - step) ** 2) / num_samples
+    np.testing.assert_allclose(expected_var, var, atol=1e-6, rtol=1e-6)
```
pnkraemer, line 60: same here: if `perturb_x` is an input of the test function, this test needs only be written once, but is called for both perturbations.

pnkraemer (commented):
Thanks a lot for all of the updates Nina! I have another round of comments, mostly clarifications of what I meant last time. I tried to make code suggestions wherever possible, please let me know if that does not help. When these are incorporated, this PR should be ready to merge :)

src/probnum/diffeq/perturbedsolvers/_perturbation_functions.py
pnkraemer: Sorry, super minor, but "Proposed by ..."?

pnkraemer (approved):
Thanks for your patience Nina! :)

Issue #346, opened 2021-03-11
Start/stop versus t/dt


pnkraemer:
the Transition API uses `t`, `dt` as interface. I think the same would make sense for the filters `FiltSmooth.filter_step` and ODE solvers (`ODESolver.step()`) which still use `start, stop`.

PR #379, opened 2021-04-20
Move filtsmooth problems to new filtsmooth problem zoo

schmidtjonathan:


src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
+import numpy as np
+
+from probnum import diffeq, filtsmooth, problems, randvars, statespace
+
+__all__ = ["car_tracking", "logistic_ode", "ornstein_uhlenbeck", "pendulum"]
+
+
+def car_tracking(delta_t=0.2, measvar=0.5, t_max=20.0):
+    r"""State space model for a simple car-tracking scenario.
```
pnkraemer, line 9: this got me thinking... should this actually be a state space model zoo, not a filtsmooth model zoo?

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+        2013.
+
+    """
+
+    dynamat = np.eye(4) + delta_t * np.diag(np.ones(2), 2)
+    dynadiff = (
+        np.diag(np.array([delta_t ** 3 / 3, delta_t ** 3 / 3, delta_t, delta_t]))
+        + np.diag(np.array([delta_t ** 2 / 2, delta_t ** 2 / 2]), 2)
+        + np.diag(np.array([delta_t ** 2 / 2, delta_t ** 2 / 2]), -2)
+    )
```
pnkraemer, line 70: would be great if these could be replaced by an `IBM` object. Then, we could also expose `ordint` (with a default set to what is happening here)

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+    initrv = randvars.Normal(mean, cov)
+
+    # Generate data
+    times = np.arange(0.0, t_max, step=delta_t)
+    states, obs = statespace.generate_samples(
+        dynmod=dynmod, measmod=measmod, initrv=initrv, times=times
+    )
+    regression_problem = problems.RegressionProblem(
+        observations=obs, locations=times, solution=states
```
pnkraemer, line 92: aah I always thought of `solution` as a callable. But on a regression problem, it makes sense that this is the `true` states at the grid. Right? 

Is this in the docs of `RegressionProblem` somewhere? If not, how about putting it there? There surely will be this kind of confusion again in the future.

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+    )
+    return dynmod, measmod, initrv, regression_problem
```
pnkraemer, line 94: super minor and petty: do you think this order of returns is optimal or should `regression_problem` be returned first?

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+    cov = 0.5 * measvar * np.eye(4)
+
+    dynmod = statespace.DiscreteLTIGaussian(
+        state_trans_mat=dynamat, shift_vec=np.zeros(4), proc_noise_cov_mat=dynadiff
+    )
+    measmod = statespace.DiscreteLTIGaussian(
+        state_trans_mat=measmat,
+        shift_vec=np.zeros(2),
+        proc_noise_cov_mat=measdiff,
+    )
```
pnkraemer, line 83: I have a feeling you used this naming here because this is how it used to be done, but secretly prefer `dynamics_model` and `measurement_model`, right? :D If so, please go ahead and change it here. We use the longer names inside the filters now anyway, so there is no real reason for `dynmod` anymore.

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+        y_n &= x(t_n) + r_n
+
+    for some scalar :math:`\lambda` and :math:`q(t_n) \sim \mathcal{N}(0, Q)`,
+    :math:`r_n \sim \mathcal{N}(0, R)` for process noise covariance matrix :math:`Q`
+    and measurement noise covariance matrix :math:`R`.
+    Note that the linear, time-invariant dynamics have an equivalent discretization.
+
+    Parameters
+    ----------
+    delta_t : float
```
pnkraemer, line 116: instead of delta_t, I'd advocate for either `step` (which is what is currently used in the old parts of probnum) or `dt` (which is what I'd like to have everywhere soon, and is used inside statespace for instance).

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+        The step size of the discretized time grid on which the model is considered.
+    lam : float
```
pnkraemer, line 118: this would be more aptly named `driftspeed` or `drift_coefficient`, right? :)

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
```diff
@@ -0,0 +1,328 @@
...
+
+    See Also
+    --------
+    :py:class:`probnum.diffeq.GaussianIVPFilter`
+
+    """
+
+    logistic = diffeq.logistic(
+        (0, t_max), initrv=randvars.Constant(np.array([0.1])), params=(6, 1)
+    )
```
pnkraemer, line 318: can we please expose y0 and the params, too? :) Same for ordint.

pnkraemer: with some defaults

tests/test_filtsmooth/test_gaussfiltsmooth/test_iterated_kalman.py
```diff
@@ -1,14 +1,13 @@
 import numpy as np
 import pytest
 
+import probnum.problems.zoo.filtsmooth as pn_filtsmooth_zoo
```
pnkraemer, line 4: do you think the `pn_` adds information? I am not sure, but this sort of import style could be something that we want to adopt package-wide, so let's do this as perfectly as possible from the beginning on.

pnkraemer (commented):
Thanks for taking care of this! :) 
I left some comments in the diff. They mostly center around (1) whether this is a filtsmooth zoo or a statespace zoo, and (2) that perhaps this PR could be a good time to clean up the variable names in these functions (it seems that you copied them from the test-cases).

Please let me know whether you disagree with any points :) Nothing that I commented is a definite "do this", but more of a "what do  you think we should do".

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
pnkraemer: See issue #346

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
schmidtjonathan: Hm, true. I don't have a strong opinion on this, but would also tend to `statespace`. Shall I change it?

src/probnum/problems/zoo/filtsmooth/_ssm_examples.py
schmidtjonathan: I would indeed prefer that, and I'll change it :)

schmidtjonathan:
Thanks for your feedback, I will incorporate the changes and then re-request you as a reviewer :)

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -0,0 +1,470 @@
...
+        - initial random variable
+
+    References
+    ----------
+    .. [1] Särkkä, Simo. Bayesian Filtering and Smoothing. Cambridge University Press,
+        2013.
+
+    """
+
+    dynamics_model = statespace.IBM(ordint=1, spatialdim=2)
```
pnkraemer, line 80: I would suggest using the equivalent discretisation here, because this is what a user might expect here.

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -0,0 +1,470 @@
...
+        initmean = np.array([0.1, 0, 0.0, 0.0])
+        initcov = np.diag([0.0, 1.0, 1.0, 1.0])
+        solver_initrv = randvars.Normal(initmean, initcov)
+
+    statespace_components = dict(
+        dynamics_model=dynamics_model,
+        measurement_model=measurement_model,
+        initrv=solver_initrv,
+    )
+    return logistic_ivp, statespace_components
```
pnkraemer, line 470: I think the return value should be a regression problem with a grid and zero-data (this is a filtsmooth zoo :)), and the logistic ODE defines the measurement that is returned as part of the state space components. what do you think?

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -0,0 +1,470 @@
...
+        - initial random variable
+
+    See Also
+    --------
+    :py:class:`probnum.diffeq.GaussianIVPFilter`
+
+    """
+
+    if timespan is None:
+        timespan = (0.0, 2.0)
```
pnkraemer, line 437: do you prefer this style? because `timespan=(0., 2.)` would've worked just as well (tuples are immutable, if I remember correctly)

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -0,0 +1,470 @@
...
+    statespace_components
+        Dictionary containing
+        - dynamics model
+        - measurement model
+        - initial random variable
+
+    Notes
+    -----
+    No ``RegressionProblem`` object is returned. The dynamics model has to be linearized
+    in order to generate data.
```
pnkraemer, line 360: hmmm this is unfortunate. Maybe we should implement some simple Euler-Maruyama scheme for non-linear SDEs, in order to solve problems like this one.

In any case, do you think it would be more clear if the output of this function would still be a tuple, e.g. `None, {...}` instead of only the dict? This way all the zoo problems would have similar signatures. What do you think?

benchmarks/filtsmooth.py
```diff
@@ -56,7 +14,8 @@ class Filtering:
     params = [[("ekf", "classic"), ("ekf", "sqrt"), ("ukf", "classic")]]
 
     def setup(self, linearization_implementation):
-        dynmod, measmod, initrv, info = load_pendulum()
+        regression_problem, statespace_components = filtsmooth_zoo.pendulum()
```
pnkraemer, line 17: I think it would be a good idea to set the pendulum parameters into stone here. Who knows when we will change the defaults.

benchmarks/filtsmooth.py
```diff
@@ -56,7 +14,8 @@ class Filtering:
     params = [[("ekf", "classic"), ("ekf", "sqrt"), ("ukf", "classic")]]
```
pnkraemer, line 14: for the sqrt implementation to be meaningful, I suppose the problem classes need an `implementation` argument (maybe `forward_implementation` as well as `backward_implementation`). Otherwise cholesky factors keep getting computed and thrown away after each step (unreasonably), which makes benchmarks a bit meaningless.

pnkraemer (commented):
Thanks for taking care of this! There are some proposed changes, but nothing major. The biggest one (I guess) would be an `implementation` argument (or even `forward_implementation` and `backward_implementation`) in the problem classes. 

In general, do you think it would be good to have a `choose_implementation` method in `Transition`s? This way one could in principle change the implementation after initialisation. Or would that be a bad idea? (In any case, this is not relevant for this PR, but for the future)

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -55,54 +57,68 @@ def car_tracking(
...
         state_trans_mat=measurement_matrix,
-        shift_vec=np.zeros(2),
+        shift_vec=np.zeros(measurement_dim),
         proc_noise_cov_mat=measurement_cov,
     )
 
     if initrv is None:
-        initrv = randvars.Normal(np.zeros(4), 0.5 * measurement_variance * np.eye(4))
+        initrv = randvars.Normal(
+            np.zeros(model_dim), 0.5 * measurement_variance * np.eye(model_dim)
```
pnkraemer, line 114: I dont understand why the 0.5 here, but hey, no problems :)

pnkraemer (approved):
tiptop, thanks for your patience! :)

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
schmidtjonathan: I just copied that 😄 I'll remove it if it's unnecessary

schmidtjonathan:
> tiptop, thanks for your patience! :)

Thanks for your support! :)

PR #380, opened 2021-04-21
Scipy Solver and testing

NinaEffenberger:
Wrapper classes for scipy RK23 and RK45-methods. In contrast to the scipy-solvers, the stepsize can be set manually.


tests/test_diffeq/test_wrappedscipyodesolution.py
```diff
@@ -0,0 +1,71 @@
...
+
+def case_lotkavolterra():
+    y0 = np.array([0.1, 0.1])
+    ode = diffeq.lotkavolterra([0.0, 1.0], y0)
+    testsolver = wrappedscipysolver.WrappedScipyRungeKutta(
+        rk.RK45(ode.rhs, ode.t0, y0, ode.tmax), order=4
+    )
+    testsolution = testsolver.solve(diffeq.AdaptiveSteps(0.1, atol=1e-12, rtol=1e-12))
+    scipysolution = testsolution.scipy_solution
+    return testsolution, scipysolution
```
pnkraemer, line 30: thanks for migrating these! :) 

2 comments: (1) there is quite some code duplication here (the only difference between theses two cases are the first two lines, the other ones are copied and pasted -- I would be happy to see those factored out into a single function :)) (2) these ODE test cases exist in both the solution as well as the solver tests. Would be good to only define them once.

tests/test_diffeq/test_wrappedscipyodesolution.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+@pytest.fixture
+def solutiontest():
+    ivp = diffeq.logistic([0.0, 10], np.array([1]))
+    scipysolver = rk.RK45(ivp.rhs, ivp.t0, np.array([1]), ivp.tmax)
+    testsolver = wrappedscipysolver.WrappedScipyRungeKutta(scipysolver, order=4)
+    return testsolver.solve(diffeq.ConstantSteps(0.5))
```
pnkraemer, line 38: where is this fixture used?

tests/test_diffeq/test_wrappedscipysolver.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def case_lotkavolterra():
+    y0 = np.array([0.1, 0.1])
+    ode = diffeq.lotkavolterra([0.0, 1.0], y0)
+    scipysolver = rk.RK45(ode.rhs, ode.t0, y0, ode.tmax)
+    testsolver = wrappedscipysolver.WrappedScipyRungeKutta(
+        rk.RK45(ode.rhs, ode.t0, y0, ode.tmax), order=4
+    )
+    return testsolver, scipysolver
```
pnkraemer, line 38: see my comment for the cases in the other module. If the ODEs were factored out of these functions, and creating a solver and solving the ODEs were subsequent fixtures, one could reuse a lot of code here.

tests/test_diffeq/test_wrappedscipysolver.py
```diff
@@ -0,0 +1,179 @@
...
+    return list([randvars.Constant(1)])
+
+
+@pytest_cases.parametrize_with_cases("testsolver,scipysolver", cases=".")
+def test_initialise(testsolver, scipysolver):
+    time, state = testsolver.initialise()
+    time_scipy = scipysolver.t
+    state_scipy = scipysolver.y
+    np.testing.assert_allclose(time, time_scipy, atol=1e-14, rtol=1e-14)
+    np.testing.assert_allclose(state.mean[0], state_scipy[0], atol=1e-14, rtol=1e-14)
```
pnkraemer, line 77: atol and rtol at 1e-14 is quite a demanding test. Can we be sure that they are supposed to be **identical** here? As soon as there is some sort of rounding errors, this test will be a bit flaky. What do you think?

tests/test_diffeq/test_wrappedscipysolver.py
```diff
@@ -0,0 +1,179 @@
...
+        scipysolver.f,
+        stop_point - start_point,
+        scipysolver.A,
+        scipysolver.B,
+        scipysolver.C,
+        scipysolver.K,
+    )
+    scipy_error_estimation = scipysolver._estimate_error(
+        scipysolver.K, stop_point - start_point
+    )
```
pnkraemer, line 95: how do you feel about comparing to `RungeKutta.step()` here, instead of `rk_step`? This would be -- loosely speaking -- harder to pass, right?

tests/test_diffeq/test_wrappedscipysolver.py
```diff
@@ -0,0 +1,179 @@
...
+    return 0.6
+
+
+@pytest.fixture
+def dense_output():
+    return [base.DenseOutput(0, 1)]
+
+
+@pytest.fixture
+def lst():
```
pnkraemer, line 67: can you maybe rename this to something that is easier to read? It can be super long, if necessary, but `lst` is a bit cryptic :)

pnkraemer (changes requested):
thanks a lot for taking care of this! :) Overall, it looks pretty good! 

I left some reviews in the code diff, please let me know if any of those are unclear. If you have other questions about what to do, please tell me about that as well.

src/probnum/diffeq/wrappedscipysolver.py
```diff
@@ -0,0 +1,127 @@
...
+        -------
+        sol : rk.RkDenseOutput
+            interpolant between the last and current location.
+        """
+
+        Q = self.solver.K.T.dot(self.solver.P)
+        sol = rk.RkDenseOutput(
+            self.solver.t_old, self.solver.t, self.solver.y_old.mean, Q
+        )
+        return sol
```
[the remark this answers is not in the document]
NinaEffenberger: Yes, I first implemented this by testing whether the solver is an instance of RK23 or RK45, but that would mean that other (self-implemented) RK-methods can't use the dense output. I decided to write a comment about the methods that can be used without limitations and the not implemented DOP853 dense output in the class docstring. Please let me know if you have a better idea.

tests/test_diffeq/test_wrappedscipysolver.py
NinaEffenberger: They should be identical, but I set the tolerances to 1e-13, the test should still fulfill its purpose.

tests/test_diffeq/test_wrappedscipysolver.py
NinaEffenberger: You're right, thanks for that hint!

tests/test_diffeq/test_wrappedscipyodesolution.py
NinaEffenberger: I talked to @marvinpfoertner and @schmidtjonathan about this once again. I think _outsourcing_ the tests is better for this purpose. The solver cases can then also be used to test the noisy solvers.

src/probnum/diffeq/wrappedscipyodesolution.py
```diff
@@ -0,0 +1,51 @@
...
+
+DenseOutputValueType = Union[
+    randvars.RandomVariable, _randomvariablelist._RandomVariableList
+]
+"""Dense evaluation of a TimeSeriesPosterior and derived classes return a RandomVariable if evaluated at a single location,
+and a _RandomVariableList if evaluated at an array of locations."""
+
+
+class WrappedScipyODESolution(diffeq.ODESolution):
+    def __init__(self, scipy_solution: OdeSolution, rvs: list):
```
pnkraemer, line 22: A little bit of a docstring would be highly appreciated :)

src/probnum/diffeq/wrappedscipysolver.py
```diff
@@ -0,0 +1,133 @@
...
+        return probnum_solution
+
+    def method_callback(self, time, current_guess, current_error):
+        """Call dense output after each step and store the interpolants."""
+        dense = self.dense_output()
+        self.interpolants.append(dense)
+
+    def dense_output(self):
+        """Compute the interpolant after each step with a quartic interpolation
+        polynomial.
```
pnkraemer, line 122: the "quartic interpolation polynomial" is not true, right?

pnkraemer: at least not always... :)

src/probnum/diffeq/wrappedscipysolver.py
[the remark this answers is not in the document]
pnkraemer: Okay, thanks for clarifying :) However, was actually *happens* when one uses DOP853 in here? Does this still work, but is suboptimal in terms of speed (because it ignores the first-same-as-last property of DOP853), or does it break? What would a user expect? A simple solution to avoid these issues would be to check for DOP853 (as in: `if isinstance(self.solver, scipy.DOP853): raise TypeError` (which should probably happen at the init, not here?)) in which case the usage of these functions would be restricted to RK45, RK23, and whichever RK methods one decides to implement :)

tests/test_diffeq/test_wrappedscipysolver.py
```diff
@@ -0,0 +1,161 @@
...
+
+from probnum import randvars
+from probnum.diffeq import odesolution, wrappedscipyodesolution
+
+
+@pytest_cases.fixture
+@pytest_cases.parametrize_with_cases("solvers", cases=".test_wrappedscipy_cases")
+
+# Workaround: usually the input of this would be "testsolver, scipysolver" instead of "solvers"
+# see issue https://github.com/smarie/python-pytest-cases/issues/202
```
pnkraemer, line 15: the empty lines seem confusing here...

pnkraemer (commented):
Thanks for updating :) I have very few minor things, after which we can merge!

src/probnum/diffeq/wrappedscipysolver.py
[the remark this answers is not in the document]
NinaEffenberger: I added this check and included a test for it.

src/probnum/diffeq/wrappedscipysolver.py
```diff
@@ -0,0 +1,136 @@
...
+        # ProbNum ODESolver needs an ivp
+        ivp = diffeq.IVP(
+            timespan=[self.solver.t, self.solver.t_bound],
+            initrv=randvars.Constant(self.solver.y),
+            rhs=self.solver._fun,
+        )
+
+        # Dopri853 as implemented in SciPy computes the dense output differently
+        if isinstance(solver, rk.DOP853):
+            raise TypeError
```
pnkraemer, line 32: can you add an error message to the type error please?

src/probnum/diffeq/wrappedscipysolver.py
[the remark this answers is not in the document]
pnkraemer: wonderful!

pnkraemer (commented):
essentially approved ;)

pnkraemer (approved):
Thanks for your patience nina! Once the tests are done passing, go ahead and merge :))

PR #386, opened 2021-04-28
Stabilise matrix inversion in discrete transitions

pnkraemer:
This PR replaces all `inv` and `solve` by more appropriate replacements. This has a surprisingly big impact on the square-root transitions in terms of stability and dealing with really ill-conditioned matrices. 

(This impact I noticed when working with the code, there is no proof I can show here :))


schmidtjonathan (approved):
Perfect, thanks for taking care of this!

Issue #349, opened 2021-03-12
Random State of Random Variables


pnkraemer:
**Description:**
Random variables should not have a random state themselves.

**Problem:**
Random variables with random state lead to unexpected behavior in particular when doing arithmetic, since then random states have to be combined.

**Solution:**
- make random variables stateless in the sense that they do not have a random state or an instance of a random number generator
- only ``sample(size=(), random_state=None)`` gets a random generator.

Closed by #387.

PR #387, opened 2021-04-28
Remove random state from RandomVariable initialisation

pnkraemer:
This PR removes the `random_state` argument from `RandomVariable.__init__` (and all subclasses), and instead moves it to `sample()` (and all related functions and methods) as discussed offline a few weeks ago. See #349.

## Question 1
Shall I place an empty function `push_forward(base_measure, samples)` into `RandomVariable` that established the push-forward sampling (see #349)? While no random variable will use it for now, at least the signature (and "desired existence") of such a method could be set in stone. An application would be https://probnum-evaluation.readthedocs.io/en/latest/tutorials/animate_gaussian_distributions.html. 

## Question 2
There were a few test cases in NormalTestCase and in RandomVariableTestCase which raised warnings (on purpose? passing np.nan and np.inf to scalar multiplication tests will not go smoothly at all). Even though this is not part of this PR, I removed those (because, honestly, they annoyed me a little bit). I will comment in the diff which changes in this PR are due to this. If these test parameters tested any significant behaviour that I missed, let me know.
In any case, if a warning shall be tested for, I think there would be better solutions -- in fact, I might be in favour of treating warnings as errors while testing. See https://docs.pytest.org/en/stable/warnings.html for potential solutions.  

As mentioned above, I left comments in the diff. Please go ahead and resolve everything you do not want to complain about :)


src/probnum/randvars/_categorical.py
```diff
@@ -54,7 +51,13 @@ def _sample_categorical(size=()):
             arrays with `ndim > 1`, but `self.support` can be just that.
             This detour via the `mask` avoids this problem.
             """
-            mask = np.random.choice(
+
+            # By the time this function is called, the random_state is a
+            # RandomState(MT19937) object, which offers a choice function.
+            # We cannot direct the call to SciPy, since it does not offer a
+            # Categorical distribution
+            # (and doing it via multinomial was too much reshaping, etc.)
```
pnkraemer, line 59: is that clear? If so, resolve.

tests/test_randvars/test_normal.py
```diff
@@ -37,14 +41,14 @@ def setUp(self):
             (
                 np.random.uniform(size=(2, 2)),
                 linops.SymmetricKronecker(
-                    A=np.array([[1.0, 2.0], [2.0, 1.0]]),
+                    A=np.array([[1.0, 2.0], [2.0, 10.0]]),
```
pnkraemer, line 44: this SymmetricKronecker matrix was not positive semi-definite. Now it is. Thus, there are no more warnings in the sampling code.

tests/test_randvars/test_normal.py
```diff
@@ -37,14 +41,14 @@ def setUp(self):
...
                     B=np.array([[5.0, -1.0], [-1.0, 10.0]]),
                 ).todense(),
             ),
             # Operatorvariate
             (
                 np.array([1.0, -5.0]),
-                linops.MatrixMult(A=np.array([[2.0, 1.0], [1.0, -0.1]])),
+                linops.MatrixMult(A=np.array([[2.0, 1.0], [1.0, 1.0]])),
```
pnkraemer, line 51: same here

tests/test_randvars/test_normal.py
```diff
@@ -61,7 +65,7 @@ def setUp(self):
             (
                 linops.MatrixMult(A=np.random.uniform(size=(2, 2))),
                 linops.SymmetricKronecker(
-                    A=np.array([[1.0, 2.0], [2.0, 1.0]]),
+                    A=np.array([[1.0, 2.0], [2.0, 10.0]]),
```
pnkraemer, line 68: same here

tests/test_randvars/test_random_variable.py
```diff
@@ -20,7 +20,7 @@ def setUp(self) -> None:
         np.random.seed(42)
 
         # Random variable instantiation
-        self.scalars = [0, int(1), 0.1, np.nan, np.inf]
+        self.scalars = [0, int(1), 0.1]
```
pnkraemer, line 23: the nan and inf screwed with the test_scalar_mult below (warnings were raised). The warnings bothered me slightly, and I didnt see the point of this piece of code, but may well have missed something. This removal will be easy to undo in any case.

src/probnum/randvars/_random_variable.py
```diff
@@ -446,7 +418,9 @@ def in_support(self, x: _ValueType) -> bool:
 
         return in_support
 
-    def sample(self, size: ShapeArgType = ()) -> _ValueType:
+    def sample(
+        self, size: ShapeArgType = (), random_state: RandomStateArgType = None
```
marvinpfoertner, line 422: I don't think it is a good idea to allow passing seeds directly to the sample method, since in this case a random state object is created for one sample process only. IMHO this is weird behavior and might also lead to weird behavior, since a user gets the exact same samples when calling sample twice with a seed as `random_state`, e.g. in a loop or in some deeply nested code.

I think we should only allow `random_state` args of type `RandomStateType`, i.e. only `np.random.Generator`s and `np.random.RandomState`s. If `None` is given, a global `np.random.Generator` should be used. At a future point, we might want to drop support for the deprecated `np.random.RandomState` alltogether.

src/probnum/randvars/_random_variable.py
```diff
@@ -457,7 +431,9 @@ def sample(self, size: ShapeArgType = ()) -> _ValueType:
         if self.__sample is None:
             raise NotImplementedError("No sampling method provided.")
 
-        return self.__sample(_utils.as_shape(size))
+        return self.__sample(
+            _utils.as_shape(size), _utils.as_random_state(random_state)
```
marvinpfoertner, line 435: See comment above, i.e. drop `as_random_state` call here

src/probnum/randvars/_random_variable.py
```diff
@@ -122,9 +117,10 @@ def __init__(
         self,
         shape: ShapeArgType,
         dtype: DTypeArgType,
-        random_state: RandomStateArgType = None,
         parameters: Optional[Dict[str, Any]] = None,
-        sample: Optional[Callable[[ShapeType], _ValueType]] = None,
+        sample: Optional[
+            Callable[[ShapeType, Optional[RandomStateArgType]], _ValueType]
```
marvinpfoertner, line 122: `RandomStateType` instead of `Optional[RandomStateArgType]` for the second argument

src/probnum/randvars/_random_variable.py
```diff
@@ -564,8 +540,7 @@ def __getitem__(self, key: ArrayLikeGetitemArgType) -> "RandomVariable":
         return RandomVariable(
             shape=np.empty(shape=self.shape)[key].shape,
             dtype=self.dtype,
-            random_state=_utils.derive_random_seed(self.random_state),
-            sample=lambda size: self.sample(size)[key],
+            sample=lambda size, random_state=None: self.sample(size, random_state)[key],
```
marvinpfoertner, line 543: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -588,8 +563,9 @@ def reshape(self, newshape: ShapeArgType) -> "RandomVariable":
         return RandomVariable(
             shape=newshape,
             dtype=self.dtype,
-            random_state=_utils.derive_random_seed(self.random_state),
-            sample=lambda size: self.sample(size).reshape(size + newshape),
+            sample=lambda size, random_state=None: self.sample(
```
marvinpfoertner, line 566: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -611,8 +587,9 @@ def transpose(self, *axes: int) -> "RandomVariable":
         return RandomVariable(
             shape=np.empty(shape=self.shape).transpose(*axes).shape,
             dtype=self.dtype,
-            random_state=_utils.derive_random_seed(self.random_state),
-            sample=lambda size: self.sample(size).transpose(*axes),
+            sample=lambda size, random_state=None: self.sample(
```
marvinpfoertner, line 590: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -631,8 +608,9 @@ def __neg__(self) -> "RandomVariable":
         return RandomVariable(
             shape=self.shape,
             dtype=self.dtype,
-            random_state=_utils.derive_random_seed(self.random_state),
-            sample=lambda size: -self.sample(size=size),
+            sample=lambda size, random_state=None: -self.sample(
```
marvinpfoertner, line 611: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -647,8 +625,9 @@ def __pos__(self) -> "RandomVariable":
         return RandomVariable(
             shape=self.shape,
             dtype=self.dtype,
-            random_state=_utils.derive_random_seed(self.random_state),
-            sample=lambda size: +self.sample(size=size),
+            sample=lambda size, random_state=None: +self.sample(
```
marvinpfoertner, line 628: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -663,8 +642,9 @@ def __abs__(self) -> "RandomVariable":
         return RandomVariable(
             shape=self.shape,
             dtype=self.dtype,
-            random_state=_utils.derive_random_seed(self.random_state),
-            sample=lambda size: abs(self.sample(size=size)),
+            sample=lambda size, random_state=None: abs(
```
marvinpfoertner, line 645: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -970,7 +946,7 @@ class DiscreteRandomVariable(RandomVariable[_ValueType]):
     ...     "p" : p}
     >>>
     >>> # Sampling function
-    >>> def sample_categorical(size=()):
+    >>> def sample_categorical(size=(), random_state=None):
```
marvinpfoertner, line 949: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -1006,9 +982,10 @@ def __init__(
         self,
         shape: ShapeArgType,
         dtype: DTypeArgType,
-        random_state: Optional[RandomStateType] = None,
         parameters: Optional[Dict[str, Any]] = None,
-        sample: Optional[Callable[[ShapeArgType], _ValueType]] = None,
+        sample: Optional[
+            Callable[[ShapeArgType, Optional[RandomStateArgType]], _ValueType]
```
marvinpfoertner, line 987: See comment in `RandomVariable.__init__`

src/probnum/randvars/_random_variable.py
```diff
@@ -1191,7 +1163,7 @@ class ContinuousRandomVariable(RandomVariable[_ValueType]):
     >>> parameters_uniform = {"bounds" : [a, b]}
     >>>
     >>> # Sampling function
-    >>> def sample_uniform(size=()):
+    >>> def sample_uniform(size=(), random_state=None):
```
marvinpfoertner, line 1166: No need for the `=None` after `random_state`

src/probnum/randvars/_random_variable.py
```diff
@@ -1228,9 +1200,10 @@ def __init__(
         self,
         shape: ShapeArgType,
         dtype: DTypeArgType,
-        random_state: Optional[RandomStateType] = None,
         parameters: Optional[Dict[str, Any]] = None,
-        sample: Optional[Callable[[ShapeArgType], _ValueType]] = None,
+        sample: Optional[
+            Callable[[ShapeArgType, Optional[RandomStateArgType]], _ValueType]
```
marvinpfoertner, line 1205: See comment in `RandomVariable.__init__`

src/probnum/randvars/_utils.py
```diff
@@ -27,10 +27,9 @@ def asrandvar(obj: Any) -> _random_variable.RandomVariable:
     >>> from scipy.stats import bernoulli
     >>> import probnum as pn
     >>> bern = bernoulli(p=0.5)
-    >>> bern.random_state = 42  # Seed for reproducibility
     >>> bern_pn = pn.asrandvar(bern)
-    >>> bern_pn.sample(size=5)
-    array([1, 1, 1, 0, 0])
+    >>> bern_pn.sample(size=5, random_state=42)
```
marvinpfoertner, line 31: See comment in `_normal.py`

tests/test_randvars/test_random_variable.py
marvinpfoertner: I don't think I wrote the test, but from my point of view it makes sense to keep it.
As random variables should behave exactly like numpy arrays, they should also mimic their behavior when being multiplied by `np.nan` and `np.inf`.

As for the warnings, we can even assert they're there https://docs.pytest.org/en/stable/warnings.html#warns.

marvinpfoertner (commented):
Question 1
-----------
I think the right way to resolve this is to give `RandomVariable.__init__` an optional argument `push_forward` of type `Callable[[str, np.ndarray], np.ndarray]` which handles the sample transform.
Then the method `push_forward` can be implemented along the lines of `sample` and need not be empty but fully functional once the user specifies it in the constructor :)

Question 2
-----------
I answered in the diff

tests/test_randvars/test_random_variable.py
pnkraemer: yes, I agree that they are worth testing, but not the way they are currently tested (by calling it and getting a bunch of warnings as outputs that need to be read by a human). I will go ahead and write a specialised test for this instead.

src/probnum/randvars/_categorical.py
```diff
@@ -54,7 +51,13 @@ def _sample_categorical(size=()):
...
+            mask = random_state.choice(
```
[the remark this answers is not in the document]
pnkraemer: Hmmm true. Until now, I thought that this is what a mask was, but seems like I was wrong. Will rename.

src/probnum/randvars/_normal.py
```diff
@@ -66,8 +62,8 @@ class Normal(_random_variable.ContinuousRandomVariable[_ValueType]):
     Examples
     --------
     >>> from probnum import randvars
-    >>> x = randvars.Normal(mean=0.5, cov=1.0, random_state=42)
-    >>> x.sample(size=(2, 2))
+    >>> x = randvars.Normal(mean=0.5, cov=1.0)
+    >>> x.sample(size=(2, 2), random_state=42)
```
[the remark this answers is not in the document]
pnkraemer: "A little unsettling" -- you put that very diplomatically :D 

I kind of agree, but had reasons to do it like this. Let's discuss it offline :)

pnkraemer:
Thanks for the feedback Marvin! 

Ad Q1: Alright, if you think it is worth putting it in already, I will do it. Let's discuss offline how the base measure should be indicated in the string. 

Ad Q2: Answered in the diff (in a nutshell: since you seem to consider it important, I will rewrite the test, but I think the current approach is no solution).

Ad Q3 (new Q! :smile:  ):  let's also discuss the random state thing offline. I will ping you.

pnkraemer:
@alpiges I had to make some changes to the Monte Carlo policies in `quad`. Can you please confirm that you are okay with the diff? If you want something to be solved differently, please let me know (either here, or offline). Thanks :)

src/probnum/quad/_bayesquad.py
```diff
@@ -118,8 +123,10 @@ def bayesquad(
         measure = LebesgueMeasure(domain=domain, dim=input_dim)
 
     # Choose Method
+    if policy == "bmc" and rng is None:
+        rng = np.random.default_rng()
```
pnkraemer, line 127: Do you want me to throw an error instead?

src/probnum/quad/_bayesquad.py
```diff
@@ -98,14 +103,14 @@ def bayesquad(
...
-    >>> np.random.seed(0)
+
     >>> input_dim = 1
     >>> domain = (0, 1)
     >>> def f(x):
     ...     return x
     >>> F, info = bayesquad(fun=f, input_dim=input_dim, domain=domain)
     >>> print(F.mean)
-    0.5000
+    0.5
```
pnkraemer, line 113: the seed does not have any effect anymore. To recreate the other functionality, I could do
```python
rng = np.random.default_rng(seed=0)
F, info = bayesquad(fun=f, input_dim=input_dim, domain=domain, rng=rng)
```
or even
```python
rng = np.random.default_rng(seed=0)
F, info = bayesquad(fun=f, input_dim=input_dim, domain=domain, policy="bmc", rng=rng)
```
Since I thought the simplest interface would be the target here, I decided against that. But let me know if you want me to change it.

src/probnum/randvars/_scipy_stats.py
```diff
@@ -221,11 +219,18 @@ def in_support(x):
     else:
         in_support = None
 
+    def sample_from_scipy_rv(rng, size):
+        return scipy_rv.rvs(size=size, random_state=rng)
+
+    if hasattr(scipy_rv, "rvs"):
+        sample_wrapper = sample_from_scipy_rv
+    else:
+        sample_wrapper = None
```
pnkraemer, line 228: @marvinpfoertner I am not sure I understood this block of code perfectly. My goal here was to squeeze the scipy sampler into the probnum randomvariable sample interface (with input rng & size). If there is a more elegant way of doing this, please let me know.

tests/test_randvars/test_random_variable.py
```diff
@@ -104,7 +104,11 @@ def test_rv_scalarmult(self):
         """Multiplication of random variables with scalar constants."""
         for (alpha, rv) in list(itertools.product(self.scalars, self.randvars2d)):
             with self.subTest():
-                z = alpha * rv
+                if np.isinf(alpha):
+                    with self.assertWarns(RuntimeWarning):
+                        z = alpha * rv
```
pnkraemer, line 109: this catches the warning reg. random variable arithmetic and np.inf.

pnkraemer:
Alright, next iteration is done. @marvinpfoertner please let me know what you think.
The major changes are that now the user is forced to provide random number generators (np.random.Generator) to sampling-related methods. We discussed this in the software meeting 3 (?) weeks ago. This necessitated quite a few changes, but they should be easy to review (I think...). 

I also wrote an explicit test that catches the warning when a random variable is multiplied with np.inf. This should resolve the discussion we had earlier. 

I decided to not introduce a `push_forward` method into RandomVariable, because the PR is already very large (and because it seems to be not very urgent). 

On a side note, the overall diff is negative (by a few hundred lines), which is not visible anymore due to the notebook changes... :( All in all, I think the complexity of the library is reduced significantly while the clarity has been improved (now it is more evident which methods rely on random sampling and which ones dont -- this is especially different for stochastic policies now). 

If there is anything non-trivial, feel free to ping me offline.

marvinpfoertner (approved):
Nice 👍🏻

> All in all, I think the complexity of the library is reduced significantly while the clarity has been improved (now it is more evident which methods rely on random sampling and which ones dont -- this is especially different for stochastic policies now).

I agree. And as a bonus, we have less state in our objects which results in less side-effects and less problems :)

src/probnum/quad/_bayesquad.py
pnkraemer: > I don't really get what you mean by
setting `np.random.seed(1)` has no effect since all the underlying sampling uses fixed random number generators

> other functionality
Fixing a seed

marvinpfoertner: Then this is fine :)

src/probnum/randvars/_random_variable.py
```diff
@@ -446,18 +415,20 @@ def in_support(self, x: _ValueType) -> bool:
...
+        rng
+            Random number generator used for sampling.
+        size
             Size of the drawn sample of realizations.
         """
         if self.__sample is None:
             raise NotImplementedError("No sampling method provided.")
 
-        return self.__sample(_utils.as_shape(size))
+        return self.__sample(rng=rng, size=_utils.as_shape(size))
```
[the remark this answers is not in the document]
pnkraemer: done

alpiges:
Re `quad`: your changes look consistent. The structure of `quad` is anyways changing substantially in #420, so we'll need to port the changes there.

src/probnum/quad/_bayesquad.py
```diff
@@ -31,6 +31,7 @@ def bayesquad(
     measure: Optional[IntegrationMeasure] = None,
     method: str = "vanilla",
     policy: str = "bmc",
+    rng: np.random.Generator = None,
```
alpiges, line 34: Why do we want `rng` to be an argument in the first place? Why would I want to use a different `rng` other than the default one?

alpiges (approved):
Looks fine from the `quad` side

src/probnum/quad/_bayesquad.py
pnkraemer: I messaged you offline :)

PR #389, opened 2021-04-29
Implement filtsmooth methods as generators

schmidtjonathan:
This PR implements Kalman filtering methods as generators in order to make partial results accessible.

### DIFF
--- a/docs/source/tutorials/filtsmooth/particle_filtering.ipynb
+++ b/docs/source/tutorials/filtsmooth/particle_filtering.ipynb
@@ -250,7 +250,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "posterior = pf.filter(regression_problem)"
+    "posterior, _ = pf.filter(regression_problem)"
    ]
   },
   {
@@ -2780,7 +2780,7 @@
     "    locations=locs,\n",
     ")\n",
     "\n",
-    "ode_posterior = ode_pf.filter(regression_problem)"
+    "ode_posterior, _ = ode_pf.filter(regression_problem)"
    ]
   },
   {
--- a/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
+++ b/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
@@ -225,6 +225,8 @@ class Kalman(BayesFiltSmooth):
         ------
         filtrv
             Random variable returned from prediction and update of the Kalman filter.
+        info_dict
+            Dictionary containing filtering information
 
         See Also
         --------
--- a/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
+++ b/src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
@@ -35,6 +35,7 @@ class Kalman(BayesFiltSmooth):
     def iterated_filtsmooth(
         self,
         regression_problem: problems.RegressionProblem,
+        init_posterior: Optional[SmoothingPosterior] = None,
         stopcrit: Optional[stoppingcriterion.StoppingCriterion] = None,
     ):
         """Compute an iterated smoothing estimate with repeated posterior linearisation.
@@ -46,6 +47,10 @@ class Kalman(BayesFiltSmooth):
         Parameters
         ----------
         regression_problem
+        init_posterior
+            Initial posterior to linearize at. Defaults to computing a (non-iterated)
+            smoothing posterior, which amounts to linearizing at the prediction
+            random variable.
         stopcrit: StoppingCriterion
             A stopping criterion for iterated filtering.
 
@@ -58,15 +63,22 @@ class Kalman(BayesFiltSmooth):
         RegressionProblem: a regression problem data class
         """
 
-        for iter_result in self.iterated_filtsmooth_generator(
-            regression_problem, stopcrit
+        smoothing_post = init_posterior
+        info_dicts = None
+        for smoothing_post, info_dicts in self.iterated_filtsmooth_generator(
+            regression_problem, smoothing_post, stopcrit
         ):
             pass
-        return iter_result  # pylint: disable=undefined-loop-variable
+
+        if smoothing_post is None or info_dicts is None:
+            raise ValueError("Iterated filter did not perform a single iteration.")
+
+        return smoothing_post, info_dicts
 
     def iterated_filtsmooth_generator(
         self,
         regression_problem: problems.RegressionProblem,
+        init_posterior: Optional[SmoothingPosterior] = None,
         stopcrit: Optional[stoppingcriterion.StoppingCriterion] = None,
     ):
         """Compute iterated smoothing estimates with repeated posterior linearisation.
@@ -78,12 +90,18 @@ class Kalman(BayesFiltSmooth):
         Parameters
         ----------
         regression_problem
+        init_posterior
+            Initial posterior to linearize at. Defaults to computing a (non-iterated)
+            smoothing posterior, which amounts to linearizing at the prediction
+            random variable.
         stopcrit: StoppingCriterion
             A stopping criterion for iterated filtering.
 
         Yields
         ------
         SmoothingPosterior
+        info_dicts
+            list of dictionaries containing filtering information
 
         See Also
         --------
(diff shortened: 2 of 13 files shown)


src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -49,6 +49,43 @@ def iterated_filtsmooth(
...
+        --------
+        RegressionProblem: a regression problem data class
+        """
+
+        result = None
+        for iter_result in self.iterated_filtsmooth_generator(
+            regression_problem, stopcrit
+        ):
+            result = iter_result
+        return result
```
schmidtjonathan, line 66: If you have a more elegant solution, feel free to request changes.
I thought about something like

```python
return next(reversed(self.iterated_filtsmooth_generator(...)))
```

but I think it might be less efficient. If efficiency is relevant at all here ....

pnkraemer: I am not sure about your second proposal. I like the for loop, but you dont need the `result = iter_result` assignment I think. ```for iter_result in self.iterated_filtsmooth_generator(...): pass; return iter_result``` would work, right? 

Apparently (https://stackoverflow.com/questions/2138873/cleanest-way-to-get-last-item-from-python-iterator) ```*_, last = self.iterated_filtsmooth_generator(...)``` would also be an option? I have not checked this though

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -93,54 +134,112 @@ def filtsmooth(
...
             This is used for iterated filtering and smoothing. For standard filtering, this can be ignored.
 
         Returns
-        -------
+        ------
         KalmanPosterior
             Posterior distribution of the filtered output
+        info_dicts
+            Only if ``returns_info_dicts`` is ``True``: list of dictionaries containing
+            filtering information
```
pnkraemer, line 180: what is the idea behind this change? I am not sure how comfortable I am about changing the number of return values depending on an input argument. However, if there is a good reason to do this, I am all ears :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: checked it, the `*_, last = ...` thing works :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: Merely for backwards compatibility. Do you prefer having it returned in all cases?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: Okay. I like the second proposal! The first, with the `for ... pass ...` loop doesn't work -- that I checked.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: Hmmm. We dont do this anywhere. Maybe make this info a part of the posterior? Would that be something interesting? If you like, feel free to come by my office and we discuss this. I suppose this would have a wider impact than `Kalman` (but would also apply to the ODE solvers for instance)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: Okay nevermind, your first proposal works, pylint was just telling me that it isn't a good idea. However, I would ignore pylint in this case and use the first proposal because this way, the objects aren't kept in memory.

docs/source/tutorials/filtsmooth/particle_filtering.ipynb
```diff
@@ -2780,7 +2780,7 @@
     "    locations=locs,\n",
     ")\n",
     "\n",
-    "ode_posterior = ode_pf.filter(regression_problem)"
+    "ode_posterior, _ = ode_pf.filter(regression_problem)"
```
pnkraemer, line 2783: I cant believe it, a readable notebook diff??

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -35,6 +35,7 @@ class Kalman(BayesFiltSmooth):
     def iterated_filtsmooth(
         self,
         regression_problem: problems.RegressionProblem,
+        init_posterior: Optional[SmoothingPosterior] = None,
```
pnkraemer, line 38: Not complaining about the variable name, but I would've expected you go for `initial_posterior` here ;)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -46,9 +47,62 @@ def iterated_filtsmooth(
         Parameters
         ----------
         regression_problem
+        init_posterior
+            Initial posterior to linearize at. Defaults to computing a (non-iterated)
+            smoothing posterior, which amounts to linearizing at the prediction
+            random variable.
```
pnkraemer, line 53: super minor comment: I find this a bit hard to parse. But that might just be me reading this at the end of my workday... ;)

pnkraemer: Definitely feel free to ignore htis!

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -46,9 +47,62 @@ def iterated_filtsmooth(
...
+
+        smoothing_post = init_posterior
+        info_dicts = None
+        for smoothing_post, info_dicts in self.iterated_filtsmooth_generator(
+            regression_problem, smoothing_post, stopcrit
+        ):
+            pass
+
+        if smoothing_post is None or info_dicts is None:
+            raise ValueError("Iterated filter did not perform a single iteration.")
```
pnkraemer, line 74: when would this valueerror happen? And would the situation you describe in the message rather be a warning than an error?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -57,23 +111,27 @@ def iterated_filtsmooth(
...
-        # Initialise iterated smoother
-        old_posterior = self.filtsmooth(
-            regression_problem,
-            _previous_posterior=None,
-        )
-        new_posterior = old_posterior
+        info_dicts = []
+        if init_posterior is None:
+            # Initialise iterated smoother
+            init_posterior, info_dicts = self.filtsmooth(
```
pnkraemer, line 117: super minor again: why `info_dicts` (as opposed to `info_dict`)?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -122,25 +184,69 @@ def filter(
...
+
+        posterior = FilteringPosterior(
+            locations=regression_problem.locations,
+            states=filtered_rvs,
+            transition=self.dynamics_model,
+        )
+
+        return posterior, info_dicts
+
+    def filter_generator(
```
pnkraemer, line 210: naming question: is this a `filter_generator` or something along the lines of a `filter_solution_generator`?

pnkraemer (commented):
Great job overall! I have only super minor comments (mostly naming questions, which -- I am aware -- is super subjective), which you may ignore if you like.

The only one that might be worth discussing is: is the filtering generator method a `filter_generator` or something along the lines of a  `filtering_solution_generator` or `filtering_posterior_generator` (`filtering_distribution_generator`?).

Once this is settled I will approve! :)

docs/source/tutorials/filtsmooth/particle_filtering.ipynb
schmidtjonathan: I edited the raw file 😄

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: decision: `filtered_states_generator`

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: Never. Decision: remove

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -57,23 +107,27 @@ def iterated_filtsmooth(
...
-        new_posterior = old_posterior
+        info_dicts = []
+        if init_posterior is None:
+            # Initialise iterated smoother
+            init_posterior, info_dicts = self.filtsmooth(
+                regression_problem,
+                _previous_posterior=None,
+            )
+
+        new_posterior = init_posterior
```
pnkraemer, line 118: just checking: so you decided against the "else" solution here? :)

pnkraemer (approved):
See my final comment. Go ahead and merge though, you have my approval :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: Nope, forgot, thanks for the reminder! Just one sec

PR #396, opened 2021-05-03
Multiple measurement models in Kalman filter

pnkraemer:
This PR introduces the possibility of using Kalman filtering with multiple measurement models.

Todos:
- [x] Write a function `merge_regression_problems` that makes a regression problem out of two problems.
- [x] Move measurement models into `filter()` and out of `init`.
- [x] iterate over zip(measurement_models, times, datapoints) in filter()
- [x] initarg into the prior (i.e. Kalman) 
- [x]  refactor filter iteration (loop over: move to next point -> update, which also covers the initial data point!) (edit: maybe this is a future PR)

Edit: turns out a lot of code had to be touch to make this bad boy work! In order to do multiple measurement models (in a generator-friendly way), I had to introduce proper zipping, which in turn required me to think about the final point above (in the Kalman filter this was fairly simple, the PF took some time to tidy up!). This, on the other hand, required thinking about t0 values for the prior (different initialisation for t0 != times[0] and t0=times[0]), which in turn forced me to include an `initarg` argument -- since this was overdue anyway, now the filters get a `MarkovProcess` at initialisation. Fixing all subsequent tests and dependent subpackages (diffeq!), now means that I have touched much more than I thought.


pnkraemer:
@schmidtjonathan please have a look now (or whenever you find time). Of course, feel free to criticise everything. The perhaps most fruitful feedback will be whether you can see any potential issues with using this for your project code :) Thanks! You can also ping me offline, if there are any bigger questions.

src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
```diff
@@ -1,42 +1,50 @@
...
         or ContinuousUKFComponent are also valid. Describes a random process in :math:`K` dimensions.
-        If an integrator, `K=spatialdim(ordint+1)` for some spatialdim and ordint.
-    measurement_model
-        Measurement model. Usually an DiscreteLTIGaussian, but any DiscreteLinearGaussian is acceptable.
-        This model maps the `K` dimensional prior state (see above) to the `L` dimensional space in which the observation ''live''.
-        For 2-dimensional observations, `L=2`.
-        If an DiscreteLTIGaussian, the measurement matrix is :math:`L \\times K` dimensional, the forcevec is `L` dimensional and the meascov is `L \\times L` dimensional.
-    initrv
-        Initial random variable for the prior. This is a `K` dimensional Gaussian distribution (not `L`, because it belongs to the prior)
+        If the transition is an integrator, `K=spatialdim(ordint+1)` for some spatialdim and ordint.
```
schmidtjonathan, line 39: `K = spatialdim * (ordint + 1)` . The asterisk could avoid some confusion.

src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
```diff
@@ -231,121 +276,60 @@ def filtered_states_generator(
...
+            measurement_model = itertools.repeat(measurement_model, len(times))
 
-        _linearise_update_at = (
-            None if _previous_posterior is None else _previous_posterior(times[0])
-        )
+        # It is not clear at the moment how to implement this cleanly.
+        if not np.all(np.diff(times) > 0):
+            raise ValueError(
+                "Gaussian filtering cannot handle repeating time points currently."
+            )
```
schmidtjonathan, line 289: This also checks whether the time points are sorted, the error message, however, only reports that the locations are not _strictly_ sorted. Should we clarify that? Would have to be sth. like

```python
if not np.all(np.diff(times) >= 0):
   # either ...
   raise ValueError(
       "Locations have to be sorted"
   )
   #or
   times = np.sort(times)

if not np.all(np.diff(times) > 0):
       raise ValueError(
           "Gaussian filtering cannot handle repeating time points currently."
       )
```

pnkraemer: I see what you mean, good point. However, for simplicity reasons, I would rather change the current error message into "Gaussian filtering expects sorted, non-repeating time points." Otherwise that would be a lot of code for a very specific corner case?!

schmidtjonathan: Yep, sounds good.

src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
```diff
@@ -231,121 +276,60 @@ def filtered_states_generator(
...
+                )
+                output = self.prior_process.transition.forward_rv(
+                    curr_rv, t, dt=dt, _linearise_at=linearise_predict_at
+                )
+                curr_rv, info_dict["predict_info"] = output
+
+            # Update (even if there is no increment)
+            linearise_update_at = (
+                None if _previous_posterior is None else _previous_posterior(t)
             )
```
schmidtjonathan, line 326: Are we still planning to get rid of the `linearise_at` arguments? Will this be dealt with in a separate PR?

schmidtjonathan: (Also regarding the performance, this might be an issue for the rather near than far future)

src/probnum/filtsmooth/_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    # Check input types of measurement models: we explicitly want numpy arrays.
+    # More abstract merging (e.g. of lists or generators) is not supported currently.
+    errormsg_type_measmods = (
+        f"The measurement models are expected to be of type `{np.ndarray}'"
+        f"but type `{type(measurement_models1)}' was received."
+    )
+    if not isinstance(measurement_models1, np.ndarray):
+        raise TypeError(errormsg_type_measmods)
+    if not isinstance(measurement_models2, np.ndarray):
+        raise TypeError(errormsg_type_measmods)
```
schmidtjonathan, line 108:
```python
if not (isinstance(measurement_models1, np.ndarray) and isinstance(measurement_models2, np.ndarray)):
  raise ...
```
?

pnkraemer: Two reasons I slightly prefer the current one:
1. an **or** would be better than an **and** here, right? 
2. The line would be so long that black would make the whole thing into 3-4 lines anyway, so the space is the same for both versions, but the current one is a bit more compact. I find it easier to read if there are fewer linebreaks for single statements

What do you think?

src/probnum/filtsmooth/_particlefiltsmooth/_importance_distributions.py
```diff
@@ -0,0 +1,142 @@
...
+        """Process the initial random variable based on data."""
+        raise NotImplementedError
+
+    def log_correction_factor(
+        self, proposal_state, importance_rv, dynamics_rv, old_weight
+    ) -> float:
+        return (
+            dynamics_rv.logpdf(proposal_state)
+            - importance_rv.logpdf(proposal_state)
+            + np.log(old_weight + 1e-14)
```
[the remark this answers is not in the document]
pnkraemer: log(0) can happen, the nugget is a bit random though. 1e-14 seems enough to stop this log(0) from happening, but 1e-12, 1e-13, and 1e-15 would probably do in the same way

src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
pnkraemer: yes, this should be changed soon. Should I do this in here, on shall we tackle it in a separate PR? I would tend towards the latter option, because it is a little bit orthogonal to the current PR

tests/test_filtsmooth/test_utils.py
```diff
@@ -0,0 +1,189 @@
...
+
+    # Single models are rejected (1st input)
+    with pytest.raises(TypeError):
+        measmod1 = info1["measurement_model"]
+        filtsmooth.merge_regression_problems((prob1, measmod1), (prob2, good_input))
+
+    # Single models are rejected (2nd input)
+    with pytest.raises(TypeError):
+        measmod2 = info2["measurement_model"]
+        filtsmooth.merge_regression_problems((prob1, good_input), (prob2, measmod2))
```
schmidtjonathan, line 71: Spontaneous thought: what do you think about the `merge_regression_problems` method repeating measurement models according to the `RegressionProblem.locations` (like `.filter()` does)?

schmidtjonathan:
Thanks for the work you put into this! I made some minor suggestions, but from my side, this can be merged :)

src/probnum/filtsmooth/_utils.py
schmidtjonathan: > 1. an or would be better than an and here, right?

For readability, yes. Otherwise it's equivalent.

> 2. 

Yep, that suggestion was just out of personal preference, to avoid if-branches if possible. But it makes no big difference, that's true :) 
We can leave it as is, if you prefer!

src/probnum/filtsmooth/_particlefiltsmooth/_importance_distributions.py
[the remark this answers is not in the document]
schmidtjonathan: Ok, as long as we are not walking too close to the edge of machine precision, it's fine :)

src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
schmidtjonathan: Yep, let's handle that in a separate PR

src/probnum/filtsmooth/_utils.py
pnkraemer: I mostly share your sentiment as to avoiding if branches, but I am mostly against nested ifs. Here, one check after the other seems fine to me. I will leave it, if you dont mind ;)

tests/test_filtsmooth/test_utils.py
pnkraemer: that could work, but I am not sure whether it is necessary. What do you think? It would be easily done

src/probnum/problems/_problems.py
```diff
@@ -1,19 +1,19 @@
...
-import probnum.type as pntp
-from probnum import randvars
+from probnum import linops, randvars
+from probnum.type import FloatArgType
 
 
 @dataclasses.dataclass
-class RegressionProblem:
-    r"""Regression problem.
+class TimeSeriesRegressionProblem:
```
pnkraemer, line 15: I think renaming this is a good idea, because (i) it is the argument type that is passed to Kalman filtering and particle filtering, and (ii) all the functions (e.g. merge_..., or filter()) rely on something iterable. This only works for time series

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -23,6 +23,7 @@ def car_tracking(
     initrv: Optional[randvars.RandomVariable] = None,
     forward_implementation: str = "classic",
     backward_implementation: str = "classic",
+    random_state: Optional[RandomStateArgType] = None,
```
pnkraemer, line 26: this makes it easier to make tests less random

src/probnum/statespace/transition.py
```diff
@@ -50,6 +50,9 @@ def __init__(self, input_dim: IntArgType, output_dim: IntArgType):
         self.input_dim = input_dim
         self.output_dim = output_dim
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}(input_dim={self.input_dim}, output_dim={self.output_dim})"
+
```
pnkraemer, line 55: good for (1) doctests, (2) usability/debugability

tests/test_diffeq/test_odefiltsmooth/test_ivpfiltsmooth.py
```diff
@@ -101,5 +101,5 @@ def test_convergence_error(ivp, algo_order):
     # Non-strict rtol, bc this test is flaky by construction
     # As long as rtol < 1., this test seems meaningful.
     np.testing.assert_allclose(
-        err_small_step, expected_decay * err_large_step, rtol=0.9
+        err_small_step, expected_decay * err_large_step, rtol=0.95
```
pnkraemer, line 104: made this test a little less sharp

tests/test_filtsmooth/test_gaussfiltsmooth/_linearization_test_interface.py
```diff
@@ -0,0 +1,168 @@
+"""Test interface for EKF and UKF."""
```
pnkraemer, line 1: this function is now the pytest version of the former `filtsmooth_testcases`.

tests/test_filtsmooth/test_particlefiltsmooth/test_particle_filter.py
```diff
@@ -30,7 +32,7 @@ def test_effective_number_of_events():
 
 @pytest.fixture
 def num_particles():
-    return 10
+    return 20
```
pnkraemer, line 35: I think we can afford this, and it makes the tests a little more reliable

src/probnum/filtsmooth/_gaussfiltsmooth/_extendedkalman.py
```diff
@@ -135,16 +135,16 @@ def __init__(
         non_linear_model,
         mde_atol=1e-5,
         mde_rtol=1e-5,
-        mde_solver="LSODA",
+        mde_solver="RK45",
```
pnkraemer, line 138: the error messages from LSODA are super cryptic (because it wraps fortran code).

schmidtjonathan: Makes sense.

pnkraemer:
@schmidtjonathan let's go! :)

tests/test_filtsmooth/test_gaussfiltsmooth/_linearization_test_interface.py
schmidtjonathan: Ah, yes, thanks for taking care of this!

src/probnum/problems/_problems.py
```diff
@@ -24,27 +24,80 @@ class RegressionProblem:
...
+
+        1. Wrap the measurement models into an iterable of measurement models (by
+        default, a list).
+        2. Check that all inputs have the same length.
+        """
+
+        # If a single measurement model has been provided, transform it into a list of models
+        # to ensure that there is a measurement model for every (t, y) combination.
+        if not isinstance(self.measurement_models, abc.Iterable):
+            self.measurement_models = [self.measurement_models] * len(self.locations)
```
schmidtjonathan, line 76: Should we notify the user about this? Or are we fine with this side-effect just happening in the background? (asking just to be sure, I think it's fine, since we did it before, as well, just in the `filter` call

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -450,15 +517,23 @@ def l(t):
...
     )
-    regression_problem = problems.RegressionProblem(
-        observations=obs, locations=time_grid, solution=states
+    prior_process = randprocs.MarkovProcess(
+        transition=dynamics_model, initrv=initrv, initarg=time_grid[0]
     )
 
     statespace_components = dict(
         dynamics_model=dynamics_model,
         measurement_model=measurement_model,
```
schmidtjonathan, line 534: We don't have to return the measurement model anymore, right?

schmidtjonathan: Since it's part of the `RegressionProblem`

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
```diff
@@ -550,16 +631,21 @@ def logistic_ode(
...
+        measurement_models=measurement_model,
         solution=solution,
     )
 
+    prior_process = randprocs.MarkovProcess(
+        transition=dynamics_model, initrv=initrv, initarg=time_grid[0]
+    )
     statespace_components = dict(
         dynamics_model=dynamics_model,
         measurement_model=measurement_model,
```
schmidtjonathan, line 646: see above

schmidtjonathan (approved):
Made some minor comments, but -- from my side -- can be merged. Thanks for your effort!

src/probnum/problems/_problems.py
pnkraemer: I think for now we leave it. If eventually, we decide to deprecate the non-iterable argument, we can still do it then?

src/probnum/problems/zoo/filtsmooth/_filtsmooth_problems.py
pnkraemer: we don't even have to return initrv and dynamics model. Shall I remove all of these, too?

schmidtjonathan: 👍

PR #401, opened 2021-05-04
Return forwarded RV

schmidtjonathan:
Where applicable, let `Transition`s return forwarded random variables in the `.backward()` call. Might be useful e.g. for computing marginal likelihoods during filtering/smoothing.

### DIFF
--- a/src/probnum/statespace/integrator.py
+++ b/src/probnum/statespace/integrator.py
@@ -292,9 +292,6 @@ class IBM(Integrator, sde.LTISDE):
             t=t,
             _diffusion=_diffusion,
         )
-        # assert info is empty. Otherwise, we need to change
-        # things in info in which case we want to be warned.
-        assert not info
 
         return _apply_precon(self.precon(dt), rv), info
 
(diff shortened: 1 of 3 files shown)


src/probnum/statespace/integrator.py
```diff
@@ -294,9 +294,6 @@ def backward_rv(
             t=t,
             _diffusion=_diffusion,
         )
-        # assert info is empty. Otherwise, we need to change
-        # things in info in which case we want to be warned.
-        assert not info
```
schmidtjonathan, line 299: @pnkraemer Correct me if I'm wrong, but we can safely remove those assertions, right?
Because `gain` is not `None` and `forwarded_rv` is not `None`, thus the `forwarded_rv` that we pass into `backward` is then just returned within the `info` dict.

pnkraemer: I can look into this in more details tomorrow, but the central question will be: does everything behave well with preconditioning? Because this is why we have those asserts there (with the transition forward/backward PR backaday, the effect on preconditioning on the elements in info was too much for my brain...

pnkraemer: Hmmm I am not entirely sure, but it seems to me that we should leave this in (because once we start returning a backward info, we will need to be careful at this stage of the code with regard to not introducing preconditioning errors). Do you see what I mean?

If you don't like the assert, we can easily change this to something along the lines of
```python
if not info:
    raise ValueError("backward_rv returns a non-empty info, which does not take preconditioning into account (yet).")
```

schmidtjonathan: But `info` is not empty anymore when returning the `meas_rv`. So this will fail.

pnkraemer: aaah okay in this case get rid of it :)

pnkraemer (approved):

Issue #388, opened 2021-04-28
Batches of random variables


pnkraemer:
As discussed offline a few weeks ago, the `RandomVariable` implementation would benefit from allowing batches.

A batch would be a collection of independent random variables of a common type, for instance a batch of multivariate Normal distributions. Currently, this case is handled in `_RandomVariableList`, which was intended as a temporary solution (hence the underscore).
Having such a batch of, e.g., Normals would allow storing mean and covariance efficiently, which in turn will speed up a lot of computation that uses lists of Normal objects (currently, this is used heavily in e.g. `KalmanPosterior` objects -- sampling for instance would be so much faster!).

A solution would be that each random variable has both, a `batch_shape` and an `event_shape`. This is how Tensorflow Probability handles it (https://github.com/tensorflow/probability/blob/master/tensorflow_probability/examples/jupyter_notebooks/Understanding_TensorFlow_Distributions_Shapes.ipynb).

Perhaps it would be a good idea to change for instance `.reshape()` into `.batch_reshape` and `.event_reshape()`. 
What I am a little bit worried about is implementing random variable arithmetic and broadcasting. Would that even be possible/recommended? What would broadcasting even refer to? A batch-broadcast, an event-broadcast, or a batch-and-event-broadcast?

JonathanWenger:
This is a duplicate of #428

PR #403, opened 2021-05-05
Build FilteringPosterior on the fly

schmidtjonathan:
Builds the filtering posterior on the fly, i.e. directly builds up the `FilteringPosterior` object, which is returned in the end.
This happens by
- allowing for initializing an empty `TimeSeriesPosterior` object
- adding an `.append(loc, rv)` method to the `TimeSeriesPosterior` class

### DIFF
--- a/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
+++ b/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
@@ -347,9 +347,9 @@ class GaussianIVPFilter(ODESolver):
         """Create an ODESolution object."""
 
         kalman_posterior = filtsmooth.FilteringPosterior(
-            times,
-            rvs,
-            self.prior_process.transition,
+            transition=self.prior_process.transition,
+            locations=times,
+            states=rvs,
             diffusion_model=self.diffusion_model,
         )
 
--- a/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
+++ b/src/probnum/diffeq/odefiltsmooth/ivpfiltsmooth.py
@@ -360,23 +360,27 @@ class GaussianIVPFilter(ODESolver):
         locations = odesol.kalman_posterior.locations
         rv_list = odesol.kalman_posterior.states
 
-        if self._calibrate_all_states_post_hoc:
-            # Constant diffusion model is the only way to go here.
-            s = self.diffusion_model.diffusion
+        kalman_posterior = filtsmooth.FilteringPosterior(
+            transition=self.prior_process.transition,
+            diffusion_model=self.diffusion_model,
+        )
 
-            for idx, (t, rv) in enumerate(zip(locations, rv_list)):
-                rv_list[idx] = randvars.Normal(
+        # Constant diffusion model is the only way to go here.
+        s = (
+            self.diffusion_model.diffusion
+            if self._calibrate_all_states_post_hoc
+            else 1.0
+        )
+
+        for idx, (t, rv) in enumerate(zip(locations, rv_list)):
+            kalman_posterior.append(
+                location=t,
+                state=randvars.Normal(
                     mean=rv.mean,
                     cov=s * rv.cov,
                     cov_cholesky=np.sqrt(s) * rv.cov_cholesky,
-                )
-
-        kalman_posterior = filtsmooth.FilteringPosterior(
-            locations,
-            rv_list,
-            self.prior_process.transition,
-            diffusion_model=self.diffusion_model,
-        )
+                ),
+            )
 
         if self.with_smoothing is True:
 
@@ -385,10 +389,10 @@ class GaussianIVPFilter(ODESolver):
                 rv_list, locations, _diffusion_list=squared_diffusion_list
             )
             kalman_posterior = filtsmooth.SmoothingPosterior(
-                locations,
-                rv_list,
-                self.prior_process.transition,
                 filtering_posterior=kalman_posterior,
+                transition=self.prior_process.transition,
+                locations=locations,
+                states=rv_list,
                 diffusion_model=self.diffusion_model,
             )
 
--- a/src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
+++ b/src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
@@ -198,7 +198,7 @@ class Kalman(BayesFiltSmooth):
         TimeSeriesRegressionProblem: a regression problem data class
         """
 
-        posterior = FilteringPosterior(transition=self.dynamics_model)
+        posterior = FilteringPosterior(transition=self.prior_process.transition)
         info_dicts = []
 
         for t, rv, info in self.filtered_states_generator(
@@ -270,7 +270,7 @@ class Kalman(BayesFiltSmooth):
                 realization_obtained=data, rv=curr_rv, _linearise_at=linearise_update_at
             )
 
-            yield curr_rv, info_dict
+            yield t, curr_rv, info_dict
             t_old = t
 
     def smooth(self, filter_posterior, _previous_posterior=None):
@@ -295,7 +295,7 @@ class Kalman(BayesFiltSmooth):
 
         return SmoothingPosterior(
             filtering_posterior=filter_posterior,
-            transition=self.dynamics_model,
+            transition=self.prior_process.transition,
             locations=filter_posterior.locations,
             states=rv_list,
         )
--- a/src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
+++ b/src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
@@ -201,10 +201,10 @@ class Kalman(BayesFiltSmooth):
         posterior = FilteringPosterior(transition=self.dynamics_model)
         info_dicts = []
 
-        for idx, (rv, info) in enumerate(
-            self.filtered_states_generator(regression_problem, _previous_posterior)
+        for t, rv, info in self.filtered_states_generator(
+            regression_problem, _previous_posterior
         ):
-            posterior.append(location=regression_problem.locations[idx], state=rv)
+            posterior.append(location=t, state=rv)
             info_dicts.append(info)
 
         return posterior, info_dicts
--- a/src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
+++ b/src/probnum/filtsmooth/_gaussfiltsmooth/_kalman.py
@@ -198,20 +198,14 @@ class Kalman(BayesFiltSmooth):
         TimeSeriesRegressionProblem: a regression problem data class
         """
 
-        filtered_rvs = []
+        posterior = FilteringPosterior(transition=self.dynamics_model)
         info_dicts = []
 
-        for rv, info in self.filtered_states_generator(
-            regression_problem, _previous_posterior
+        for idx, (rv, info) in enumerate(
+            self.filtered_states_generator(regression_problem, _previous_posterior)
         ):
-            filtered_rvs.append(rv)
+            posterior.append(location=regression_problem.locations[idx], state=rv)
             info_dicts.append(info)
-
-        posterior = FilteringPosterior(
-            locations=regression_problem.locations,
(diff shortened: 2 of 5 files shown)


src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -36,9 +36,30 @@ class TimeSeriesPosterior(abc.ABC):
...
-    def __init__(self, locations: np.ndarray, states: np.ndarray) -> None:
-        self.locations = np.asarray(locations)
-        self.states = _randomvariablelist._RandomVariableList(states)
+    def __init__(
+        self,
+        locations: Optional[Iterable[FloatArgType]] = None,
+        states: Optional[Iterable[randvars.RandomVariable]] = None,
+    ) -> None:
+        self._locations = list(locations) if locations is not None else []
+        self._states = list(states) if states is not None else []
```
schmidtjonathan, line 45: @pnkraemer What do you think of this? This way, we save ourselves an additional `initialize` method that might add unnecessary complexity

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
```diff
@@ -42,9 +42,9 @@ class KalmanPosterior(TimeSeriesPosterior, abc.ABC):
 
     def __init__(
         self,
-        locations: np.ndarray,
-        states: _randomvariablelist._RandomVariableList,
         transition: GaussMarkovPriorTransitionArgType,
+        locations: Optional[Iterable[FloatArgType]] = None,
+        states: Optional[Iterable[randvars.RandomVariable]] = None,
     ) -> None:
```
schmidtjonathan, line 48: @pnkraemer Do you approve of this change of the ordering of the arguments?
This is necessary in order to also create empty `FilteringPosterior` objects. Since non-optional arguments may not follow optional ones. And the transition **always** has to be provided. Even for empty posteriors.

schmidtjonathan:
Also, do we even need another method? As is now, we provide
- `.filter()` for when you want just the end result of the Kalman filter on the entire time period, which now builds on-the-fly, but gains no novel functionalities from this PR
- `filtered_states_generator()` which I did not touch in this PR, that can be wrapped from outside, in case somebody wants to investigate their posterior in every step.

So, somebody who wants to create e.g. an animation, could do sth like this:

```python
posterior = FilteringPosterior(transition=self.dynamics_model)

for idx, (rv, _) in enumerate(
  self.filtered_states_generator(regression_problem, _previous_posterior)
):
  posterior.append(location=regression_problem.locations[idx], state=rv)
  plot_posterior(posterior)
```
, right?

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: Yes, I could live with that ;) However, do you think it would improve readability even more if at initialisation one would (need to) pass the empty lists? This might make it even more obvious to a code reader what happens, but that is just a guess. What do you think?

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: yes, I am fine with that. When I mentioned the `initialise` methods, I meant one in the filter (not in the posterior). See also my other comment :)

src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -36,9 +36,30 @@ class TimeSeriesPosterior(abc.ABC):
...
+        self._locations.append(location)
+        self._states.append(state)
+
+    @property
+    def locations(self):
+        return np.asarray(self._locations)
+
+    @property
+    def states(self):
+        return _randomvariablelist._RandomVariableList(self._states)
```
mmahsereci, line 62: this accesses a private class.

src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -36,9 +36,30 @@ class TimeSeriesPosterior(abc.ABC):
...
+    ) -> None:
+        self._locations = list(locations) if locations is not None else []
+        self._states = list(states) if states is not None else []
+
+    def append(
+        self,
+        location: FloatArgType,
+        state: randvars.RandomVariable,
+    ) -> None:
+        # TODO: Type-checks here?
```
mmahsereci, line 52: Is this TODO still open?

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: see the "hence the underscore" bit in #388

mmahsereci: thanks

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
schmidtjonathan: I personally don't think that it would improve readability, or ease-of-use. But that's personal taste :)

src/probnum/filtsmooth/timeseriesposterior.py
schmidtjonathan: Yep.
Technically, we should check the types here, because it wouldn't throw an error if somebody appends all kidns of stuff to those lists.
But this method is probably invoked very often during filtering. So, I wanted to raise the question whether or not we want type checks here?
Safety vs. speed, yet again.

schmidtjonathan: What do you think, @mmahsereci , @pnkraemer ?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -189,21 +189,15 @@ def filter(
...
         info_dicts = []
 
-        for rv, info in self.filtered_states_generator(
-            regression_problem, _previous_posterior
+        for idx, (rv, info) in enumerate(
+            self.filtered_states_generator(regression_problem, _previous_posterior)
         ):
-            filtered_rvs.append(rv)
+            posterior.append(location=regression_problem.locations[idx], state=rv)
             info_dicts.append(info)
```
pnkraemer, line 200: do you think the generator should yield t as well? This enumerate seems a bit awkward here, doesnt it?

src/probnum/filtsmooth/gaussfiltsmooth/kalmanposterior.py
pnkraemer: alright, then not :)

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: I think its fine like this. An extra type check seems a bit unnecessary to me...

schmidtjonathan: Ok thanks to both of you :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: Done

schmidtjonathan:
@pnkraemer I resolved what we discussed. But we should discuss whether we should merge this PR **after** merging #396 . I would say _yes_, because this here has way less changes and might be less confusing to integrate.

pnkraemer:
That would be wonderful. Thank you for suggesting that.

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
```diff
@@ -265,7 +259,7 @@ def filtered_states_generator(
                 _linearise_update_at=_linearise_update_at,
             )
 
-            yield filtrv, info_dict
+            yield times[idx], filtrv, info_dict
```
pnkraemer, line 262: sorry if this has been explained before.
Can you please remind me what the reason for returning the time value here was? It seems like this value would be accessible outside of the generator, but I might be missing something.

src/probnum/filtsmooth/timeseriesposterior.py
```diff
@@ -36,9 +36,29 @@ class TimeSeriesPosterior(abc.ABC):
...
+        state: randvars.RandomVariable,
+    ) -> None:
+        self._locations.append(location)
+        self._states.append(state)
+
+    @property
+    def locations(self):
+        return np.asarray(self._locations)
+
+    @property
```
pnkraemer, line 59: do you think it makes sense to cache those? And does it also make sense to use (i) check in the append() whether the location is right of the last location (ii) add a finalise() method that blocks the append() for the future? I am not sure whether these things are good or bad ideas, but would like to hear your opinion :)

pnkraemer (commented):
I have two minor questions reg. the design of the functions. Mostly a brainstorm. When these  things are resolved (and the branch is up to date again), we can merge :)

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
schmidtjonathan: It is returned such that when building the `FilteringPosterior` on the fly, we can call `posterior.append(location=t, state=rv)`.
Do you think there's a more elegant solution?

src/probnum/filtsmooth/timeseriesposterior.py
schmidtjonathan: > do you think it makes sense to cache those? 

I don't know what happens then when the object is altered (say, by appending another location+state)

(i): 
Mh technically, I would agree. But isn't that checked elsewhere? Resp., does the algorithm, which calls this method, even allow this?

(ii)
Absolutely, good idea. I'll add that. Do we want another class (e.g. `FrozenFilteringPosterior` ) for that? Or just a flag?

src/probnum/filtsmooth/gaussfiltsmooth/kalman.py
pnkraemer: not sure. I guess the alternative would be to include zip over `regression_problem.locations, filter_generator(regression_problem)` but (a) this looks awkward, and (b) the current implementation *might* be more robust w.r.t. multiple measurements at a single locations.

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: > I don't know what happens then when the object is altered (say, by appending another location+state)
I guess it remains cached. But I am not 100% sure (though, isn't the idea of caching to not check these things anymore?)

(I) the method does not do this, but what happens if a user randomly decides to append a new measurement. This would be possible (though stupid? I don't know). I don't think it is really checked, but perhaps there are some weird errors. I don't know. I think it would be reasonable to check.

(ii) I think a flag would work, unless the FrozenFilteringPosterior makes the code more elegant. Do what you consider most simple :)

schmidtjonathan: I used a flag now, since I didn't find an elegant way to transfer the changes of a `FrozenTimeSeriesPosterior` to subclasses of `TimeSeriesPosterior` , like `FilteringPosterior` for instance. Before creating a hundred subclasses, I chose the easier way. Although it would have been nice to convert the property into a cached_property after freezing.

schmidtjonathan:
@pnkraemer Ready for review.

src/probnum/filtsmooth/timeseriesposterior.py
pnkraemer: Alright! 

> Although it would have been nice to convert the property into a cached_property after freezing

Technically, one can always manually cache it :) But I think it is okay without

schmidtjonathan: Only if you know beforehand how many states/locations you will add. But that's a good point, we could make that optionally possible. Shall we?

Sth like
```python
def __init__(..., total=None):
  if total is not None:
    self.locations = np.empty((total, ))
  ...
  # etc.
```

schmidtjonathan: Don't know how much we gain by this, though. I suspect not much, tbh

pnkraemer (approved):

PR #404, opened 2021-05-05
Nail down stability of IBM with very high orders.

pnkraemer:
When playing with the code, I noticed that the changes in #386 were absolutely crucial for stability of high order IBMs.
I thought it would be good to have a test that nails this down. The changes in this PR are:

1) add an optional mini-damping factor for IBM covariances: this way (using 1e-15), the maximum order for IBM kalman filtering is not 10 anymore, but 15! :D If we do crazy small steps, the max order is "only" 14 now. @nathanaelbosch 
2) write tests for the above point, as well as the no-damping limit of 10.
3) made mini updates in the car tracking problem zoo that improve its square-root usage in a very minor way.

### DIFF
--- a/src/probnum/statespace/discrete_transition.py
+++ b/src/probnum/statespace/discrete_transition.py
@@ -501,10 +501,10 @@ class DiscreteLTIGaussian(DiscreteLinearGaussian):
         super().__init__(
             input_dim,
             output_dim,
-            lambda t: state_trans_mat,
-            lambda t: shift_vec,
-            lambda t: proc_noise_cov_mat,
-            lambda t: proc_noise_cov_cholesky,
+            state_trans_mat_fun=lambda t: state_trans_mat,
+            shift_vec_fun=lambda t: shift_vec,
+            proc_noise_cov_mat_fun=lambda t: proc_noise_cov_mat,
+            proc_noise_cov_cholesky_fun=lambda t: proc_noise_cov_cholesky,
             forward_implementation=forward_implementation,
             backward_implementation=backward_implementation,
         )
--- a/src/probnum/statespace/integrator.py
+++ b/src/probnum/statespace/integrator.py
@@ -168,7 +168,6 @@ class IBM(Integrator, sde.LTISDE):
         spatialdim,
         forward_implementation="classic",
         backward_implementation="classic",
-        _process_noise_damping=0.0,
     ):
         # initialise BOTH superclasses' inits.
         # I don't like it either, but it does the job.
@@ -181,8 +180,6 @@ class IBM(Integrator, sde.LTISDE):
             forward_implementation=forward_implementation,
             backward_implementation=backward_implementation,
         )
-
-        self._process_noise_damping = _process_noise_damping
 
     @cached_property
     def _driftmat(self):
@@ -219,10 +216,7 @@ class IBM(Integrator, sde.LTISDE):
         process_noise = np.kron(np.eye(self.spatialdim), process_noise_1d)
         empty_shift = np.zeros(self.spatialdim * (self.ordint + 1))
 
-        val = self._process_noise_damping
-        process_noise_cholesky_1d = np.linalg.cholesky(
-            process_noise_1d + val * np.eye(self.ordint + 1)
-        )
+        process_noise_cholesky_1d = np.linalg.cholesky(process_noise_1d)
         process_noise_cholesky = np.kron(
             np.eye(self.spatialdim), process_noise_cholesky_1d
         )
--- a/src/probnum/statespace/integrator.py
+++ b/src/probnum/statespace/integrator.py
@@ -202,42 +202,39 @@ class IBM(Integrator, sde.LTISDE):
 
     @cached_property
     def equivalent_discretisation_preconditioned(self):
-        """Discretised IN THE PRECONDITIONED SPACE."""
-        empty_shift = np.zeros(self.spatialdim * (self.ordint + 1))
-        process_noise_cov_cholesky = np.linalg.cholesky(
-            self._proc_noise_cov_mat
-            + self._process_noise_damping * np.eye(len(self._proc_noise_cov_mat))
+        """Discretised IN THE PRECONDITIONED SPACE.
+
+        The preconditioned state transition is the flipped Pascal matrix.
+        The preconditioned process noise covariance is the flipped Hilbert matrix.
+        The shift is always zero.
+
+        Reference: https://arxiv.org/abs/2012.10106
+        """
+
+        state_transition_1d = np.flip(
+            scipy.linalg.pascal(self.ordint + 1, kind="lower", exact=False)
         )
+        state_transition = np.kron(np.eye(self.spatialdim), state_transition_1d)
+        process_noise_1d = np.flip(scipy.linalg.hilbert(self.ordint + 1))
+        process_noise = np.kron(np.eye(self.spatialdim), process_noise_1d)
+        empty_shift = np.zeros(self.spatialdim * (self.ordint + 1))
+
+        val = self._process_noise_damping
+        process_noise_cholesky_1d = np.linalg.cholesky(
+            process_noise_1d + val * np.eye(self.ordint + 1)
+        )
+        process_noise_cholesky = np.kron(
+            np.eye(self.spatialdim), process_noise_cholesky_1d
+        )
+
         return discrete_transition.DiscreteLTIGaussian(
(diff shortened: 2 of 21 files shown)


src/probnum/statespace/integrator.py
```diff
@@ -201,11 +204,15 @@ def _dispmat(self):
     def equivalent_discretisation_preconditioned(self):
         """Discretised IN THE PRECONDITIONED SPACE."""
         empty_shift = np.zeros(self.spatialdim * (self.ordint + 1))
+        process_noise_cov_cholesky = np.linalg.cholesky(
+            self._proc_noise_cov_mat
+            + self._process_noise_damping * np.eye(len(self._proc_noise_cov_mat))
+        )
```
schmidtjonathan: Is this nugget necessary for numerical stability?

src/probnum/statespace/discrete_transition.py
```diff
@@ -501,10 +501,10 @@ def __init__(
...
             input_dim,
             output_dim,
-            lambda t: state_trans_mat,
-            lambda t: shift_vec,
-            lambda t: proc_noise_cov_mat,
-            lambda t: proc_noise_cov_cholesky,
+            state_trans_mat_fun=lambda t: state_trans_mat,
+            shift_vec_fun=lambda t: shift_vec,
+            proc_noise_cov_mat_fun=lambda t: proc_noise_cov_mat,
+            proc_noise_cov_cholesky_fun=lambda t: proc_noise_cov_cholesky,
```
schmidtjonathan, line 507: 👍 Thanks! I always get a bit twitchy when more than 2 arguments are passed positionally

src/probnum/statespace/integrator.py
```diff
@@ -319,10 +322,19 @@ def discretise(self, dt):
             @ self.precon(dt).T
         )
         zero_shift = np.zeros(len(state_trans_mat))
+
+        # The Cholesky factor of the process noise covariance matrix of the IBM
+        # always exists, even for non-square root implementations.
+        proc_noise_cov_cholesky = (
+            self.precon(dt)
+            @ self.equivalent_discretisation_preconditioned.proc_noise_cov_cholesky
+        )
```
schmidtjonathan, line 331: Do we get any benefit from computing this Cholesky factor when not using the `sqrt` implementation?
Otherwise, we might make this conditional.

src/probnum/statespace/integrator.py
pnkraemer: only for orders > 11 :)

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -50,3 +50,127 @@ def test_info_dicts(setup):
...
+def test_kalman_smoother_high_order_ibm_with_process_noise_damping_crazy_small_steps():
+    """If we add a small damping factor to the computation of the Cholesky factor of the
+    process noise covariance, we can do order 15 on the car tracking model at least."""
+    regression_problem, statespace_components = filtsmooth_zoo.car_tracking(
+        model_ordint=14,
+        timespan=(0.0, 1e-10),
+        step=1e-12,
+        forward_implementation="sqrt",
+        backward_implementation="sqrt",
+        _process_noise_damping=1e-14,
```
schmidtjonathan, line 97: I am surprised that a nugget of `1e-14` does anything at all.

Do you think it necessary to establish an entirely new kwarg, solely for the case that somebody wants to use a 14 times integrated model? 😄

src/probnum/statespace/integrator.py
pnkraemer: Hmmm my thinking here was that this cholesky factor is created for the equivalent discretisation (preconditioned), so it was super weird not to translate it here, too. What do you think?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
```diff
@@ -50,3 +50,127 @@ def test_info_dicts(setup):
...
+    obs_rmse = np.mean(np.abs(regression_problem.observations - truth[:, :2]))
+
+    assert smooms_rmse < filtms_rmse < obs_rmse
+
+
+def test_kalman_filter_high_order_ibm_with_process_noise_damping():
+    """With damping, the filter achieves really high orders (we test 25, but 50 was
+    possible too, for some reason)."""
+    regression_problem, statespace_components = filtsmooth_zoo.car_tracking(
+        model_ordint=25,
```
schmidtjonathan, line 154: What is happening.

src/probnum/statespace/integrator.py
schmidtjonathan: Could we make the computation of the Cholesky factor - both in `equivalent_discretisation_preconditioned` and in `discretise` - conditional on `self.forward`/`backward_implementation`?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
pnkraemer: Just for the jokes, the test passes for order=75 as well (no typo!), but I thought 25 was sufficiently extreme :D

src/probnum/statespace/integrator.py
pnkraemer: yes, we could. Thought it only happens once, so the impact on runtime should be almost neglectible. Do you think that it is worth it nevertheless?

schmidtjonathan:
Just checking: Your PR description is outdated, right?

> the maximum order for IBM kalman filtering is not 10 anymore, but 15!

this should be 25 by now, right? I don't want to be petty here, just check if the 25-order thing is not some kind of special case. 😆

pnkraemer:
25 is only filter. filtsmooth starts complaining at 15

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
schmidtjonathan: All just by removing the matrix inversion, right ?

schmidtjonathan:
> 25 is only filter. filtsmooth starts complaining at 15

I see. Still. That's quite remarkable

src/probnum/statespace/integrator.py
schmidtjonathan: Should we then actually add this argument? I guess it will be needed quite rarely, if not only for the test you added in this PR 😄 Correct me if this is wrong

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
pnkraemer: Well, I dont want to stretch the word "necessary" too much here... :) I just thought it was cool to get these high orders! ;) 
And it is a "hidden" kwarg that defaults to 0, so I thought it doesnt hurt too much. Though if you find it annoying I am open for talking about it :)

pnkraemer:
as I said in one of the comments as well: 75 worked, too, but I thought no one would believe me so I used 25 for the test ;)

src/probnum/statespace/integrator.py
schmidtjonathan: Can be ignored, I read your answer below

pnkraemer: no, it is only for a bunch of extreme cases (would we be the only ODE solver of order 15 in python then :thinking:?) I will add a sentence into the docstring.

pnkraemer: no, not really. I was playing around with it and thought it was cool, but I see your concerns. Shall I remove it?

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
pnkraemer: and by using this mini damping factor that makes the process noise always positive definite.

tests/test_filtsmooth/test_gaussfiltsmooth/test_kalman.py
schmidtjonathan: Well maybe I am biased, but I quickly get confused when a method offers me to tweak every part of the method, and especially if the documentation takes half an hour to read (not that this would be the case here!) So I often prefer clarity over control, given that often, the kwargs are only used in very special cases. Do hidden kwargs appear in the docs?

schmidtjonathan (approved):
Crazy stuff! Thanks for working that out :)

PR #410, opened 2021-05-07
Abstract base class `ProbabilisticNumericalMethod`

### LINKED ISSUE #280: Write an abstract interface `ProbabilisticNumericalMethod`
Define in an abstract way how implementations of probabilistic numerical methods look like. This enables uniformity across methods and forces us to define what PN methods look like and are.

```python
import abc


class ProbabilisticNumericalMethod(abc.ABC):
    def __init__(
        prior: Union[RandomVariable, RandomProcess]],
        action_rule: ActionRule,
        observation_process: ObservationProcess,
        belief_update: BeliefUpdate,
        stopping_criterion: StoppingCriterion,
        hyperparam_opt: HyperparameterOptimization,
    ):
        raise NotImplementedError

    def infer_solution(
        self, problem: Problem
    ) -> Tuple[Collection[Union[RandomVariable, RandomProcess]], Dict]:
        raise NotImplementedError
```

Beyond this implementation an convenience function exists which follows the NumPy/SciPy function signatures as described in https://probnum.readthedocs.io/en/latest/development/implementing_a_probnum_method.html#Interface.

JonathanWenger:
This PR introduces an abstract base class `ProbabilisticNumericalMethod` which should be implemented by any of the PN method implementations. This base class can be extended in the future with components such as
- solve_iter
- policy
- stopping criteria
- hyperparameter optimization
- ...

In order for `probnum` to appear in the API section of the docs, the docs header structure was updated as well and doubled docstrings in the docs and files were removed.

## Minor changes
- Removed `RandomVariable` from namespace `probnum`

This closes #280


src/probnum/_probabilistic_numerical_method.py
```diff
@@ -0,0 +1,53 @@
...
+    All PN methods should subclass this base class. Typically convenience functions
+    (such as :meth:`~probnum.linalg.problinsolve`) will instantiate an object of a
+    derived subclass.
+    """
+
+    def __init__(self, prior: BeliefType):
+        self.prior = prior
+
+    @abstractmethod
+    def solve(
```
pnkraemer, line 42: what happened to `infer_solution`? :) I thought it had some elegance to it. Any particular reason you went back to solve? (If I remember correctly, we decided against solve back in winter, because it sounds a bit weird to solve a quadrature problem.) Also, `infer_solution` also teaches a user about PN, just from the function name! :)

pnkraemer (commented):
Really cool! Thanks for working on this Jonathan! :) 

I will have a more thorough look later (maybe on monday?), but wanted to leave one question in the code already :)

src/probnum/_probabilistic_numerical_method.py
JonathanWenger: I realize this is what your comment says in #280, but I actually thought that we agreed on solve in the verbal discussion afterwards :grin:, because in particular for linear systems you don't always just infer the solution but also other quantities of interest (i.e. inverse, rhs, etc.). Also for quadrature I find "solving an integral/integration problem/quadrature problem" pretty natural.

As with all things naming, how about we finish the PR, put it on the agenda for the software meeting and then rename the method, once we all agree?

JonathanWenger:
Regarding test coverage, I refactored a part of the library which is not well tested `probnum.linalg` which now shows poor test coverage for this PR. This part will be removed and much better tested in a future PR, so I will not invest time in writing arbitrary tests which will be removed soon anyway. I hope this is okay @pnkraemer.

src/probnum/_probabilistic_numerical_method.py
pnkraemer: Sure, that is fine by me. I was just wondering :)

docs/source/development/adding_to_the_api_documentation.ipynb
```diff
@@ -71,7 +70,10 @@
     "    callback: Optional[Callable] = None,\n",
     "    **kwargs\n",
     ") -> Tuple[\n",
-    "    \"probnum.RandomVariable\", \"probnum.RandomVariable\", \"probnum.RandomVariable\", Dict\n",
+    "    \"randvars.RandomVariable\",\n",
+    "    \"randvars.RandomVariable\",\n",
+    "    \"randvars.RandomVariable\",\n",
+    "    Dict,\n",
```
pnkraemer, line 76: wow, readable notebook diff! Did you do anything special that I should adapt?

docs/source/development/adding_to_the_api_documentation.ipynb
```diff
@@ -71,7 +70,10 @@
...
     "]:\n",
     "    \"\"\"Infer a solution to the linear system :math:`A x = b` in a Bayesian framework.\n",
```
pnkraemer, line 78: ...just pointing out this formulation reg. the `solve/infer_solution` discussion :laughing:

docs/source/public_api/linops.rst
```diff
@@ -1,19 +1,7 @@
...
-This package implements finite dimensional linear operators. It can be used to
-represent linear maps between finite-dimensional vector spaces without explicitly
-constructing their matrix representation in memory. This is particularly useful for
-sparse and structured matrices and often allows for the definition of a more
-efficient matrix-vector product. Linear operators support common algebraic
-operations, including matrix-vector products, addition, multiplication, and
-transposition.
-
-Several algorithms in the :mod:`probnum.linalg` subpackage are able to operate on
-:class:`~probnum.linops.LinearOperator` instances.
```
pnkraemer, line 15: why is this text removed?

pnkraemer: aah is it because it was somehow doubled? (you mention something about this in the PR description)

src/probnum/__init__.py
```diff
@@ -44,7 +45,17 @@
     statespace,
     utils,
 )
-from .randvars import RandomVariable, asrandvar
+from ._probabilistic_numerical_method import ProbabilisticNumericalMethod
+from .randvars import asrandvar
+
+# Public classes and functions. Order is reflected in documentation.
+__all__ = [
+    "asrandvar",
```
pnkraemer, line 53: is there a reason that this method remains in the probnum namespace, even though all the random variables do not?

src/probnum/_probabilistic_numerical_method.py
```diff
@@ -0,0 +1,53 @@
...
+    ----------
+    .. [1] Hennig, P., Osborne, Mike A. and Girolami M., Probabilistic numerics and
+       uncertainty in computations. *Proceedings of the Royal Society of London A:
+       Mathematical, Physical and Engineering Sciences*, 471(2179), 2015.
+    .. [2] Cockayne, J., Oates, C., Sullivan Tim J. and Girolami M., Bayesian
+       probabilistic numerical methods. *SIAM Review*, 61(4):756–789, 2019
+
+    Notes
+    -----
+    All PN methods should subclass this base class. Typically convenience functions
```
pnkraemer, line 33: just wondering: so far, no class implements this interface, right?

src/probnum/_probabilistic_numerical_method.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def __init__(self, prior: BeliefType):
+        self.prior = prior
+
+    @abstractmethod
+    def solve(
+        self,
+        problem: ProblemType,
+    ) -> Tuple[BeliefType, StateType]:
```
pnkraemer, line 45: IMO, the chances are high that the output types here will be points of discussion, but let's move that to a point when at least 2 algorithms implement this interface (to analyse common denominators, etc.)

src/probnum/_probabilistic_numerical_method.py
```diff
@@ -0,0 +1,53 @@
...
+    derived subclass.
+    """
+
+    def __init__(self, prior: BeliefType):
+        self.prior = prior
+
+    @abstractmethod
+    def solve(
+        self,
+        problem: ProblemType,
```
pnkraemer, line 44: if I remember correctly, it was on purpose that the `solve` has only a single argument: `problem`, right? I like this idea, but expect that it will require some thinking to implement everywhere :)

pnkraemer (commented):
Thanks a lot for doing this! I have a few super minor questions, but we should be good to go without many modifications. 

Thanks also for taking care of the `randvars` namespaces (no more `probnum.RandomVariable`). I think it improves the clarity of the code.

One question for further consideration (can be a future PR, but the line of thought might matter here): do you think it would be worth establishing a general testcase for a PN method, which would streamline the tests for this general interface?

pnkraemer:
Oh, and the test coverage thing I dont mind.

JonathanWenger:
> One question for further consideration (can be a future PR, but the line of thought might matter here): do you think it would be worth establishing a general testcase for a PN method, which would streamline the tests for this general interface?

Yes i definitely think so. I am not quite sure how to best realize this yet, but the [kinds of tests](https://probnum.readthedocs.io/en/latest/development/unit_testing.html#Testing-Probabilistic-Numerical-Methods) we always write for PN methods should be written for this class. Once a new PN method is added a lot of testing is already done. Conceptually in the same way we do it for `LinearOperators`.

src/probnum/__init__.py
JonathanWenger: It's a deliberate choice (similar to `np.asarray`, `torch.as_tensor`) but I can be convinced to move it to randvars. Future functions which can move into the `probnum` namespace are `probnum.sum`, etc. which can operate on random variables / processes.

docs/source/public_api/linops.rst
JonathanWenger: The text is removed so there is only a single source of documentation (the docstrings), otherwise there is a doubling between the `.rst` files and the docstrings in the module level `__init__.py` files. Thanks for noting it though, I had not moved all of these paragraphs to the corresponding docstrings! Should be fixed now.

docs/source/development/adding_to_the_api_documentation.ipynb
JonathanWenger: Nope, just corrected `probnum.RandomVariable` to `randvars.RandomVariable`. Readable diffs are a consequence of not having rerun any cells that actually produce output.

src/probnum/_probabilistic_numerical_method.py
JonathanWenger: Agreed.

src/probnum/_probabilistic_numerical_method.py
JonathanWenger: Yes, the result of our discussion was that any other parameter / method component choices are part of the instantiation of a new method. However, I agree that this might have to change in the future.

src/probnum/__init__.py
pnkraemer: I would move it to randvars, and find it a bit out of place here personally, but do not mind if you want to leave it in. You're opinion on this matter may be stronger than mine.

PR #420, opened 2021-06-08
BQ loop structure

alpiges:
This PR adds structure to run BQ in a loop.
`bayesian_quadrature.integrate` now calls an iterator that, given a batch size and a policy, updates the integral belief. 

The only present policy is still random sampling. A new feature is that there is a `bayesquad_fixed` interface that accepts fixed nodes and function evaluations to estimate the integral.


src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns
+        -------
+        has_converged:
+            Boolean whether a stopping criterion has been reached
+        """
+        return self._stopping_criterion(bq_state)
+
+
+class IntegralVariance(StoppingCriterion):
```
mmahsereci, line 42: Let;s make it more clear that this is a stopping criterion and not the integral variance, e.g., `IntegralVarianceThreshold`. Otherwise the namespace starts to be confusing.

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __init__(self, var_tol: FloatArgType = None):
+        self.var_tol = var_tol
+        super().__init__(stopping_criterion=self.__call__)
+
+    def __call__(self, integral_belief: Normal, bq_state: BQState) -> bool:
+        return integral_belief.var <= self.var_tol
+
+
+class RelativeError(StoppingCriterion):
```
mmahsereci, line 59: We also need a better name for this one (Although I can;t think of a good one right now.)

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
```diff
@@ -0,0 +1,114 @@
...
+        nodes = np.concatenate((bq_state.nodes, new_nodes), axis=0)
+        fun_evals = np.append(bq_state.fun_evals, new_fun_evals)
+
+        # kernel quantities
+        if old_nodes.size == 0:
+            gram = bq_state.kernel(new_nodes)
+            kernel_means = bq_state.kernel_embedding.kernel_mean(new_nodes)
+        else:
+            gram_new_new = np.atleast_2d(bq_state.kernel(new_nodes))
+            gram_old_new = np.atleast_2d(bq_state.kernel(new_nodes, old_nodes))
```
mmahsereci, line 61: Add dimensions explicitly if needed, and not use `np.atleast_...` functionality.

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
```diff
@@ -0,0 +1,114 @@
...
+                    bq_state.kernel_means,
+                    bq_state.kernel_embedding.kernel_mean(new_nodes),
+                )
+            )
+
+        initial_error = bq_state.kernel_embedding.kernel_variance()
+        weights = self._solve_gram(gram, kernel_means)
+
+        # integral mean and variance
+        integral_mean = np.squeeze(weights @ fun_evals)
```
mmahsereci, line 79: Remove dimensions explicitly if needed instead of `np.squeeze`.

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
```diff
@@ -0,0 +1,114 @@
...
+                )
+            )
+            kernel_means = np.concatenate(
+                (
+                    bq_state.kernel_means,
+                    bq_state.kernel_embedding.kernel_mean(new_nodes),
+                )
+            )
+
+        initial_error = bq_state.kernel_embedding.kernel_variance()
```
mmahsereci, line 75: `initial_integral_variance` to be consistent with naming? I personally don;t like error so much as it implies units of f and not f^2.

src/probnum/quad/bq_methods/policies/_policy.py
```diff
@@ -0,0 +1,108 @@
...
+        Objective to sample from. Need to have a method ``sample``.
+    batch_size :
+        Size of batch that is found in every iteration of calling the policy.
+    """
+
+    def __init__(self, sampling_objective, batch_size: int) -> None:
+        self.sampling_objective = sampling_objective
+        self.batch_size = batch_size
+
+        super().__init__(batch_size=batch_size, is_deterministic=False)
```
mmahsereci, line 93: same here, call `super` first, and setting instance variable `batch_size` is already done there.

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+    """Base class for a stopping criterion.
+
+    Parameters
+    ----------
+    stopping_criterion :
+        A function that determines whether a convergence criterion is reached.
+    """
+
+    def __init__(self, stopping_criterion: Callable):
+        self._stopping_criterion = stopping_criterion
```
mmahsereci, line 22: `stopping_criterion` argument not needed I suppose. Rather raise NotImplementedError in `__call__`

mmahsereci (commented):
Thanks, Alex. I put some initial comments for now.

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if info is None:
+            info = BQInfo(nevals=self.fun_evals.size)
+        self.info = info
+        self.gram = gram
+        self.kernel_means = kernel_means
+
+    @classmethod
+    def from_new_data(
+        cls, nodes, fun_evals, integral_belief, prev_state, gram, kernel_means
```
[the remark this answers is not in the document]
alpiges: Both work I guess. We stuck with the structure used for the linear solvers, see [here](https://github.com/JonathanWenger/probnum/blob/linsolve-refactor/src/probnum/linalg/solvers/_state.py)

src/probnum/quad/bq_methods/policies/_policy.py
```diff
@@ -0,0 +1,108 @@
...
+    batch_size :
+        Size of batch that is found in every iteration of calling the policy.
+    is_deterministic :
+        Is the policy a deterministic function of its arguments or stochastic (i.e.
+        sampling-based)?
+    """
+
+    def __init__(self, batch_size: int, is_deterministic: bool) -> None:
+        self.batch_size = batch_size
+        self._is_deterministic = is_deterministic
```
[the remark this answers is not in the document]
alpiges: I don't get what should go, and why. Please clarify.

src/probnum/quad/bq_methods/policies/_policy.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class Policy(abc.ABC):
+    """An abstract class for acquisition functions for Bayesian quadrature.
+
+    Parameters
+    ----------
+    batch_size :
+        Size of batch that is found in every iteration of calling the policy.
+    is_deterministic :
```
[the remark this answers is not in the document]
alpiges: True that. The optimizer should probably compute the maximum with a few random restarts. I think `is_stochastic` is just `~is_deterministic`, but maybe this flag is not actually meaningful. I just want to distinguish between random sampling and optimization schemes.

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
```diff
@@ -0,0 +1,114 @@
...
+from typing import Optional
+
+import numpy as np
+from scipy.linalg import cho_factor, cho_solve
+
+from probnum.quad.bq_methods.bq_state import BQState
+from probnum.randvars import Normal
+
+
+class BQBeliefUpdate(abc.ABC):
```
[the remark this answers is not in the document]
alpiges: I don't mind. @JonathanWenger? This should then be consistent throughout `probnum`.

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+    fun_evals:
+        Function evaluations at nodes.
+    """
+
+    def __init__(
+        self,
+        measure: IntegrationMeasure,
+        kernel: Kernel,
+        integral_belief: Optional[Normal] = None,
+        previous_integral_beliefs: Optional[Tuple[Normal]] = (),
```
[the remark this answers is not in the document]
alpiges: Because you might be interested in convergence behavior as a function of samples. Also there's a stopping criterion `RelativeError` that stops when the difference of consecutive integral estimates drop below a certain value.

src/probnum/quad/_bayesquad.py
```diff
@@ -13,24 +13,30 @@
...
-from probnum.type import FloatArgType
+from probnum.type import FloatArgType, IntArgType
 
 from ._integration_measures import IntegrationMeasure, LebesgueMeasure
 from .bq_methods import BayesianQuadrature
 
 
 # pylint: disable=too-many-arguments
+# TODO: Do we really need input_dim as a separate argument? Both kernel and measure have
+# TODO: either input_dim or dim that we could straightforwardly extract.
```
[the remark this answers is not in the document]
alpiges: I left `input_dim`. I don't think it's a huge overhead.
`scipy` has separate routines for univariate and bivariate integration, just saying ;)

src/probnum/quad/_bayesquad.py
```diff
@@ -13,24 +13,30 @@
...
 def bayesquad(
     fun: Callable,
     input_dim: int,
     kernel: Optional[Kernel] = None,
     domain: Optional[
         Tuple[Union[np.ndarray, FloatArgType], Union[np.ndarray, FloatArgType]]
```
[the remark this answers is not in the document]
alpiges: In 1d you might want to pass e.g. `(0,1)`, and not fiddle with making natural floats be arrays.

src/probnum/quad/bq_methods/policies/_policy.py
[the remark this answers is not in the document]
mmahsereci: Since `Policy` is an abstract class anyways, you can do 

```
@property    
@abc.abstractmethod
def is_deterministic(self) -> None:
    pass
```

and not use the instance variable `self._is_deterministic`. Then, the child class simply implements it, since that value is fixes anyways.

src/probnum/quad/_bayesquad.py
```diff
@@ -55,45 +61,49 @@ def bayesquad(
...
     integral :
-        The integral of ``func`` on the domain.
+        The integral of ``fun`` on the domain.
     info :
         Information on the performance of the method.
 
+    Raises
+    ------
+    ValueError
+        If neither a domain nor a measure are given.
```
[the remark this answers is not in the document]
alpiges: If both are given we could print a warning saying that only the measure will be used. For example, the Lebesgue measure already has a domain, and additionally passing domain is either redundant or wrong.

src/probnum/quad/bq_methods/policies/_policy.py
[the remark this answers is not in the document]
mmahsereci: We can keep it if you think it;s useful, but I am also OK with removing it, if it is currently not used. My point was that `is_stochastic` my imply random numbers explicitly, so the negation is not entirely the same as `is_deterministic` and would not cause the confusion that, even though on paper some policies are deterministic, in practice they are not due to the implicit randomness (seed of the optimizer and numerical errors due to early stopping). The more I think about it though, the more confusing it gets though.

src/probnum/quad/bq_methods/bq_state.py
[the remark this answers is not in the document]
mmahsereci: OK, thanks, this wasn;t obvious to me.

src/probnum/quad/_bayesquad.py
```diff
@@ -113,19 +123,100 @@ def bayesquad(
...
+        input_dim=input_dim,
+        kernel=kernel,
+        measure=measure,
+        max_nevals=n_evals,
+        batch_size=n_evals,
+    )
 
     # Integrate
-    integral, info = bq_method.integrate(fun=fun, measure=measure, nevals=nevals)
+    integral_belief, bq_state = bq_method.integrate(nodes=nodes, fun_evals=f_evals)
```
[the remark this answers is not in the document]
alpiges: Let's do `fun_evals`

src/probnum/quad/_bayesquad.py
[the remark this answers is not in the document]
mmahsereci: Shall we allow `(0, 1)` for 1D and multi D as well (all lower bounds the same, all upper bounds the same)? That is an API that is flexible but easy to understand? And its easier to handle internally. So either `Tuple[np.ndarray, np.ndarray]` or `Tuple[float, float]`, and the latter is expanded to the appropriate dimension.

src/probnum/quad/policies/_policy.py
```diff
@@ -0,0 +1,108 @@
...
+        Size of batch that is found in every iteration of calling the policy.
+    bq_state :
+        Current state of the BQ method.
+    """
+
+    def __init__(self, acquisition: Acquisition, batch_size: int) -> None:
+        self.acquisition = acquisition
+        self.batch_size = batch_size
+
+        super().__init__(batch_size=batch_size, is_deterministic=True)
```
mmahsereci, line 62: remove the instance variable `self.batch_size` as it is set in the super class already. Further, always call `super` first if it is possible, i.e., move the super call above the definitions of the instance variables.

src/probnum/quad/policies/_policy.py
```diff
@@ -0,0 +1,108 @@
...
+        Objective to sample from. Needs to have a method ``sample``.
+    batch_size :
+        Size of batch that is found in every iteration of calling the policy.
+    """
+
+    def __init__(self, sampling_objective, batch_size: int) -> None:
+        self.sampling_objective = sampling_objective
+        self.batch_size = batch_size
+
+        super().__init__(batch_size=batch_size, is_deterministic=False)
```
mmahsereci, line 93: same here

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -22,104 +25,234 @@ class BayesianQuadrature:
...
-        if method != "vanilla":
-            raise NotImplementedError
+
+        # Select policy and belief update
         if kernel is None:
             kernel = ExpQuad(input_dim=input_dim)
         if policy == "bmc":
-            policy = sample_from_measure
+            policy = RandomPolicy(measure, batch_size=batch_size)
+            belief_update = BQStandardBeliefUpdate()
```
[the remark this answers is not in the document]
alpiges: keeping it for now for consistent naming.

src/probnum/quad/_bayesquad.py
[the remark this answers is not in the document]
mmahsereci: Let;s raise a ValueError if that happens sth along the lines of e.g., *"GaussianMeasure cannot be used with finite bounds. Either set domain to None, or use another measure." * That is safer.

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -22,104 +25,234 @@ class BayesianQuadrature:
...
-        r"""Integrate the function ``fun``.
+        # Set stopping criteria
+        _stopping_criteria = []
+        if max_nevals is not None:
+            _stopping_criteria.append(MaxNevals(max_nevals))
+        if var_tol is not None:
+            _stopping_criteria.append(IntegralVariance(var_tol))
+        if rel_tol is not None:
+            _stopping_criteria.append(RelativeError(rel_tol))
+
```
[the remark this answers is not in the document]
alpiges: If multiple stopping criteria are given, BQ stops once the first criterion is fulfilled. Added the comment.

src/probnum/quad/bq_methods/bayesian_quadrature.py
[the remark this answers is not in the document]
mmahsereci: ok

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
alpiges: `RelativeMeanChange`?

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
alpiges: Yes this is a weird construction that I copied from [here](https://github.com/JonathanWenger/probnum/blob/linsolve-refactor/src/probnum/linalg/solvers/stop_criteria/_stopping_criterion.py).
Why would the `__call__`  method be an argument of the class, @JonathanWenger ?

src/probnum/quad/bq_methods/bayesian_quadrature.py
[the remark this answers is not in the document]
mmahsereci: Thanks.

src/probnum/quad/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Parameters
+    ----------
+    rel_tol:
+        Relative error tolerance on consecutive integral mean values.
+    """
+
+    def __init__(self, rel_tol: FloatArgType = None):
+        self.rel_tol = rel_tol
+        super().__init__(stopping_criterion=self.__call__)
```
mmahsereci, line 75: same here.

alpiges: See https://github.com/probabilistic-numerics/probnum/pull/420#discussion_r651924839

src/probnum/quad/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Parameters
+    ----------
+    var_tol:
+        Tolerance value of the variance.
+    """
+
+    def __init__(self, var_tol: FloatArgType = None):
+        self.var_tol = var_tol
+        super().__init__(stopping_criterion=self.__call__)
```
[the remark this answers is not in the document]
alpiges: see https://github.com/probabilistic-numerics/probnum/pull/420#discussion_r651924839

alpiges:
We now get odd failures on windows. It seems like it does not like the name `aux`. I'll rename...

src/probnum/quad/bq_methods/policies/_policy.py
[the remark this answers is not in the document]
alpiges: I'll remove `is_deterministic` entirely. So far it's just clutter and the name of the policy anyways tells what one deals with.

src/probnum/quad/_bayesquad.py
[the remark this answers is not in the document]
alpiges: @tskarvone that might be for you: Currently `LebesgueMeasure` even accepts different dimensions for the bounds. Maybe we should enforce at least to have the same dimension, but allow `(0,1)` as a multi-D box?

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
alpiges: went for this now

src/probnum/quad/_bayesquad.py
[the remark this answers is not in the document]
alpiges: Now the error handling is back to `bayesquad` (setting defaults happens in the `BayesianQuadrature` constructor). An error is raised for GaussianMeasure+domain, a warning for LebesgueMeasure+domain.

src/probnum/quad/_bayesquad.py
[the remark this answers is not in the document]
tskarvone: I agree with your suggestion. Convenience benefits gained from allowing domain inputs of unequal dimensions are marginal while making the code unnecessarily difficult to understand.

src/probnum/quad/_bayesquad.py
```diff
@@ -54,46 +59,50 @@ def bayesquad(
...
-         Bayesian Monte Carlo    ``bmc``
-         Uncertainty Sampling    ``us``
-         Mutual Information      ``mi``
-         Integral Variance       ``iv``
-        =======================  =======
+        ==========================  =======
+         Bayesian Monte Carlo [2]_  ``bmc``
+        ==========================  =======
+
+    max_nevals :
```
JonathanWenger, line 79:
```suggestion
    max_evals :
```

src/probnum/quad/_bayesquad.py
```diff
@@ -107,25 +116,123 @@ def bayesquad(
...
 
-    if nevals is None:
-        nevals = input_dim * 25
+    # Integrate
+    integral_belief, bq_state = bq_method.integrate(fun=fun)
+
+    return integral_belief, bq_state.info
+
+
+def bayesquad_fixed(
```
JonathanWenger, line 153: What's the reasoning behind having a second interface here instead of adding optional arguments `nodes` and `fun_evals` in `bayesquad`? Potentially with a warning if a policy other than the default is specified and nodes and function evaluations are given.

src/probnum/quad/_bayesquad.py
```diff
@@ -27,10 +28,14 @@ def bayesquad(
     domain: Optional[
         Tuple[Union[np.ndarray, FloatArgType], Union[np.ndarray, FloatArgType]]
     ] = None,
-    nevals: int = None,
     measure: Optional[IntegrationMeasure] = None,
-    method: str = "vanilla",
-    policy: str = "bmc",
+    policy: Optional[str] = "bmc",
+    max_nevals: Optional[IntArgType] = None,
```
JonathanWenger, line 33:
```suggestion
    max_evals: Optional[IntArgType] = None,
```

src/probnum/quad/_bayesquad.py
```diff
@@ -54,46 +59,50 @@ def bayesquad(
...
     fun :
-        Function to be integrated.
-    input_dim:
+        Function to be integrated. It needs to accept a shape=(n_eval, input_dim) ``np.ndarray`` and return a
+        shape=(n_eval,) ``np.ndarray``
+    input_dim :
         Input dimension of the integration problem
-    kernel:
+    kernel :
         the kernel used for the GP model
```
JonathanWenger, line 67:
```suggestion
        The kernel used for the GP model.
```

src/probnum/quad/_bayesquad.py
```diff
@@ -54,46 +59,50 @@ def bayesquad(
     Parameters
     ----------
     fun :
-        Function to be integrated.
-    input_dim:
+        Function to be integrated. It needs to accept a shape=(n_eval, input_dim) ``np.ndarray`` and return a
+        shape=(n_eval,) ``np.ndarray``
```
JonathanWenger, line 63:
```suggestion
        shape=(n_eval,) ``np.ndarray``.
```

src/probnum/quad/_bayesquad.py
```diff
@@ -54,46 +59,50 @@ def bayesquad(
...
+    input_dim :
         Input dimension of the integration problem
```
JonathanWenger, line 65:
```suggestion
        Input dimension of the integration problem.
```

src/probnum/quad/_bayesquad.py
```diff
@@ -54,46 +59,50 @@ def bayesquad(
...
-    kernel:
+    kernel :
         the kernel used for the GP model
     domain :
-        *shape=(dim,)* -- Domain of integration. Contains lower and upper bound as int or ndarray.
+        *shape=(input_dim,)* -- Domain of integration. Contains lower and upper bound as ``int`` or ``np.ndarray``.
     measure:
         Integration measure, defaults to the Lebesgue measure.
```
JonathanWenger, line 71:
```suggestion
        Integration measure. Defaults to the Lebesgue measure.
```

src/probnum/quad/_bayesquad.py
```diff
@@ -107,25 +116,123 @@ def bayesquad(
     >>> print(F.mean)
     0.5000
     """
+
+    # Error management
```
JonathanWenger, line 120:
```suggestion
    # Check input argument compatibility
```

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -22,104 +30,244 @@ class BayesianQuadrature:
...
     Parameters
     ----------
-    kernel:
+    kernel :
         The kernel used for the GP model.
-    policy:
+    measure :
+        The integration measure.
+    policy :
         The policy for acquiring nodes for function evaluations.
```
JonathanWenger, line 38:
```suggestion
        The policy choosing nodes at which to evaluate the integrand.
```

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -22,104 +30,244 @@ class BayesianQuadrature:
...
+    ) -> None:
         self.kernel = kernel
+        self.measure = measure
         self.policy = policy
+        self.belief_update = belief_update
+        self.stopping_criteria = stopping_criteria
 
     @classmethod
-    def from_interface(
+    def from_bayesquad(
```
JonathanWenger, line 60: I recommend renaming this (e.g. `from_problem`), since it is not only called from the interface, but also as a convenience for a more experienced user compiling a quadrature method. You might also want to pass the integrand here to determine differentiability or else in the future.

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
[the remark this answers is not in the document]
JonathanWenger: I don't feel strongly about this, but let's stick with `BQBeliefUpdate` for now.

src/probnum/quad/policies/_policy.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+
+    def __init__(self, acquisition: Acquisition, batch_size: int) -> None:
+        super().__init__(batch_size=batch_size)
+        self.acquisition = acquisition
+
+    def __call__(self, bq_state: BQState) -> np.ndarray:
+        """Find nodes according to the acquisition policy.
+
+        The nodes found are the argmin of the given acquistion function
```
JonathanWenger, line 56:
```suggestion
        The nodes found are the :math:`\\argmin` of the given acquisition function.
```

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
JonathanWenger: This was originally intended to be a convenient way for a user to create a (custom) policy simply from a callable. This turned out to be bad design as you rightfully point out. We will simply force subclassing instead for all components of PN methods.

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+            Current Gaussian belief about the integral.
+        bq_state:
+            State of the BQ loop.
+
+        Returns
+        -------
+        has_converged:
+            Boolean whether a stopping criterion has been reached
+        """
+        return self._stopping_criterion(bq_state)
```
JonathanWenger:
```suggestion
        raise NotImplementedError
```

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
```diff
@@ -0,0 +1,101 @@
...
+class StoppingCriterion:
+    """Base class for a stopping criterion.
+
+    Parameters
+    ----------
+    stopping_criterion :
+        A function that determines whether a convergence criterion is reached.
+    """
+
+    def __init__(self, stopping_criterion: Callable):
```
JonathanWenger: `__init__` is now obsolete in light of the comment below.

JonathanWenger (changes requested):
I am so happy that BQ is moving ahead in the library. Also great that you adopted the design of the linear solver refactor. Your PR seems like it will be done before mine :grin:. I made some suggestions and answered the question about `__call__` as an argument.

src/probnum/quad/_bayesquad.py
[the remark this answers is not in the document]
tskarvone: I have implemented this.

src/probnum/quad/_bayesquad.py
mmahsereci: Yeah, that;s a good question, and it;s good you raised it. From my current point of view, `bayesquad` and `bayesquad_fixed` are two quite different methods. One runs an acquisition loop and actively selects data. so you provide the problem (callable) in the API. The other one does not acquire data, so you request not the problem but the data in the API (basically what `bayesquad` *is*, you request there). I find convoluting those to different problems settings into one API more confusing than helpful. You;d probably have to read the docs several minutes and then still be puzzled (but perhaps I just get puzzled easily :)) Scipy also has several APIs for quadrature methods that differ for this reason. I;d be interested to hear more opinions on this though @alpiges and @tskarvone .

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
alpiges: will be fixed in the future

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
alpiges: will be fixed in the future

src/probnum/quad/bq_methods/stop_criteria/_stopping_criterion.py
alpiges: done

src/probnum/quad/_bayesquad.py
alpiges: Great you bring this up---we've had extended conversations about this. We went for this setup mostly for the reasons given by @mmahsereci. In `scipy.integrate` you need to use a spline method such as `simpson` or `romb` to use on a data array.
Integrating this method into `bayesquad` might make the interface more confusing when all parameters are optional, but either `fun` or `fun-evals` must be given, and if `fun_evals` is given you also need nodes etc.

However, we might consider a merger in the future, in case the user already has some `fun_evals`, but wants to use BQ to collect some more. But that's too niche for the instant.

(on a different note: What might be confusing is that `fixed_quad` in `scipy` denoted fixed order Gaussian quadrature. Maybe we want to rename our method not to be mistaken for a probabilistic version of `fixed_quad`. Any ideas?)

src/probnum/quad/_bayesquad.py
alpiges: If I ever learn to terminate sentences with full stops :sweat_smile:

src/probnum/quad/bq_methods/bayesian_quadrature.py
alpiges: Renamed :white_check_mark: 
Didn't we discuss at some point that the function should not be an argument of `BayesianQuadrature`, so that one instance of `BayesianQuadrature` can in principle integrate multiple functions? Maybe I don't fully understand your design choice with the classmethod there :thinking:

src/probnum/quad/_bayesquad.py
alpiges: How about `bayesquad_from_data`?

JonathanWenger: Thanks for the clarification. I understand the difference now. I like `bayesquad_from_data`. `bayesquad_fixed` sounds a little like the method has been fixed (as in it was broken before). So really this is a convenient interface for the constructor `BayesianQuadrature.from_data()`, correct?

alpiges: Currently there is no separate constructor for the `from_data` version in `BayesianQuadrature` (see [here](https://github.com/probabilistic-numerics/probnum/pull/420#discussion_r657021351)), because the constructor is problem-agnostic at this point. This is something we could discuss.
Both `bayesquad` and `bayesquad_from_data` therefore use the same constructor.

src/probnum/quad/bq_methods/bayesian_quadrature.py
JonathanWenger: Correct the `__init__` method should not require the function, but only the prior information. 

However, you want to support (at least one, possibly multiple) constructors which define standard ways to instantiate a BQ method from the information you have available. For example if you have a function which is made up from basic differentiable functions, your `classmethod` can then define default prior choices. You can also use this to construct a BQ method with an example function which has properties shared across all the functions you want to integrate with that method.

alpiges: That sounds reasonable once we impose structure on how the integrand shall be passed (alongside additional information about the function). For now this is not used and the function is not needed in the constructor. Does it make sense to address this at a later point? Otherwise there will be an unused argument `fun`.

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -22,104 +30,244 @@ class BayesianQuadrature:
...
-        gram :
-            *shape=(nevals, nevals)* -- Symmetric pos. def. kernel Gram matrix :math:`K`.
-        rhs :
-            *shape=(nevals, ...)* -- Right-hand-side :math:`b`, matrix or vector.
+        fun :
+            The integrand. Optional when fixed function evaluations are instead passed.
+        nodes :
+            Optional nodes available from the start.
+        fun_evals :
+            Optional function evaluations available from the start.
```
mmahsereci, line 253: similar to above, add shapes and what fun expects and returns.

src/probnum/quad/bq_methods/bayesian_quadrature.py
```diff
@@ -22,104 +30,244 @@ class BayesianQuadrature:
...
 
         Returns
         -------
-        x:
-            The solution to the linear system :math:`K x = b`.
+        integral_belief:
+            Posterior belief about the integral.
+        bq_state:
+            Final state of the Bayesian quadrature methods.
```
mmahsereci, line 260:
```suggestion
            Final state of the Bayesian quadrature method.
```

src/probnum/quad/bq_methods/belief_updates/_belief_update.py
```diff
@@ -0,0 +1,113 @@
...
+        """Updates integral belief and BQ state according to the new data given.
+
+        Parameters
+        ----------
+        bq_state :
+            Current state of the Bayesian quadrature loop.
+        new_nodes :
+            New nodes that have been added.
+        new_fun_evals :
+            Function evaluations at the given node.
```
mmahsereci, line 38: add shapes here, too.

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+    stopping_criterion:
+        The stopping criterion used to determine convergence.
+    """
+
+    def __init__(
+        self,
+        iteration: int = 0,
+        nevals: int = 0,
+        has_converged: bool = False,
+        stopping_criterion=None,
```
mmahsereci, line 33: type hint missing

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+    ):
+        self.iteration = iteration
+        self.nevals = nevals
+        self.has_converged = has_converged
+        self.stopping_criterion = stopping_criterion
+
+    def update_iteration(self, batch_size):
```
mmahsereci, line 40:
```suggestion
    def update_iteration(self, batch_size: int) -> None:
```

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+    fun_evals:
+        Function evaluations at nodes.
+    """
+
+    def __init__(
+        self,
+        measure: IntegrationMeasure,
+        kernel: Kernel,
+        integral_belief: Optional[Normal] = None,
+        previous_integral_beliefs: Optional[Tuple[Normal]] = (),
```
mmahsereci, line 78:
```suggestion
        previous_integral_beliefs: Tuple[Normal] = (),
```

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+        info: Optional[BQInfo] = None,
+        batch_size: Optional[int] = 1,
```
mmahsereci, line 80:
```suggestion
        batch_size: int = 1,
```

src/probnum/quad/bq_methods/bq_state.py
```diff
@@ -0,0 +1,123 @@
...
+        nodes: Optional[np.ndarray] = None,
+        fun_evals: Optional[np.ndarray] = None,
+        gram=np.array([[]]),
+        kernel_means=np.array([]),
```
mmahsereci, line 84:
```suggestion
        gram: np.ndarray=np.array([[]]),
        kernel_means: np.ndarray=np.array([]),
```

mmahsereci (commented):
Some more comments, but most of them are quite fast to resolve. Looks fairly good otherwise!

src/probnum/quad/_integration_measures.py
```diff
@@ -18,28 +18,28 @@ class IntegrationMeasure(abc.ABC):
...
+        self._set_dimension_domain(input_dim, domain)
 
     def __call__(self, points: Union[float, np.floating, np.ndarray]) -> np.ndarray:
         """Evaluate the density function of the integration measure.
 
         Parameters
         ----------
         points :
-            *shape=(n_points,) or (n_points, dim)* -- Input locations.
+            *shape=(n_points,) or (n_points, input_dim)* -- Input locations.
```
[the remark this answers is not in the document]
tskarvone: Went with `(n_points, input_dim)` and made some related changes.

src/probnum/quad/_integration_measures.py
```diff
@@ -66,79 +66,67 @@ def sample(
...
+                "of equal dimension."
             )
-        if dim < 1:
-            raise ValueError(f"Domain dimension dim = {dim} must be positive.")
-
-        # Use same domain limit in all dimensions if only one limit is given
-        if domain_a_dim == 1:
-            domain_a = np.full((dim,), domain[0])
+        if input_dim is not None and input_dim < 1:
+            raise ValueError("If given, input dimension must be positive.")
```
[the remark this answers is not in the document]
tskarvone: Done.

src/probnum/quad/_integration_measures.py
```diff
@@ -66,79 +66,67 @@ def sample(
...
+
+        of dimension ``input_dim``.
+        """
+        # Domain limits must have equal dimensions and input dimension must be positive
+        if np.size(domain[0]) != np.size(domain[1]):
             raise ValueError(
-                "Domain limits must have the same length or at least "
-                "one of them has to be one-dimensional."
+                "Domain limits must be given either as scalars or arrays "
+                "of equal dimension."
```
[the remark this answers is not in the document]
tskarvone: Done.

src/probnum/quad/_integration_measures.py
```diff
@@ -18,28 +18,28 @@ class IntegrationMeasure(abc.ABC):
...
+        *shape=(input_dim,)* -- Domain of integration. Contains lower and upper bound as
+         ``int`` or ``np.ndarray``.
     """
 
     def __init__(
         self,
-        dim: IntArgType,
-        domain: Tuple[Union[np.ndarray, FloatArgType], Union[np.ndarray, FloatArgType]],
+        input_dim: IntArgType,
+        domain: Union[Tuple[FloatArgType, FloatArgType], Tuple[np.ndarray, np.ndarray]],
```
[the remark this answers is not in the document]
tskarvone: Done.

src/probnum/quad/_integration_measures.py
```diff
@@ -194,72 +182,71 @@ def sample(
...
         if np.isscalar(mean):
-            dim = 1
+            input_dim = 1
         else:
-            dim = mean.size
+            input_dim = mean.size
 
         # If cov has been given as a vector of variances, transform to diagonal matrix
-        if isinstance(cov, np.ndarray) and np.squeeze(cov).ndim == 1 and dim > 1:
+        # TODO: Get rid of squeezes
```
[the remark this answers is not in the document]
tskarvone: Decided to remove this functionality as the user can very well do `cov=np.diag(...)` if they want a diagonal covariance matrix.

src/probnum/quad/_integration_measures.py
```diff
@@ -194,72 +182,71 @@ def sample(
...
 
         # Set dimension based on the mean vector
         if np.isscalar(mean):
-            dim = 1
+            input_dim = 1
         else:
-            dim = mean.size
+            input_dim = mean.size
 
         # If cov has been given as a vector of variances, transform to diagonal matrix
```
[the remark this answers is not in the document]
tskarvone: I have removed this functionality; see above.

src/probnum/quad/_integration_measures.py
```diff
@@ -194,72 +182,71 @@ def sample(
...
         self.cov = self.random_variable.cov
 
         # Set diagonal_covariance flag
-        if dim == 1:
+        if input_dim == 1:
             self.diagonal_covariance = True
         else:
             self.diagonal_covariance = (
                 np.count_nonzero(self.cov - np.diag(np.diagonal(self.cov))) == 0
             )
```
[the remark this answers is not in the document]
tskarvone: Do you mean by calling some kind of `is_diagonal` type function? I am not aware of any such method. Could make this a separate class method though.

src/probnum/quad/_integration_measures.py
```diff
@@ -194,72 +182,71 @@ def sample(
...
 
         super().__init__(
-            dim=dim,
-            domain=(np.full((dim,), -np.Inf), np.full((dim,), np.Inf)),
+            input_dim=input_dim,
+            domain=(np.full((input_dim,), -np.Inf), np.full((input_dim,), np.Inf)),
         )
```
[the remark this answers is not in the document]
tskarvone: Done.

src/probnum/quad/_integration_measures.py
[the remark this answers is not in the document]
mmahsereci: Ah, nevermind, my bad. I somehow thought we could get the info if the covariance is diagonal from the input shapes, but this is not the case. This is fine then.

alpiges:
Following Dagstuhl, @tskarvone  and I have made an attempt to finally make the huge BQ PR mergeable.
@JonathanWenger @mmahsereci the bulk of the initial issues have been addressed. Could you have a look at the unresolved comments and check if there remain any changes that prevent the PR from being merged? Thanks a lot

mmahsereci:
@alpiges @tskarvone I'll merge this PR soon. Having had a closer look, i think there are a couple of things we might need to change in the near future, for instance:
- separating better what `BQInfo` and `BQState` is
- inheriting from the base class for Stopping criteria
- having some kind of belief object similar to the linear solver package (`BQState` is quite similar to it already)
- perhaps using dataclasses for `BQInfo`
- fixing the interface, especially where the checks and error message happen is quite buggy atm.

and some other minor things. I removed the optimal policy for now as it was unclear what the methods should look like but this code can stay in one of our forks. Let me know if you're OK with this (thumbs up for merging).

alpiges:
Thanks a lot @mmahsereci 
> 
>     * separating better what `BQInfo` and `BQState` is
> 
I wholeheartedly agree. The way this is done now is quite hacky.

> 
>     * having some kind of belief object similar to the linear solver package (`BQState` is quite similar to it already)
> 
I believe `BQBeliefUpdate` keeps track of the current belief and the way it should be updated.

mmahsereci (approved):

PR #423, opened 2021-06-09
Converting randvars tutorial to a quickstart tutorial

mmahsereci:
Converting the existing random variables tutorial into a quickstart. 

At some later point we can re-arrange the doctree (once there are more quickstart tutorials), so this one is currently still sitting where it was.


JonathanWenger (approved):
Thanks for reworking this tutorial. Looks all good to me except for the explanation of `LinearOperator`s (see the review comment on ReviewNB).

PR #424, opened 2021-06-10
updated readme and docs index

mmahsereci:
Changes:

- some minor changes in the paragraphs
- added list of currently available solvers 
- added list of other lower level structures (shall I add the sub-package names to each?)

Later, we can add links to the quickstarts. 

I used unicode for the formulas in the Readme.md as there is no good way to render LaTeX in the github readme. I think it looks OK enough..


README.md
```diff
@@ -8,16 +8,30 @@
...
+ProbNum is a Python toolkit for solving numerical problems in linear algebra, optimization, quadrature and
+differential equations. ProbNum solvers not only estimate the solution of the numerical problem, but also its
+uncertainty (the error) which certainly arises from finite computational resources, discretization, and stochastic input.
 
----
+Currently, available solvers are:
+
+- **Linear solvers:** Solve *Ax = b* for *x*.
+- **ODE solvers:** Solve *&#7823;(t) = f(&#8201;y(t), t&#8201;)* for *y*.
+- **Quadrature solvers:** Solve *F = &#x222b; f(x)&#8201;p(x)&#8201;dx* for *F*.
```
JonathanWenger, line 19: **Quadrature routines**

README.md
```diff
@@ -8,16 +8,30 @@
 
 ---
 
-**ProbNum implements probabilistic numerical methods in Python.** Such methods solve numerical problems from linear
-algebra, optimization, quadrature and differential equations using _probabilistic inference_. This approach captures 
-uncertainty arising from _finite computational resources_ and _stochastic input_. 
+ProbNum is a Python toolkit for solving numerical problems in linear algebra, optimization, quadrature and
+differential equations. ProbNum solvers not only estimate the solution of the numerical problem, but also its
+uncertainty (the error) which certainly arises from finite computational resources, discretization, and stochastic input.
```
JonathanWenger, line 13: just phrasing the uncertainty as error is misleading i think, since it ignores cross-correlation and also has the connotation of worst case error. This begs the question of why care about PN at all. 

"which ~certainly~" unnecessary and a duplication of uncertainty

JonathanWenger, line 13: I am missing the point about numerical approximation being framed as inference and a clearer pointer to probabilistic numerics to be honest.

JonathanWenger (changes requested):

README.md
```diff
@@ -8,16 +8,30 @@
...
+- **Linear solvers:** Solve *Ax = b* for *x*.
+- **ODE solvers:** Solve *&#7823;(t) = f(&#8201;y(t), t&#8201;)* for *y*.
+- **Quadrature solvers:** Solve *F = &#x222b; f(x)&#8201;p(x)&#8201;dx* for *F*.
+
+Lower level structure includes:
+
+- Some efficient **random variable arithmetics**, especially for Gaussian distributions.
+- Structure for efficient computation with **linear operators**.
+- Structure for **random processes** (especially Gauss-Markov processes).
+- **Filters and smothers** for probabilistic state space models, mostly variants of Kalman filters.
```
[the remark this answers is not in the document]
mmahsereci: done

README.md
```diff
@@ -8,16 +8,30 @@
...
+Currently, available solvers are:
+
+- **Linear solvers:** Solve *Ax = b* for *x*.
+- **ODE solvers:** Solve *&#7823;(t) = f(&#8201;y(t), t&#8201;)* for *y*.
+- **Quadrature solvers:** Solve *F = &#x222b; f(x)&#8201;p(x)&#8201;dx* for *F*.
+
+Lower level structure includes:
+
+- Some efficient **random variable arithmetics**, especially for Gaussian distributions.
+- Structure for efficient computation with **linear operators**.
```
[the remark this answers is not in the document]
mmahsereci: yeah, I agree. replaced.

README.md
```diff
@@ -8,16 +8,30 @@
...
----
+Currently, available solvers are:
+
+- **Linear solvers:** Solve *Ax = b* for *x*.
+- **ODE solvers:** Solve *&#7823;(t) = f(&#8201;y(t), t&#8201;)* for *y*.
+- **Quadrature solvers:** Solve *F = &#x222b; f(x)&#8201;p(x)&#8201;dx* for *F*.
+
+Lower level structure includes:
+
+- Some efficient **random variable arithmetics**, especially for Gaussian distributions.
```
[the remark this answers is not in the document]
mmahsereci: see below.

README.md
```diff
@@ -8,16 +8,30 @@
...
+- Structure for efficient computation with **linear operators**.
+- Structure for **random processes** (especially Gauss-Markov processes).
+- **Filters and smothers** for probabilistic state space models, mostly variants of Kalman filters.  
+- Structure for **probabilistic state space models**.
 
-[Probabilistic Numerics](http://probabilistic-numerics.org/) (PN) aims to quantify uncertainty arising from 
-intractable or incomplete numerical computation and from stochastic input using the tools of probability theory. The 
-vision of probabilistic numerics is to provide well-calibrated probability measures over the output of a numerical 
-routine, which then can be propagated along the chain of computation.
+The ProbNum library is related to [probabilistic numerics](http://probabilistic-numerics.org/) (PN)
```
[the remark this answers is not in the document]
mmahsereci: We should not write this. If you like this better we can write "The ProbNum library is inspired by probabilistic numerics, .."

README.md
```diff
@@ -8,16 +8,30 @@
...
+- Structure for **random processes** (especially Gauss-Markov processes).
+- **Filters and smothers** for probabilistic state space models, mostly variants of Kalman filters.  
+- Structure for **probabilistic state space models**.
 
-[Probabilistic Numerics](http://probabilistic-numerics.org/) (PN) aims to quantify uncertainty arising from 
-intractable or incomplete numerical computation and from stochastic input using the tools of probability theory. The 
-vision of probabilistic numerics is to provide well-calibrated probability measures over the output of a numerical 
-routine, which then can be propagated along the chain of computation.
+The ProbNum library is related to [probabilistic numerics](http://probabilistic-numerics.org/) (PN) 
+which is a research field existing at the intersection of machine learning, and numerics.
```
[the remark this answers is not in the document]
mmahsereci: see above

docs/source/index.rst
```diff
@@ -6,16 +6,37 @@ ProbNum
...
+uncertainty (the error) which certainly arises from finite computational resources, discretization, and stochastic input.
 
-----
+Currently, available solvers are:
+
+- **Linear solvers:** Solve :math:`Ax=b` for :math:`x`.
+
+- **ODE solvers:** Solve :math:`\dot{y}(t)= f(y(t), t)` for :math:`y`.
+
+- **Quadrature Solvers:** solve :math:`F = \int_{\Omega} f(x) p(x) dx` for :math:`F`.
```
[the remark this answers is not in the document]
mmahsereci: oups, corrected.

docs/source/index.rst
```diff
@@ -6,16 +6,37 @@ ProbNum
...
+
+Lower level structure in ProbNum includes:
```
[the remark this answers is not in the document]
mmahsereci: I'll change both files for all the comments that apply to both.

README.md
mmahsereci: I was thinking about this, too, as it sounded unfamiliar. But I couldn't find a reason not to use solvers analogous to the other solvers. After all, quadrature methods solve the integration problem for the internal value *F* from incomplete data.

README.md
mmahsereci: Hm, I was thinking about this, too when writing. The error is not identical with a standard deviation (standard error). It;s meant as "statistical error" or as some PN papers call it, the "discretization error". Both include the correlations. However, I wasn't happy with either since statistical error perhaps implies aleatoric uncertainty, and discretization error is only true for continuous domains (linear solvers are the exception there). Hence, I just said "error".

I find only writing "uncertainty" sub-optimal as it is a lesser understood (and less well defined term?) as e.g., statistical error, so I want to have a proper technical term in there. We could write "(the error of the solution)". This should be fairly agnostic. 

I removed the 'certainly'.

README.md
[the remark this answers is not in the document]
mmahsereci: see other comment

README.md
mmahsereci: I appended the following sentence as it illustrates a use case of the ProbNum library: "These uncertainties can be used in downstream decisions."

The pointer to the research field "probabilistic numerics" and what it implies is just below this paragraph.

README.md
JonathanWenger: I certainly agree with the last sentence. What sounded strange to me is that quadrature routines do not solve quadrature as ODE solvers solve ODEs. Rather quadrature already is the solution process, i.e. "solving an equation by quadrature". With the same logic integral solvers is much more sensible, but not really well known language.

JonathanWenger (approved):
Looks good! Thanks for the effort to go through this line-by-line. The only change I still have a comment on is "quadrature solvers".

README.md
mmahsereci: that makes total sense. what would you prefer to go for? "quadrature routines" or "integral solvers"? I wouldn't mind either.

JonathanWenger: Let's go with integral solvers. I think integration is actually probably more understandable to the layperson than quadrature.

mmahsereci: I went for "Integral solvers (quadrature):" now.

mmahsereci: oups, sorry did not see your comment as page wasn't updated. Will leave it at "Integral solvers (quadrature):" then.

Issue #421, opened 2021-06-09
Is instance variable `self._mean` needed in `Normal`?


mmahsereci:
or can it be replaced by `self.mean` in all methods?

`self._mean` is defined [here](https://github.com/probabilistic-numerics/probnum/blob/master/src/probnum/randvars/_normal.py#L120), but I am wondering if it is needed at all, as there is also the cached property `self.mean` from the parent class. Same question for `self._cov` (L121)

mmahsereci:
@JonathanWenger mentioned @marvinpfoertner may have the answer to this.

marvinpfoertner:
Yes we can get rid of it. To be honest, I don't really know why this is even there. Might be a legacy artifact from the old design of `NormalDistribution` that I missed.

PR #425, opened 2021-06-11
Removing obsolete instance variables in `Normal`

mmahsereci:
Issue: #421 

- removed `self._cov` and `self_mean` in `Normal` class as they seem obsolete.
- moved instance variable definition for cholesky functionality of `Normal` to after the `super` call.


marvinpfoertner (approved):

mmahsereci:
@JonathanWenger or @pnkraemer , can one of you merge this PR? I don;t have the rights for it it seems. @marvinpfoertner mentioned offline that the previously missing tests would be dealt with separately.

PR #457, opened 2021-07-06
Improved implementation of `random_sparse_spd_matrix`

JonathanWenger:
# In a Nutshell
With the recent changes to `random_sparse_spd_matrix` in #455 , the function now returns a sparse matrix. However, the input arguments restricting the range of entries in the Cholesky factor was no longer considered and also the output sparsity format was not tested.

### DIFF
--- a/src/probnum/problems/zoo/linalg/_random_spd_matrix.py
+++ b/src/probnum/problems/zoo/linalg/_random_spd_matrix.py
@@ -134,9 +134,9 @@ def random_sparse_spd_matrix(
     >>> sparsemat = random_sparse_spd_matrix(dim=5, density=0.1, random_state=42)
     >>> sparsemat.todense()
     matrix([[1.        , 0.        , 0.        , 0.        , 0.        ],
-            [0.        , 1.        , 0.        , 0.30424224, 0.        ],
+            [0.        , 1.        , 0.        , 0.37381802, 0.        ],
             [0.        , 0.        , 1.        , 0.        , 0.        ],
-            [0.        , 0.30424224, 0.        , 1.09256334, 0.        ],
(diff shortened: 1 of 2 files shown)


src/probnum/problems/zoo/linalg/_random_spd_matrix.py
```diff
@@ -97,16 +97,14 @@ def random_sparse_spd_matrix(
...
 ) -> np.ndarray:
-    """Random sparse symmetric positive definite matrix.
+    r"""Random sparse symmetric positive definite matrix.
 
     Constructs a random sparse symmetric positive definite matrix for a given degree
     of sparsity. The matrix is constructed from its Cholesky factor :math:`L`. Its
-    diagonal is set to one and all other entries of the lower triangle are sampled
-    from a uniform distribution with bounds :code:`[chol_entry_min, chol_entry_max]`.
-    The resulting sparse matrix is then given by :math:`A=LL^\\top`.
+    diagonal is set to one and all other nonzero entries of the lower triangle are sampled from a uniform distribution with bounds :code:`[chol_entry_min, chol_entry_max]`. The resulting sparse matrix is then given by :math:`A=LL^\top`.
```
marvinpfoertner, line 107: Why is this on a single line? And why doesn't `black` complain about line-length?

marvinpfoertner (approved):

src/probnum/problems/zoo/linalg/_random_spd_matrix.py
JonathanWenger: Not sure why black doesnt format this, possibly because of the Sphinx directives. I fixed the docstring.

PR #283, opened 2020-12-12
Refactoring of Probabilistic Linear Solvers

JonathanWenger:

Issue #51, opened 2020-04-21
Refactor the design of the probabilistic linear solver functions and classes


JonathanWenger:
- [ ] Cleanly separate `_preprocess_linear_system` from `_init_solver` and make sure only one takes the assumptions on A, i.e. `assume_A` as an argument.

- [ ] Make the noisy version of the symmetric matrix-based solver a subclass of the non-noisy version and reuse existing functions, wherever possible, i.e. `_mean_update`. Possibly even factor out some functions to simplify the `solve` method and reduce code duplication.

nathanaelbosch:
In the initial pylint PR #150 I found many pylint messages that mostly relate to `matrixbased`. I'll post a full list in here since it seems to relate to this PR. If you think that those are better tackled through individual tasks I could also open individual issues:
- `too-many-arguments`
- `abstract-method`
- `too-many-branches`
- `arguments-differ`
- `redefined-builtin`
- `too-many-lines`
- `abstract-class-instantiated`
- `too-complex`
- `too-many-instance-attributes`
- `too-many-statements`
- `attribute-defined-outside-init`

These are not *all* the errors that pylint raises, neither are those errors that *only* appear in `linalg`, but I think that most of these should be considered when working on this refactoring, and ideally I would hope for `pylint src/probnum/linalg/` to pass with these messages enabled (e.g. through the `--enable=<message>` argument).

nathanaelbosch:
Relating to my previous comment: I just opened #158 to track the work on the pylint exceptions for the `linalg` module.

PR #460, opened 2021-07-07
[PLS Refactor] `ProbabilisticLinearSolverState`, `LinearSystemBelief` and `LinearSolverPolicy`

JonathanWenger:
# In a Nutshell
First components for the new compositional structure of the `ProbabilisticLinearSolver`. Will be followed up with more PRs adding the remaining components.

# Detailed Description
This PR begins the refactoring of the probabilistic linear solver into a compositional pattern and adds the following component classes:
- `LinearSystemBelief`
- `ProbabilisticLinearSolverState`
- `LinearSolverPolicy`, incl. sub-classes implementing a random and deterministic policy

It also restructures the `probnum.linalg` API reference into a more understandable hierarchy in line with the module and semantic structure.

Note that this PR leaves the original implementation intact until all components of the new solver are included.

# Related Issues
First part of #51 and supersedes part of #283.


src/probnum/linalg/solvers/beliefs/_linear_system_belief.py
```diff
@@ -0,0 +1,126 @@
...
+                raise dim_mismatch_error(A, Ainv, "A", "Ainv")
+
+        self._x = x
+        self._A = A
+        self._Ainv = Ainv
+        self._b = b
+        self._hyperparams = hyperparams
+
+    @property
+    def hyperparams(
```
marvinpfoertner, line 90: Mutability might be an issue here.
Maybe introduce getters `hyperparameter(key)`?

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
```diff
@@ -0,0 +1,31 @@
+"""Policy returning :math:`A`-conjugate actions."""
+
+import numpy as np
+
+import probnum  # pylint: disable="unused-import"
```
marvinpfoertner, line 5: Why do we need this here?

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
```diff
@@ -0,0 +1,31 @@
...
+
+from . import _linear_solver_policy
+
+
+class ConjugateGradientPolicy(_linear_solver_policy.LinearSolverPolicy):
+    r"""Policy returning :math:`A`-conjugate actions.
+
+    Selects the negative gradient / residual as an initial action :math:`s_0 = A x_0 - b` and then successively generates :math:`A`-conjugate actions, i.e. the actions satisfy :math:`s_i^\top A s_j = 0` iff :math:`i \neq j`.
```
marvinpfoertner, line 13: `s_0 = b - A x_0`?

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
```diff
@@ -0,0 +1,31 @@
...
+    def __call__(
+        self, solver_state: "probnum.linalg.solvers.LinearSolverState"
+    ) -> np.ndarray:
+
+        action = -solver_state.residual.copy()
+
+        if solver_state.step > 0:
+            # A-conjugacy correction (in exact arithmetic)
+            beta = (
+                np.linalg.norm(solver_state.residual)
```
marvinpfoertner, line 25: This can be cached and reused in the stopping criteria

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
```diff
@@ -0,0 +1,31 @@
...
+                / np.linalg.norm(solver_state.residuals[solver_state.step - 1])
```
marvinpfoertner, line 26: See above

src/probnum/problems/_problems.py
```diff
@@ -194,6 +194,26 @@ class LinearSystem:
     # For testing and benchmarking
     solution: Optional[Union[np.ndarray, randvars.RandomVariable]] = None
 
+    @classmethod
+    def from_matrix(
```
marvinpfoertner, line 198: I feel like this should go into the problem zoo, since there is the specific choice of the normally distributed rhs

src/probnum/randvars/_scipy_stats.py
```diff
@@ -11,7 +11,7 @@
 
 _ValueType = Union[np.generic, np.ndarray]
 
-# pylint: disable=protected-access
+# pylint: disable="protected-access,too-many-branches"
```
marvinpfoertner, line 14: Why the `"`?

pyproject.toml
```diff
@@ -65,13 +65,7 @@ protected-access,
 no-self-use,
 arguments-differ,
 attribute-defined-outside-init,
-too-many-statements,
 too-many-instance-attributes,
-too-complex,
-too-many-lines,
-redefined-builtin,
-abstract-method,
-too-many-branches,
```
marvinpfoertner, line 74: Instead of removing these here and adding them in `tox.ini` for other packages and in files of packages other than `linalg`, should we maybe make a dedicated PR discussing which of these we want to keep in here in the long run?
This way we disentangle PRs and can discuss things with all maintainers

tests/test_linalg/cases/linear_systems.py
```diff
@@ -0,0 +1,37 @@
...
+
+@pytest_cases.parametrize_with_cases(
+    "spd_matrix", cases=cases_matrices, has_tag=["symmetric", "positive_definite"]
+)
+def case_spd_linsys(
+    spd_matrix: Union[np.ndarray, scipy.sparse.spmatrix, linops.LinearOperator],
+    rng: np.random.Generator,
+) -> problems.LinearSystem:
+    """Linear system with symmetric positive definite matrix."""
+    if hasattr(spd_matrix, "valuegetter"):
```
marvinpfoertner, line 34: I guess this can be resolved with the new release of `pytest_cases`?

tests/test_linalg/conftest.py
```diff
@@ -0,0 +1,15 @@
...
+
+@pytest_cases.fixture()
+def rng() -> np.random.Generator:
+    """Random number generator."""
+    return np.random.default_rng(42)
+
+
+@pytest_cases.fixture()
+def my_fixture():
+    pass
```
marvinpfoertner, line 15: Is this an artifact from the `pytest_cases` issue?

tests/test_linalg/test_state.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+@parametrize_with_cases("state", cases=cases_states)
+def test_next_step(state: LinearSolverState):
+    """Test whether advancing a state to the next step updates all state attributes
+    correctly."""
+    initial_step = state.step
+    state.next_step()
+
+    assert initial_step + 1 == state.step
```
marvinpfoertner, line 26: I think, according to the docstring, there are some assertions missing here

marvinpfoertner (changes requested):
Very nice! Looking forward to working in this abstraction!

This review is very picky. Feel free to mark as out-of-scope if things are irrelevant.

General points:
- Personally I would factor out all pylint related changes which affect packages other than `linalg` into a separate PR and discuss with all maintainers. Let's discuss this offline.

JonathanWenger:
- [ ] Remove `LinearSolverState.cache` for now since it is not used.

pyproject.toml
JonathanWenger: - [ ] remove changes to `tox.ini`

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
JonathanWenger: We need `probnum.linalg.solvers.LinearSolverState` as a type hint for the policies. Due to circular dependencies we need to use a forward references. However, Sphinx can only resolve this if the corresponding module is actually imported.

src/probnum/problems/_problems.py
JonathanWenger: I hesitated creating a problem zoo method which is really just a convenience constructor for a `LinearSystem`. Now arguably `random_spd_matrix` could then also be called in a constructor for `LinearSystem`. To alleviate the choice of normal distribution I gave `from_matrix` an optional `solution_rv` argument where one can specify a random variable to draw the solution from.

Let me know if you prefer a function under problem zoo which has the equivalent functionality and is called `random_linear_system`, which takes a matrix / callable sampling a system matrix and a randvar / solution to generate the right hand side from.

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
JonathanWenger: As discussed the cache in `LinearSolverState` will be handled in a future PR.

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
JonathanWenger: As discussed the cache in `LinearSolverState` will be handled in a future PR.

src/probnum/linalg/solvers/beliefs/_linear_system_belief.py
```diff
@@ -0,0 +1,126 @@
...
+try:
+    # functools.cached_property is only available in Python >=3.8
+    from functools import cached_property
+except ImportError:
+    from cached_property import cached_property
+
+# pylint: disable="invalid-name"
+
+
+class LinearSystemBelief:
```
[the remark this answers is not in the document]
JonathanWenger: as discussed, we'll stick with one for now. The main reason being, that if we want to unify matrix / solution-based inference both x and Ainv should be in the base belief (even if only one of them is instantiated). But then in the exact case, beliefs over A and b can always be output by the algorithm's `solve` function, when it gets the `problem`. To then create a different belief object seems unnecessarily complicated. Instead of having `A` and `b` in the init one could think of having a `qoi` dictionary, which then can also hold other quantities of interest (like the belief over the residual, spectrum or else).

src/probnum/linalg/solvers/_state.py
```diff
@@ -0,0 +1,126 @@
...
+    ):
+        self.problem = problem
+        self.belief = prior
+
+        self.step = 0
+
+        self._actions: List[np.ndarray] = [None]
+        self._observations: List[Any] = [None]
+        self._residuals: List[np.ndarray] = [
+            self.problem.A @ self.belief.x.mean - self.problem.b,
```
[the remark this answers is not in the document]
JonathanWenger: Not yet. Once we'll introduce `Observation` objects, these will be replaced accordingly.

src/probnum/linalg/_problinsolve.py
```diff
@@ -13,12 +13,12 @@
...
-from probnum.linalg.linearsolvers.matrixbased import (
+from probnum.linalg.solvers.matrixbased import (
     AsymmetricMatrixBasedSolver,
     NoisySymmetricMatrixBasedSolver,
     SymmetricMatrixBasedSolver,
 )
-from probnum.linalg.linearsolvers.solutionbased import SolutionBasedSolver
+from probnum.linalg.solvers.solutionbased import SolutionBasedSolver
 
 # Type aliases
```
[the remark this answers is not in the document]
JonathanWenger: I'd argue for having `pn.<subpackage>.typing` modules for types where we are sure that they are only used within the submodule. Anything else should be in `pn.typing`.

src/probnum/linalg/_problinsolve.py
```diff
@@ -13,12 +13,12 @@
...
 SquareLinOp = Union[
```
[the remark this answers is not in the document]
JonathanWenger: In order to stay consistent with NumPy/Scipy's interface we should support deterministic inputs to the interface functions like `problinsolve`. That type-hint would exclude anything that's not a random variable.

src/probnum/linalg/solvers/_state.py
```diff
@@ -0,0 +1,126 @@
...
+        return self._observations[self.step]
+
+    @observation.setter
+    def observation(self, value: Any) -> None:
+        assert self._observations[self.step] is None
+        self._observations[self.step] = value
+
+    @property
+    def actions(self) -> Tuple[np.ndarray, ...]:
+        """Actions taken by the solver."""
```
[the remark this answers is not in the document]
JonathanWenger: To me "taken" already implies past tense. Since this also gives the most recent action in the _current_ iteration, "past actions" may be conflated with actions from all previous iterations.

src/probnum/linalg/solvers/_state.py
```diff
@@ -0,0 +1,126 @@
...
+        if self._residuals[self.step] is None:
+            self._residuals[self.step] = (
+                self.problem.A @ self.belief.x.mean - self.problem.b
+            )
+        return self._residuals[self.step]
+
+    @property
+    def residuals(self) -> Tuple[np.ndarray, ...]:
+        r"""Residuals :math:`\{Ax_i - b\}_i`."""
+        return tuple(self._residuals)
```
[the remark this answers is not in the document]
JonathanWenger: Currently this would mean some entries in `residuals` will be `None` as for `actions` and `observations`. I don't see this as a problem. Am I missing something?

src/probnum/linalg/solvers/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+from probnum.linalg.solvers.matrixbased import (
+    AsymmetricMatrixBasedSolver,
+    MatrixBasedSolver,
+    ProbabilisticLinearSolver,
+    SymmetricMatrixBasedSolver,
+)
+from probnum.linalg.solvers.solutionbased import SolutionBasedSolver
+
+from . import beliefs, policies
+from ._state import LinearSolverState
```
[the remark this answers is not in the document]
JonathanWenger: While I think from a developer perspective this is a nice idea, i decided against it. The reason was that an inner class doesn't seem to play nice with `automodapi` for the documentation. I couldn't find a way to actually have it show up in the documentation with our current setup. Creating a custom solution just to have `ProbabilisticLinearSolver.State` instead of `ProbabilisticLinearSolverState` is nor worth the time imo.

src/probnum/linalg/solvers/_state.py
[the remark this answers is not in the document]
marvinpfoertner: This might be nasty for an outside user, since it is not really transparent when the residuals are `None` and when they are `np.ndarray`.
I think it's fine for now though.

src/probnum/problems/_problems.py
marvinpfoertner: I think I would prefer the `random_linear_system`, but don't feel obliged to change things if this causes inconvenience.

marvinpfoertner (approved):
Nice :)

Issue #450, opened 2021-07-02
Unify API Reference Structure


JonathanWenger:
# Problem
With the new Sphinx template, the API Reference has lost some of its navigational structure, making it hard for users to understand the package structure.

# Proposed Solution
Across all subpackages of ProbNum, ensure that 
- collapsing sidebar navigation exists for at least two levels `probnum.subpackage.module`
- Solver components are grouped and named correctly

![Screenshot from 2021-07-02 11-20-42](https://user-images.githubusercontent.com/29815984/124252763-9a11d700-db27-11eb-9c5d-c6278c53735e.png)

# ToDo
- [ ] create folders in `docs/source/api` for each subpackage with submodules that should appear in the documentation
- [ ] create `.rst` files with the corresponding `automodapi` directive

PR #461, opened 2021-07-07
Improved API reference structure for `probnum.problems`

JonathanWenger:
# In a Nutshell
Changes the API reference structure for the `probnum.problems` and `probnum.utils` subpackage to make the hierarchy immediately understandable and closer to the actual package hierarchy.

# Detailed Description
Improves and stratifies the sidebar menu for `probnum.problems` in the API reference, like so:
Screenshot from 2021-07-07 17-05-21

# Related Issues
This implements #450 for `probnum.problems`.

### DIFF
--- a/docs/source/api/utils/linalg.rst
+++ b/docs/source/api/utils/linalg.rst
@@ -1,0 +1,6 @@
+probnum.utils.linalg
+====================
+
+.. automodapi:: probnum.utils.linalg
+    :no-heading:
+    :headings: "-"
--- a/docs/source/api/utils/linalg.rst
+++ b/docs/source/api/utils/linalg.rst
@@ -1,6 +1,0 @@
-probnum.utils.linalg
-====================
-
-.. automodapi:: probnum.utils.linalg
-    :no-heading:
-    :headings: "-"
--- a/docs/source/api/problems.rst
+++ b/docs/source/api/problems.rst
@@ -7,27 +7,7 @@ probnum.problems
     :headings: "="
     :no-inheritance-diagram:
 
-probnum.problems.zoo
-====================
+.. toctree::
+    :hidden:
 
-Subpackage offering implementations of standard example problems or convenient
-interfaces to benchmark problems for probabilistic numerical methods. These test
-problems are meant for rapid experimentation and prototyping.
-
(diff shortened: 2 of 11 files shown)


docs/source/api/utils/linalg.rst
```diff
@@ -0,0 +1,6 @@
+probnum.utils.linalg
```
mmahsereci, line 1: Souldn't these go under linalg API? I.e., `pobnum.linalg.utils` instead of `probnum.utils.linalg`? It does't seem right to move sth specific to a general package.

mmahsereci (commented):
Thanks for doing these tedious bits. I like the re-arrangement of the problems docs. Just one comment on where the linalg utils should go.

docs/source/api/utils/linalg.rst
JonathanWenger: Possibly yes. This should be a separate refactoring issue however. Since this is up in the air, I removed the `utils` part of the refactoring.

mmahsereci: Sure, we can discuss that later as well. It might be easier then to leave the utils part in this PR since it is easy to see which util function belong to `linalg`. and we can still see later if we want to move it. Does that make sense?

mmahsereci (approved):

docs/source/api/utils/linalg.rst
pnkraemer: The idea behind moving the Cholesky updates to (the thereby created) `utils.linalg` instead of `linalg.utils` was (a) Cholesky updates are a linear algebra tool for PN algorithms, and not a probabilistic linear solver, and (b) they are needed in e.g. `statespace`, and a linalg dependency for this small thing would've been a bit weird in my opinion. If you have stronger reasons to move it there, though, I am happy to discuss :)

mmahsereci: I see, then simply the namespace is confusing as is may imply that the `utils.linalg` are only required by `linalg` subpackage. If the utils are used by multiple subpackages they should stay where they are (perhaps with a different namespace.) Thanks for the explanation @pnkraemer !

Issue #398, opened 2021-05-04
Tuples in Pytest_cases  


NinaEffenberger:
Currently `@pytest_cases.parametrize_with_cases("a,b", cases=cases)` makes the tests dependant of each other.
see issue https://github.com/smarie/python-pytest-cases/issues/202

Workaround: 

`@pytest_cases.parametrize_with_cases("c", cases=cases) 
def fun(c):
 a,b = c.`

This is used in test_wrappedscipysolver.

NinaEffenberger:
Solved with release 3.5.0 and updated in PR #382.

PR #382, opened 2021-04-23
Perturbed-Step Solver and testing

NinaEffenberger:

PR #463, opened 2021-07-08
Fixed pytest_cases to correct version

JonathanWenger:
# In a Nutshell
#382 introduced test code which necessitates `pytest_cases>=3.5`. See also discussion in #398.


pnkraemer (approved):
Took me a while to get through the big diff, but I finally made it ;) 

Thanks!

PR #464, opened 2021-07-08
Large scale linear solver benchmarks

### LINKED ISSUE #394: Large-scale linear solver benchmarks
# Problem
The problem size for the linear solver benchmarks are too small to be informative (n<100). They therefore are easily misinterpreted, since the construction of linear operators and other operations dominate the overall cost.

# To Do
- [ ] increase problem size of all linear solver benchmarks
- [ ] dense matrices n ~1,000 - 10,000
- [ ] sparse matrices n ~10,000 to 100,000

JonathanWenger:
# In a Nutshell
Introduces larger scale linear solver benchmarks for different types of matrices.

# Detailed Description
Benchmarks the probabilistic linear solver...
- on dense and sparse arrays as well as linear operators
- for problem dimensions up to `n=1000` for dense and up to `n=100.000` with `nnz=1000` for sparse arrays
- by tracking different errors in the solution estimate

# Related Issues
Closes #394


marvinpfoertner (approved):
:+1

But I'm missing a `suitesparse` benchmark :)

JonathanWenger:
> :+1
> 
> But I'm missing a `suitesparse` benchmark :)

I was a little hesitant here since sometimes the SuiteSparse Website is not reachable possibly making the whole benchmark run fail. We sometimes have issues with tests as well. However I do agree that those are probably the real problems we want to benchmark on.

PR #465, opened 2021-07-08
Global Configuration

### LINKED ISSUE #399: Settings as context managers
# Current State and Problem
No package wide settings exist leading to local definitions and switches of variables such as `jitter`, `damping`, etc.

## Use Cases
see below for some examples

# To Do
- [ ] Create a `probnum.settings` module which collects context managers that let you set settings for a code segment.
- [ ] Write a context manager class for settings (see for example [`gpytorch.settings`](https://docs.gpytorch.ai/en/v1.1.1/settings.html))
- [ ] Settings for specific modules should be under `probnum.settings.linalg` or `probnum.settings.filters`

```python
from probnum import settings

with settngs.fast_computations():
     probsolve_ivp(...)
```

Don't start with overly general settings! Only implement what you actually use (in more than one place)

marvinpfoertner:
# In a Nutshell
This PR introduces a global configuration manager.

# Detailed Description
- the global configuration manager is a context-manager

# Related Issues
Closes #399.


schmidtjonathan:
I integrated the documentation of the configuration options in the API reference. This would then look sth. like this:

image

schmidtjonathan:
@JonathanWenger , @pnkraemer , and @mmahsereci , I requested all of you as reviewers, since this is something that impacts everyone who works with `probnum` and should be designed such that everyone is happy, imo. :)

docs/source/api.rst
```diff
@@ -47,3 +49,4 @@ API Reference
     api/randvars
     api/statespace
     api/utils
+    api/config
```
pnkraemer, line 52: just out of curiousity, nothing : why prefer `config` over `settings` as a name? :)

src/probnum/_config.py
```diff
@@ -0,0 +1,83 @@
+import contextlib
+from typing import Any
+
+
+class Configuration:
+    """
+    Configuration over which some mechanics of probnum can be controlled dynamically.
+
+    ``probnum`` provides some configurations together with default values. These
```
pnkraemer, line 9: Super minor: I think "ProbNum" instead of "probnum" is more consistent with the rest of the package here (http://www.probabilistic-numerics.org/en/latest/development/styleguide.html)

src/probnum/_config.py
```diff
@@ -0,0 +1,83 @@
...
+    are listed in the tables below.
+    Additionally, the user can register own configuration entries via the method
+    :meth:`register`. Configuration entries can only be registered once and can only
```
pnkraemer, line 12: How public do we want to make this functionality?

pnkraemer: As in, it obviously exists, but should it appear in the docs like this?

src/probnum/randvars/_config.py
```diff
@@ -0,0 +1,10 @@
+from probnum import config
+
+
+def _register_default_config():
+    _defaults = [
+        ("covariance_inversion_damping", 1e-12),
+    ]
+
+    for k, v in _defaults:
+        config.register(k, v)
```
pnkraemer, line 10: what is the rationale behind registering this parameter from `randvars` and not in top-level? (I think this parameter should be more general, as it would apply to e.g. the Kalman filtering equations, but my question is also about "why not register all the parameters in top-level).

src/probnum/randvars/_normal.py
```diff
@@ -403,8 +404,11 @@ def _sub_normal(self, other: "Normal") -> "Normal":
 
     # Univariate Gaussians
     def _univariate_cov_cholesky(
-        self, damping_factor: Optional[FloatArgType] = COV_CHOLESKY_DAMPING
+        self,
+        damping_factor: Optional[FloatArgType] = None,
     ) -> np.floating:
+        if damping_factor is None:
+            damping_factor = config.covariance_inversion_damping
```
pnkraemer, line 411: I think it is best if in these low-level functions, the damping factor is not optional anymore (the damping factor is overwritten with the global config in "precompute_cov_cholesky" already, which is also why the test coverage is not available here :)

src/probnum/randvars/_normal.py
```diff
@@ -530,10 +537,14 @@ def _dense_entropy(self) -> np.float_:
 
     # Matrixvariate Gaussian with Kronecker covariance
     def _kronecker_cov_cholesky(
-        self, damping_factor: Optional[FloatArgType] = COV_CHOLESKY_DAMPING
+        self,
+        damping_factor: Optional[FloatArgType] = None,
```
pnkraemer, line 541: same here :)

src/probnum/randvars/_normal.py
```diff
@@ -452,10 +456,13 @@ def _univariate_entropy(self: _ValueType) -> np.float_:
 
     # Multi- and matrixvariate Gaussians
     def dense_cov_cholesky(
-        self, damping_factor: Optional[FloatArgType] = COV_CHOLESKY_DAMPING
+        self,
+        damping_factor: Optional[FloatArgType] = None,
```
pnkraemer, line 460: same here :)

src/probnum/randvars/_normal.py
```diff
@@ -552,13 +563,16 @@ def _kronecker_cov_cholesky(
     # factors
     def _symmetric_kronecker_identical_factors_cov_cholesky(
         self,
-        damping_factor: Optional[FloatArgType] = COV_CHOLESKY_DAMPING,
+        damping_factor: Optional[FloatArgType] = None,
```
pnkraemer, line 566: same here :)

src/probnum/_config.py
```diff
@@ -0,0 +1,83 @@
...
+    +----------------------------------+-----------------------------------------------+
+    | Config entry                     | Description                                   |
+    +==================================+===============================================+
+    | ``covariance_inversion_damping`` | A (typically small) value that is per         |
+    |                                  | default added to the diagonal of covariance   |
+    |                                  | matrices in order to make inversion           |
+    |                                  | numerically stable.                           |
+    +----------------------------------+-----------------------------------------------+
+    | ``...``                          | ...                                           |
+    +----------------------------------+-----------------------------------------------+
```
pnkraemer, line 26: I dont understand how/why a subpackage (randvars) is explicitly mentioned here. Is it because this parameter is only used there? Is it because it needs to be accessed with `probnum.config.randvars` or something like this? Would be great if you could explain this to me :)

src/probnum/_config.py
```diff
@@ -0,0 +1,83 @@
...
+
+    - ``probnum.diffeq``
+
+    - ``...``
```
pnkraemer, line 30: what about this? Is it needed?

src/probnum/_config.py
```diff
@@ -0,0 +1,83 @@
...
+                    "accessed."
+                )
+
+            old_entries[key] = getattr(self, key)
+
+            setattr(self, key, value)
+
+        try:
+            yield
+        finally:
```
pnkraemer, line 64: just to clarify I understand @contextlib.contextmanager correctly: this "try, yield, finally" is required by contextlib to turn this into a valid contextmanager, right?

src/probnum/_config.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def __setattr__(self, key: str, value: Any) -> None:
+        if not hasattr(self, key):
+            raise KeyError(
+                f"Configuration entry {key} does not exist yet."
+                "Configuration entries must be `register`ed before they can be "
+                "accessed."
+            )
+
+        self.__dict__[key] = value
```
pnkraemer, line 75: just to clarify my understanding: the difference between this and `settatr(self, key, value)` is that settatr does not throw errors? If not, what is the difference? (this is no real review-comment, but I'd like to learn this :))

src/probnum/randvars/_config.py
pnkraemer: is it because the parameter only becomes available once `randvars` is imported?

tests/test_config.py
```diff
@@ -0,0 +1,87 @@
...
+        self.defaults = [
+            ("covariance_inversion_damping", 1e-12),
+        ]
+
+    @pytest.fixture
+    def zero_cov_normal(self):
+        return randvars.Normal(np.random.rand(5), np.zeros((5, 5)))
+
+    def test_randvars_config(self, zero_cov_normal):
+        chol_1 = zero_cov_normal.dense_cov_cholesky()
```
pnkraemer, line 43: maybe via `precompute_cov_cholesky()` here? I am not sure how public the `dense_cov_cholesky()` was intended to be (but I didnt write it, so I am not the expert here).

pnkraemer (commented):
Thanks a lot for taking care of this @schmidtjonathan  (and @marvinpfoertner). I think it will be great to start working with this!

I left a few comments in the diff, nothing major, I mostly ask you to explain some of the constructs/reasonings to me. From my side this is mostly fine :)

docs/source/api.rst
schmidtjonathan: Personal taste, but I don't have a strong opinion on this one. I'm happy to change it to settings, if you prefer :)

src/probnum/_config.py
schmidtjonathan: Changed it :)

src/probnum/_config.py
schmidtjonathan: I was asking myself the same question. I think there are arguments pro and contra, making it public.
This way the user can utilize this for their own code when using probnum, and I think they cannot break too much.

src/probnum/randvars/_normal.py
schmidtjonathan: Okay

src/probnum/randvars/_config.py
pnkraemer: thanks for the explanation! Is there anything that speaks against registering all parameters that are available "globally" in the toplevel `_config`? I thought that the potential module-wide configs refer to `with diffeq.config(do_something=True)`, i.e. with a separate config in diffeq. Maybe we misunderstood each other.

src/probnum/_config.py
pnkraemer: Hmmm if it is impossible/hard to break something, I think there is nothing wrong with it. Let's see what the others think.

docs/source/api.rst
pnkraemer: I am 55/45 towards settings. Not sufficiently strong to ask for a change ;)

schmidtjonathan: I am 51/49 towards config, so your opinion is stronger so far 😆 Let's ask the others for their opinions

src/probnum/_config.py
schmidtjonathan: That's not strictly necessary. I think we agreed on module-wide configs when discussing this PR, so I mirrored that in the documentation, as well. We can readily change this decision in favor of `ProbNum`-wide configs/settings.

src/probnum/_config.py
schmidtjonathan: It's a stub for config options yet to come.

src/probnum/_config.py
schmidtjonathan: The `yield` is, in particular. The `try: ... finally: ...` is needed in order to restore the previous values correctly, when nesting contextmanagers.

src/probnum/_config.py
pnkraemer: I think this relates to the other discussion above, so I wont repeat my questions/confusions here :)

src/probnum/_config.py
pnkraemer: Sure, but I think with the existing `randvars` entry right above this stub, it does not really serve a purpose. Or do I misunderstand?

src/probnum/_config.py
schmidtjonathan: To the best of my knowledge, this is pretty much all the difference, yes. 

**To make this a review-comment, however:**
We just added this additional check in order to forbid setting random configs from outside, without registering that. 
Although this is as well debatable, imo. So far I am not sure what benefit the need for `register`ing config options brings. @marvinpfoertner has to answer this, since I did not implement this.

My opinion on this: Either forbid the user to add their own options or get rid of the `register`.

docs/source/api.rst
marvinpfoertner: I am 70/30 toward config, since I feel like settings is more often used for UI applications and config(uration) appears more often in lower-level contexts (as in config file)

src/probnum/randvars/_config.py
marvinpfoertner: I think we should try to only have one `config` object in which all subpackages inject their settings, i.e. for the `diffeq` config options you would add the settings to the global object, but the registration calls should still be made in `diffeq/_config.py`

marvinpfoertner:
> I integrated the documentation of the configuration options in the API reference. This would then look sth. like this:
> 
> image
> 
> I will then also include a description and an example, how users can register their own configuration options for their personal use. The default options in `probnum` (as in this example: `covariance_inversion_damping`, the nugget which is added to PSD matrices for numerical stability) are will then be documented and maintained by us.
> Let me know what you think.

How about we import `Configuration` in `probnum/__init__.py` and then add `Configuration.__module__ = "probnum"` to remove the hidden package structure?

src/probnum/randvars/_config.py
pnkraemer: Why wouldnt I add the settings in the global config then? What is the advantage of doing it in diffeq (if we stay with this example)? (Not sure whether "global config" is the proper term, but you know what I mean.)

schmidtjonathan:
> > I integrated the documentation of the configuration options in the API reference. This would then look sth. like this:
> > [...]
> > I will then also include a description and an example, how users can register their own configuration options for their personal use. The default options in `probnum` (as in this example: `covariance_inversion_damping`, the nugget which is added to PSD matrices for numerical stability) are will then be documented and maintained by us.
> > Let me know what you think.
> 
> How about we import `Configuration` in `probnum/__init__.py` and then add `Configuration.__module__ = "probnum"` to remove the hidden package structure? @schmidtjonathan Do you think this would introduce problems?

Wouldn't this simplify creating multiple Configuration objects? Which we want to avoid by making it protected?

src/probnum/_config.py
schmidtjonathan: Only if we want to have module-wide configs and also document them. Then, the intention is to have a list of tables for the possible configs.
But -- again -- I see your point, in that I also find that debatable, and would -- at least for now -- agree to implement probnum-wide (global) configs only.

src/probnum/_config.py
marvinpfoertner: The `try: ... finally: ...` is not strictly necessary when nesting context managers but highly recommended, since otherwise an exception thrown before returning to the context manager function after the `yield` would mess up the consistency of the settings.
So the TL;DR is, if one wants to use exceptions, one should use the `try: ... finally: ...` construct in context managers :) @pnkraemer

src/probnum/_config.py
marvinpfoertner: The idea behind the `register` is to force `probnum` developers to register the "semantics" of a configuration option "statically", by which I mean as module-level code which gets executed while importing `probnum`.
This will prevent random scattered writes to the `pn.config` and provide central places in the code, where we get an overview of all registered probnum configuration options

Also, having a separate `register` call allows us to document configuration options automatically in the future.

A bit of context: The idea behind first registering (with a default value), then writing/setting/changing a configuration option comes from experience I had with a similar system, where people started to write new config keys in the middle of the library and in the end, nobody knew that certain config options already existed. It was a mess.

src/probnum/randvars/_config.py
marvinpfoertner: One reason might be the tendency to produce merge conflicts when working in one large config file.
Another one is the possibility to not register a subpackage's options, should we decide in the future that we do not automatically import all subpackages in `probnum/__init__.py` (similar to scipy).

Moreover, this way we increase code locality while maintaining a single interface towards the end-user of the library

marvinpfoertner: But I'm not set on the idea of just having one config object.
I just fear that we end up producing a mess (from the perspective of an end-user).

tests/test_config.py
marvinpfoertner: `dense_cov_cholesky` is meant to be this public :) It helps to reduce code if you're not shure whether the covariance is an `ndarray` or a `LinearOperator`

src/probnum/_config.py
marvinpfoertner: Another option might be to add the tables to the landing pages of the corresponding subpackages

marvinpfoertner:
> > > I integrated the documentation of the configuration options in the API reference. This would then look sth. like this:
> > > [...]
> > > I will then also include a description and an example, how users can register their own configuration options for their personal use. The default options in `probnum` (as in this example: `covariance_inversion_damping`, the nugget which is added to PSD matrices for numerical stability) are will then be documented and maintained by us.
> > > Let me know what you think.
> > 
> > 
> > How about we import `Configuration` in `probnum/__init__.py` and then add `Configuration.__module__ = "probnum"` to remove the hidden package structure? @schmidtjonathan Do you think this would introduce problems?
> 
> Wouldn't this simplify creating multiple Configuration objects? Which we want to avoid by making it protected?

Yeah that's a valid concern.
I just feel like having the `probnum._config` prefix in the docs might be worse. But that's just personal taste.

tests/test_config.py
pnkraemer: ayt

src/probnum/randvars/_config.py
pnkraemer: Hmmm okay I see what you mean. Thanks for explaining the thought process :) 

> One reason might be the tendency to produce merge conflicts when working in one large config file.

I see this as a potential problem, but I dont think the problem would be severe. I don't expect the config file to be touched often (though this is speculation, ofc). 

> Another one is the possibility to not register a subpackage's options, should we decide in the future that we do not automatically import all subpackages in probnum/__init__.py (similar to scipy).

Okay, but if all the configs are gathered in one file, it is easy to move them to the subpackages if this decision is made at some point, right?

> Moreover, this way we increase code locality while maintaining a single interface towards the end-user of the library

I agree that the latter is useful. But I personally think that not setting the parameters that are available globally right in the global scope will also make the implementation more tedious to work with; at least in the beginning (as in "where has this parameter been set? What does it mean?" or even "do we have this yet"). 
From a different angle: how would we try to enforce that some parameter that has been registered in `randvars` is not used somewhere else (we dont have to, but it might help avoiding a mess)?

In principle, these downsides could of  course be solved with more extensive docs, but maybe this is not necessary.

All in all, I see the thought process behind this -- thanks for explaining. I still think that registering all options in the global config (at least for now) is the simpler and perhaps more appropriate solution for the moment. Of course, this is subjective.

src/probnum/_config.py
pnkraemer: Please keep on discussing, but my comment was much more minor than that :D I just wanted to point out that the line 28 serves no purpose in terms of "you would put the info here", because an example is already given with randvars :)

src/probnum/_config.py
schmidtjonathan: >  people started to write new config keys in the middle of the library

I see what you mean. But cannot then people just add random `.register()` calls in the middle of nowhere, instead? Since register does not add any functionality, other than checking for duplicate `register`s

marvinpfoertner: Yes that's true, but I think the threshold for actually doing that is much higher than just writing something like `config.option = value`.

schmidtjonathan:
@JonathanWenger @pnkraemer @mmahsereci @marvinpfoertner 
We need a decision on the following points:

- [ ] `settings` vs. `config`
- [ ] module-level configs vs. global, probnum-wide configs
  - [ ] Do we want calls like `diffeq.config(x=y)` or only `probnum.config(x=y)`?
  - [ ] Also, where do we _register_ the config? Right now, we only ever call `probnum.config(x=y)`, but the config options are registered in the respective modules.

src/probnum/_config.py
schmidtjonathan: That's no example. That's the documentation. The bullet point `diffeq` will then also be filled with a respective table full of config options and their descriptions.

pnkraemer: ah okay. My bad.

src/probnum/_config.py
```diff
@@ -0,0 +1,91 @@
+import contextlib
+from typing import Any
+
+
+class Configuration:
+    """
+    Configuration over which some mechanics of probnum can be controlled dynamically.
```
JonathanWenger, line 7:
```suggestion
    Configuration over which some mechanics of ProbNum can be controlled dynamically.
```

src/probnum/_config.py
```diff
@@ -0,0 +1,91 @@
...
+
+    ``ProbNum`` provides some configurations together with default values. These
```
JonathanWenger, line 9:
```suggestion
    ProbNum provides some configurations together with default values. These
```

src/probnum/_config.py
```diff
@@ -0,0 +1,91 @@
...
+    are listed in the tables below.
+    Additionally, the user can register own configuration entries via the method
```
JonathanWenger, line 11:
```suggestion
    Additionally, users can register their own configuration entries via 
```

docs/source/api.rst
JonathanWenger: I'd vote for `config` as well.

src/probnum/_config.py
```diff
@@ -0,0 +1,91 @@
...
+                "cannot be registered again."
+            )
+        self.__dict__[key] = default_value
+
+
+# Create a single, global configuration object,...
+_GLOBAL_CONFIG_SINGLETON = Configuration()
+
+# ... define some configuration options, and the respective default values
+# (which have to be documented in the Configuration-class doctstring!!), ...
```
JonathanWenger, line 83:
```suggestion
# (which have to be documented in the Configuration-class docstring!!), ...
```

JonathanWenger: I assume there is no good way of documenting these options here automatically (including their default values), is there? Manually updating entries and their value in the documentation is prone to breaking frequently.

src/probnum/_config.py
```diff
@@ -0,0 +1,91 @@
...
+    ``ProbNum`` provides some configurations together with default values. These
+    are listed in the tables below.
+    Additionally, the user can register own configuration entries via the method
+    :meth:`register`. Configuration entries can only be registered once and can only
+    be used (accessed or overwritten) once they have been registered.
+
+    +----------------------------------+-----------------------------------------------+
+    | Config entry                     | Description                                   |
+    +==================================+===============================================+
+    | ``covariance_inversion_damping`` | A (typically small) value that is per         |
```
JonathanWenger, line 18: It would be good to have the default values here for documentation purposes. I realize this may be problematic with changes that do not update the documentation.

docs/source/api.rst
```diff
@@ -4,31 +4,33 @@ API Reference
...
+    +-----------------------------------------+--------------------------------------------------------------+
+    | :mod:`~probnum.randprocs`               | Random processes representing uncertain functions.           |
+    +-----------------------------------------+--------------------------------------------------------------+
+    | :mod:`~probnum.randvars`                | Random variables representing uncertain values.              |
+    +-----------------------------------------+--------------------------------------------------------------+
+    | :mod:`~probnum.statespace`              | Probabilistic state space models.                            |
+    +-----------------------------------------+--------------------------------------------------------------+
+    | :mod:`~probnum.utils`                   | Utility functions.                                           |
+    +-----------------------------------------+--------------------------------------------------------------+
+    | :class:`~probnum._config.Configuration` | Global configuration options.                                |
```
JonathanWenger, line 32: I would suggest to have `probnum.config` here for consistency. I assume you used the class to not break linking, but this breaks the pattern of the modules here linking to the sidebar menu entries

JonathanWenger (changes requested):
> @JonathanWenger @pnkraemer @mmahsereci @marvinpfoertner
> We need a decision on the following points:
> 
>     * [ ]  `settings` vs. `config`
I vote for `config`.
>     * [ ]  module-level configs vs. global, probnum-wide configs
I think having ~module~**package**-level configs is simpler and also easier to maintain and understand for the user for now. Having only ~module~**package**-wide configs has a few advantages like
- it makes it more restrictive what actually should be a config, which avoids devs registering everything as a config option
- it forces one to think about combining configs and avoids maintaining similar configs across subpackages   
>       * [ ]  Do we want calls like `diffeq.config(x=y)` or only `probnum.config(x=y)`?
>       * [ ]  Also, where do we _register_ the config? Right now, we only ever call `probnum.config(x=y)`, but the config options are registered in the respective modules.

schmidtjonathan:
> * avoids maintaining similar configs across subpackages

I don't agree with this point. I think this is the main disadvantage of module-wide configs: that similar/equal configs might be needed in multiple modules and might end up being defined/registered multiple times.
But with the rest of your points, I agree.

src/probnum/_config.py
schmidtjonathan: > I assume there is no good way of documenting these options here automatically (including their default values), is there?

Not that I know of.

> Manually updating entries and their value in the documentation is prone to breaking frequently.

Agreed 100%. Still - in my opinion - better than a half-baked automated solution.

docs/source/api.rst
schmidtjonathan: Not sure, if I understand you, correctly. Do you mean sth like this?

```
:mod:`probnum.config <~probnum._config.Configuration>`
```

schmidtjonathan: Because that doesn't work 😆 This way, the side bar works (as before) but in the table, the link is broken.

JonathanWenger:
> > * avoids maintaining similar configs across subpackages
> 
> I don't agree with this point. I think this is the main disadvantage of module-wide configs: that similar/equal configs might be needed in multiple modules and might end up being defined/registered multiple times.
> But with the rest of your points, I agree.

Maybe I should have had a coffee first. I actually meant to write **package-level** config.

docs/source/api.rst
JonathanWenger: How about this?
```suggestion
    | :class:`config <~probnum._config.Configuration>` | Global configuration options.                                |
```
That way the link should work, even though the destination is the `Configuration` class. In practice users will want to see that doc page anyway.

schmidtjonathan:
> > > * avoids maintaining similar configs across subpackages
> > 
> > 
> > I don't agree with this point. I think this is the main disadvantage of module-wide configs: that similar/equal configs might be needed in multiple modules and might end up being defined/registered multiple times.
> > But with the rest of your points, I agree.
> 
> Maybe I should have had a coffee first. I actually meant to write **package-level** config.

Know that feeling ☕ 
Should we make a clear decision then? I am quite torn and would say ... 60/40 in favor of package-wide config, for now. Also because an eventual transition to module-wide configs will be easier than the other way round.

docs/source/api.rst
JonathanWenger: And this does not work?
```suggestion
    | :mod:`config <~probnum._config` | Global configuration options.                                |
```

JonathanWenger: I assume above the link is broken since `probnum._config.Configuration` is not a module.

src/probnum/_config.py
JonathanWenger: Then I suggest to manually add the default values to the table in the docstring of `Configuration`. That way they are exposed to the user.

docs/source/api.rst
schmidtjonathan: > And this does not work?

For some reason, it does not. Weirdly, the link in the sidebar does work, the link in the table, does not.

JonathanWenger: Then i would suggest to just rename in the table, i.e. to use:

```suggestion
    | :class:`config <~probnum._config.Configuration>` | Global configuration options.                                |
```

schmidtjonathan: For some reason, the renaming breaks the link in the table.

schmidtjonathan: Ah, the tilde was the problem. Fixed it

src/probnum/_config.py
```diff
@@ -4,24 +4,24 @@
...
-    +==================================+===============================================+
-    | ``covariance_inversion_damping`` | A (typically small) value that is per         |
-    |                                  | default added to the diagonal of covariance   |
-    |                                  | matrices in order to make inversion           |
-    |                                  | numerically stable.                           |
-    +----------------------------------+-----------------------------------------------+
-    | ``...``                          | ...                                           |
-    +----------------------------------+-----------------------------------------------+
+    +----------------------------------+----------------------------------------------+
+    | Config entry                     | Description                                  |
```
JonathanWenger, line 16: I sugges

```python
"""
+----------------------------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------+
| Config entry                     | Default value | Description                                                                                                                               |
+==================================+===============+===========================================================================================================================================+
| ``covariance_inversion_damping`` | 1e-12         | A (typically small) value that is per default added to the diagonal of covariance matrices in order to make inversion numerically stable. |
+----------------------------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------+
| ``...``                          | ``...``       | ...                                                                                                                                       |
+----------------------------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------+
"""
```

JonathanWenger (approved):
Great job! Thanks for introducing the `config`s. This will undoubtedly be useful. I have one minor suggestion for the docstring of `Configuration`, otherwise this is good to go from my side.

src/probnum/_config.py
marvinpfoertner: I'm hacking together a quick sphinx extension that generates the documentation automatically atm.
I will include the default value column there.

pnkraemer (approved):
Thanks a lot for doing this Jonathan!

schmidtjonathan:
Thanks for your support, @JonathanWenger, @pnkraemer , @marvinpfoertner !
I added the change, you requested, @JonathanWenger , and added a docstring to register. As soon as the checks have passed, I merge :)

PR #468, opened 2021-07-09
Website Navigation Bar Update

JonathanWenger:
# In a Nutshell
Renames the navigation option on the landing page from "Research" to "Probabilistic Numerics".

### DIFF
--- a/docs/source/_templates/footer.html
+++ b/docs/source/_templates/footer.html
@@ -8,7 +8,7 @@
           <h4>Probabilistic Numerics</h4>
           <ul class="nav flex-column">
             <li class="nav-item">
-              <a class="nav-link" href="{{ pathto('research') }}#probabilistic-numerics-in-a-nutshell">In a Nutshell</a>
+              <a class="nav-link" href="{{ pathto('research') }}#in-a-nutshell">In a Nutshell</a>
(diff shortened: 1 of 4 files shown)


philipphennig:
Thanks! As requested by Mike.

mmahsereci (approved):

philipphennig:
Looks good to me. 

I get an error message from `dot` when building with `tox`, but can't confirm whether this is just a local problem for me. The change in this PR is definitely good. Thanks!

JonathanWenger:
> Looks good to me.
> 
> I get an error message from `dot` when building with `tox`, but can't confirm whether this is just a local problem for me. The change in this PR is definitely good. Thanks!

Is it possible you are missing `graphviz` (see http://www.probabilistic-numerics.org/en/latest/development/pull_request.html#tox)?
```bash
sudo apt install graphviz
```
This is needed to generate the inheritance diagrams. While `tox` almost reproduces the CI exactly on a local machine for system dependent installation of packages we do not yet have a good solution.

philipphennig:
Thanks for the info. I fixed the issue now (my graphviz install was indeed outdated, I haven’t used it in years). Compiles correctly now.

PR #475, opened 2021-07-12
Move ODE example problems into the problem zoo

### LINKED ISSUE #312: Problem classes in diffeq
The code in diffeq needs to be refactored to use the problem classes in `probnum.problems`, concretely the `InitialValueProblem` class.

This entails the following changes:
* `ODESolver.solve` gets the problem as an argument (in this case, the IVP). In the respective inits, everything else is passed. This also applies to all of its subclasses, including the filter.
* The `ODE` and `IVP` classes are replaced entirely by the problem class. This implies:
* The IVP examples need to be rewritten in terms of implementing the problem class (and not `IVP` anymore)
* I expect `probsolve_ivp` to have a signature leading with an "unpacked" `InitialValueProblem`. That means `probsolve_ivp(f, t0, tmax, y0, df, **other_parameters_like_steps_etc)` 
* Initialisation of the higher order derivatives of the IVP is then best handled by "extending" the problem class so it includes all higher order derivatives (these guys are data, not self-generated). This however will be the question for a future PR, and slightly out of scope here.

pnkraemer:
# In a Nutshell
* No more `diffeq.ODE` and `diffeq.IVP` objects (only `problems.InitialValueProblems` now)
* ODE example problems `diffeq.lotkavolterra` are moved to `problems.zoo.diffeq`.

# Related Issues
Closes #312


schmidtjonathan (approved):
Thanks!

PR #477, opened 2021-07-13
Reorganise `diffeq` packages

pnkraemer:
# In a Nutshell
The title says it all.

# Detailed Description
This PR organises the functionality in the `diffeq` package into general objects, filtering-based solvers, perturbation-based solvers, and more. See the respective discussion.
This PR only moves around objects. No functionality is touched/refactored.


src/probnum/diffeq/__init__.py
```diff
@@ -1,48 +1,22 @@
...
-    perturb_uniform,
-)
-from .steprule import AdaptiveSteps, ConstantSteps, StepRule, propose_firststep
-from .wrappedscipysolver import WrappedScipyRungeKutta
+# Publicly visible subpackage "perturbed" needs to be imported here
+# Without this line, a double-subpackage import as in
+#     import diffeq
+#     solver = diffeq.perturbed.step.PerturbedStepSolver(...)
+# does not work.
+from . import perturbed
```
pnkraemer, line 11: This is annoying. I am all ears for a better solution.

src/probnum/diffeq/_probsolve_ivp.py
```diff
@@ -33,7 +32,7 @@ def probsolve_ivp(
     step=None,
     diffusion_model="dynamic",
 ):
-    r"""Solve initial value problem with Gaussian filtering and smoothing.
+    r"""Solve an initial value problem with a probabilistic ODE solver, i.e. with Gaussian filtering and smoothing.
```
nathanaelbosch, line 35: Instead of "probabilistic ODE solver", should we specify "filtering-based" or something now that we also have the sampling/perturbation-based probabilistic ODE solvers?

nathanaelbosch (commented):
Looks good! Two things:
1. I don't know sphinx enough to judge if the *.rst changes should be there or not
2. What do you think about renaming `scipy_wrapper` to `scipy_wrappers`? It seems to contain two wrappers, for the solver and the solution. I'm not sure though what convention is used in most cases, so if we typically do not use the plural for stuff like that then we should keep it how it currently is. 

Thanks for the PR!

src/probnum/diffeq/_probsolve_ivp.py
pnkraemer: My idea here was to nudge the narrative slowly towards filtering-based solvers being **the** probabilistic ODE solver :)

pnkraemer:
Thanks for the review!

1. Yes, they should be there (otherwise the subpackages would not appear in the docs)
2. I followed the convention (https://softwareengineering.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural) to use singular for a package with heterogeneous content (here: wrappers for both solution and solver) as opposed to plural for a package with homogeneous content. However, I am aware that one can certainly argue for a bunch of scipy wrappers being homogeneous. What do you think?

src/probnum/diffeq/__init__.py
```diff
@@ -1,48 +1,22 @@
 """Differential Equations.
 
-This package defines common dynamical models and probabilistic solvers
-for differential equations.
+This package defines probabilistic solvers for differential equations.
```
schmidtjonathan, line 3:
```suggestion
This package provides probabilistic solvers for differential equations.
```

schmidtjonathan: or `implements` ?

src/probnum/diffeq/_probsolve_ivp.py
schmidtjonathan: Agreed, the `i.e., ...` - part suggests that `probabilistic Ode solver == filtering-based ODE solver`.

src/probnum/diffeq/perturbed/__init__.py
```diff
@@ -0,0 +1,3 @@
+"""Perturbed solvers."""
```
schmidtjonathan, line 1:
```suggestion
"""Perturbation-based solvers."""
```

src/probnum/diffeq/perturbed/scipy_wrapper/_wrapped_scipy_odesolution.py
```diff
@@ -6,11 +6,11 @@
 from probnum.filtsmooth._timeseriesposterior import DenseOutputValueType
 from probnum.typing import DenseOutputLocationArgType
 
-from .odesolution import ODESolution
+from ..._odesolution import ODESolution
```
schmidtjonathan, line 9: Correct me if I am wrong, but I thought those relative imports of modules further up in the hierarchy, are discouraged 😬

src/probnum/diffeq/perturbed/step/_perturbedstepsolution.py
```diff
@@ -8,11 +8,11 @@
 from probnum import _randomvariablelist, randvars
 from probnum.typing import FloatArgType
 
-from ..odesolution import ODESolution
+from ..._odesolution import ODESolution
```
schmidtjonathan, line 11: See above

src/probnum/diffeq/perturbed/step/_perturbedstepsolver.py
```diff
@@ -7,13 +7,13 @@
 from probnum import _randomvariablelist, randvars
 from probnum.typing import FloatArgType
 
-from ..odesolver import ODESolver
-from ..wrappedscipysolver import WrappedScipyRungeKutta
+from ..._odesolver import ODESolver
+from .. import scipy_wrapper
```
schmidtjonathan, line 11: See above

src/probnum/diffeq/_probsolve_ivp.py
nathanaelbosch: As much as I like this idea, I'm not sure how people outside of our group agree with this assessment?

But apart from that wider discussion: The solvers in `perturbed` are also probabilistic ODE solvers, and I imagine that tutorials and the documentation will describe them as such. To say that `probsolve_ivp` solves an IVP with "a probabilistic ODE solver" seems to me a bit misleading then, since it might suggest that sampling-based solvers could be used here.

schmidtjonathan: Oh, @pnkraemer , your comment wasn't displayed, when I wrote my answer.
I see your point as well, it just depends on the message we want to send. As a package which offers different kinds of probabilistic ODE solvers as well, this may be ... inconsistent I guess.

schmidtjonathan:
> 2. What do you think about renaming `scipy_wrapper` to `scipy_wrappers`? It seems to contain two wrappers, for the solver and the solution. I'm not sure though what convention is used in most cases, so if we typically do not use the plural for stuff like that then we should keep it how it currently is.

Or maybe `wrap_scipy`, actually. That's what I thought after our meeting yesterday, but felt it too minor to bring up again.

src/probnum/diffeq/_probsolve_ivp.py
nathanaelbosch: Whoops, was that "i.e. ..." part there all the time? Then my brain somehow completely ignored that. With that addition everything is totally fine, sorry for the discussion I started!

nathanaelbosch:
>     2. I followed the convention (https://softwareengineering.stackexchange.com/questions/75919/should-package-names-be-singular-or-plural) to use singular for a package with heterogeneous content (here: wrappers for both solution and solver) as opposed to plural for a package with homogeneous content. However, I am aware that one can certainly argue for a bunch of scipy wrappers being homogeneous. What do you think?

I don't really know if that's homogeneous or heterogeneous. Knowing that you thought about it and decided on the singular is enough for me, so feel free to keep it the way it is :)

src/probnum/diffeq/perturbed/scipy_wrapper/_wrapped_scipy_odesolution.py
pnkraemer: I dont know. I couldnt find anything. Though the google style guide says "use absolute paths all the time". This way we would get syntax like (in `_ivpfiltsmooth.py`)
```python

from probnum import filtsmooth, problems, randprocs, randvars, statespace, utils

from probnum.diffeq import _odesolver
from probnum.diffeq.odefiltsmooth import _kalman_odesolution, initialize
```
If I import ODESolver and KalmanODESolution from probnum.diffeq(.odefiltsmooth), I get circular imports. I am fine with the above though, shall I change it? I think it would clear things up in terms of imports.

src/probnum/diffeq/_probsolve_ivp.py
pnkraemer: Hmmm okay, I change it to 
```
    r"""Solve an initial value problem with a (filtering-based) probabilistic ODE solver.
```
for now (I understand your points), but the last word has not been spoken here! ;)

nathanaelbosch (approved):
From my side everything looks good :+1:

schmidtjonathan (approved):
Thanks, @pnkraemer !

PR #478, opened 2021-07-13
Restructure `filtsmooth`

pnkraemer:
# In a Nutshell
Reorganise the package structure in filtsmooth.

# Detailed Description
filtsmooth is now structured into `gaussian`, `particle`, `optim`, and `utils`. The IEKS is moved to `optim` (and becomes `GaussNewton`). There is now a convenience interface `kalman_filter`, and `rauch_tung_striebel_smoother` for linear systems.

# Note
This bases on #477. So no need to review before this is merged.


src/probnum/filtsmooth/_kalman_filter_smoother.py
```diff
@@ -0,0 +1,192 @@
...
+
+import numpy as np
+
+from probnum import problems, randprocs, randvars, statespace
+from probnum.filtsmooth import gaussian
+
+__all__ = ["kalman_filter", "rauch_tung_striebel_smoother"]
+
+
+def kalman_filter(
```
schmidtjonathan, line 11:
```suggestion
def filter_kalman(
```

I think it is better to name a function after what it does (i.e. using a verb if possible). It does return a result, not a Kalman filter.

src/probnum/filtsmooth/_kalman_filter_smoother.py
```diff
@@ -0,0 +1,192 @@
...
+        H=H, R=R, observations=observations, locations=locations
+    )
+    prior_process = _setup_prior_process(
+        F=F, L=L, m0=m0, C0=C0, t0=locations[0], prior_model=prior_model
+    )
+    kalman = gaussian.Kalman(prior_process)
+    return kalman.filter(regression_problem)[0]
+
+
+def rauch_tung_striebel_smoother(
```
schmidtjonathan, line 87:
```suggestion
def smooth_rts(
```

Same argument as above, and I think `rts` is ok, instead of `rauch_tung_striebel` if we write it out in the docstring headline. But that's debatable :)

src/probnum/filtsmooth/_kalman_filter_smoother.py
```diff
@@ -0,0 +1,192 @@
+"""Convenience functions for filtering and smoothing."""
```
schmidtjonathan, line 1: All in all I think these functions will improve usability a lot! Thanks for adding them.
Aside from the comments I made below: Should we add `Example`s to the respective docstrings?

src/probnum/filtsmooth/optim/_gauss_newton.py
```diff
@@ -0,0 +1,35 @@
+"""Gauss-Newton methods in state-space models."""
```
schmidtjonathan, line 1: I think it's good that you moved this out of `_kalman.py`, besides the other "regular" Kalman filter functionalities. 👍 

Do you think the name `optim` for the package could suggest that we optimize - for instance - hyperparameters for state-space model regression, or sth like that? Is it clear enough that the content of the package **is** itself an optimizer, rather than it **does** optimize sth.? (minor remark)

src/probnum/filtsmooth/optim/_state_space_optimizer.py
```diff
@@ -0,0 +1,32 @@
+"""Gauss-Newton methods in state-space models."""
+
+import abc
+
+from probnum.filtsmooth.optim import _stoppingcriterion
+
+
+class StateSpaceOptimizer(abc.ABC):
```
schmidtjonathan, line 8: > Do you think the name optim for the package could suggest that we optimize - for instance - hyperparameters for state-space model regression, or sth like that? Is it clear enough that the content of the package is itself an optimizer, rather than it does optimize sth.? (minor remark)

Same concern here

src/probnum/filtsmooth/_kalman_filter_smoother.py
```diff
@@ -0,0 +1,192 @@
...
+import numpy as np
+
+from probnum import problems, randprocs, randvars, statespace
+from probnum.filtsmooth import gaussian
+
+__all__ = ["kalman_filter", "rauch_tung_striebel_smoother"]
+
+
+def kalman_filter(
+    observations, locations, F, L, H, R, m0, C0, prior_model="continuous"
```
schmidtjonathan, line 12: Do you think, using `drift`, `dispersion`, etc. would be more consistent and more clear than using `F`, `L`, etc.?

schmidtjonathan (changes requested):
Thanks for taking care of this.
Other than renaming the utility functions for Kalman filtering and RTS smoothing, I made only minor remarks. Those two renamings, though, I think, would benefit the usability/clarity a lot.

src/probnum/filtsmooth/_kalman_filter_smoother.py
schmidtjonathan: More consistent, due to the attributes `driftmat` and `dispmat` etc. in the `Transition`s

pnkraemer: Yes, I was thinking this before. But perhaps it is easier to parse this interface for a user, for instance because other implementations of similar algorithms do this (e.g. matlab: https://uk.mathworks.com/help/control/ug/kalman-filtering.html). I can even change the letters to match the matlab style?!

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: I agree. I dont particularly like "filter_kalman" though... how about `filter(..., method="kalman")`? This would allow extending this interface to e.g. ensemble kalman filters or the like. I think the fact that this redefines a built-in, is not *the* worst thing in the world, as one would probably use it as `filtsmooth.filter(...)`. Any thoughts in this direction? A potential extension of this convenience function to nonlinear system might motivate naming as in `filter_linear_system(..., method="kalman")`?

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: Could be done, but I was too lazy. Should I do it?

src/probnum/filtsmooth/optim/_gauss_newton.py
pnkraemer: I have a hunch that we might even want to think about moving hyperparameter optimisation in here. This stems from the notion that e.g. EM updates to, for instance, the initial distribution parameters, are a one-linear change to an IEKS. Perhaps we *want* to blur the lines here?! I dont know.

src/probnum/filtsmooth/optim/_state_space_optimizer.py
pnkraemer: Maybe this problem is resolved if I mention in the first docstring line that the optimiser -- in the current concept --  computes states, not parameters?

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: Agreement: `filter_kalman`

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: Agreement: `smooth_rts`

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: Agreement: Not today

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: @schmidtjonathan this one has not been discussed yet :)

schmidtjonathan: Mh, personally, if I had to choose between F, L, etc. and the Matlab notation, I would choose the former. Additionally, if we start following Matlab conventions, we probably have to do it everywhere, right? 😅
From my side, we can just leave it as is, as long as it is well documented :)

src/probnum/filtsmooth/optim/_state_space_optimizer.py
schmidtjonathan: Yep, that would help, thanks! :)

src/probnum/filtsmooth/_kalman_filter_smoother.py
pnkraemer: I agree. The Matlab comparison was just to motivate my choice of single-letter variables here :)

schmidtjonathan (approved):
Thanks!

PR #480, opened 2021-07-13
Auto-generate documentation for `probnum.config` options

marvinpfoertner:
# In a Nutshell
This PR implements a custom Sphinx extension which generates a table documenting the available configuration options automatically.

# Detailed Description
- [x] Sphinx extension generating configuration option documentation automatically
- [x] add per-option docstrings to `probnum.config` via a required (?) `docstring` argument in the `register` method
- [x] store configuration options as inner dataclasses `Configuration.Option` with fields `key`, `value`, `default_value`, and ~`docstring`~ `description`
- ~[ ] add iterator over registered `Configuration.Option`s~

### DIFF
--- a/docs/source/_ext/probnum-config-options.py
+++ b/docs/source/_ext/probnum-config-options.py
@@ -35,17 +35,15 @@ class ProbNumConfigOptions(SphinxDirective):
         tbody = nodes.tbody("")
         group.append(tbody)
 
-        for key, value in pn.config.__dict__.items():
+        for config_option in pn.config.__dict__.values():
             row = nodes.row("")
             tbody.append(row)
 
-            default_value = value
-            description = (  # TODO: Read this from the configuration registry
-                "A (typically small) value that is per default added to the diagonal "
-                "of covariance matrices in order to make inversion numerically stable."
-            )
+            name = config_option.name
+            default_value = config_option.default_value
+            description = config_option.description
 
-            row.append(nodes.entry("", nodes.literal(text=key)))
+            row.append(nodes.entry("", nodes.literal(text=name)))
             row.append(nodes.entry("", nodes.literal(text=repr(default_value))))
             row.append(nodes.entry("", self._parse_string(description)))
 
--- a/docs/source/_ext/probnum-config-options.py
+++ b/docs/source/_ext/probnum-config-options.py
@@ -10,11 +10,15 @@ class ProbNumConfigOptions(SphinxDirective):
 
     The ``name``s, ``default_value``s, and ``description``s of the pre-defined
     ProbNum configuration options are read and dynamically added to a table in the docs.
-    This plugin avoids maintaining the table in the config docstring manually.
+    This plugin [1]_ avoids maintaining the table in the config docstring manually.
 
     See Also
     --------
     probnum.config : Register and set configs to control specific ProbNum behavior.
+
+    References
+    ----------
+    .. [1] ``https://www.sphinx-doc.org/en/master/development/tutorials/helloworld.html``
     """
 
     def run(self):
--- a/docs/source/_ext/probnum-config-options.py
+++ b/docs/source/_ext/probnum-config-options.py
@@ -35,7 +35,7 @@ class ProbNumConfigOptions(SphinxDirective):
         tbody = nodes.tbody("")
         group.append(tbody)
 
-        for config_option in pn.config.__dict__.values():
+        for config_option in pn.config._options_registry.values():
             row = nodes.row("")
             tbody.append(row)
 
--- a/docs/source/_ext/probnum-config-options.py
+++ b/docs/source/_ext/probnum-config-options.py
@@ -6,6 +6,8 @@ import probnum as pn
 
 
 class ProbNumConfigOptions(SphinxDirective):
+    """Sphinx plugin that automatically generates the docs for ProbNum config options."""
+
     def run(self):
         table = nodes.table(
             "",
--- a/docs/source/_ext/probnum-config-options.py
+++ b/docs/source/_ext/probnum-config-options.py
@@ -11,6 +11,10 @@ class ProbNumConfigOptions(SphinxDirective):
     The ``name``s, ``default_value``s, and ``description``s of the pre-defined
     ProbNum configuration options are read and dynamically added to a table in the docs.
     This plugin avoids maintaining the table in the config docstring manually.
+
+    See Also
+    --------
+    probnum.config : Register and set configs to control specific ProbNum behavior.
     """
 
     def run(self):
--- a/docs/source/_ext/probnum-config-options.py
+++ b/docs/source/_ext/probnum-config-options.py
@@ -6,7 +6,12 @@ import probnum as pn
 
 
 class ProbNumConfigOptions(SphinxDirective):
-    """Sphinx plugin that automatically generates the docs for ProbNum config options."""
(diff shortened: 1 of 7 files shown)


docs/source/_ext/probnum-config-options.py
```diff
@@ -0,0 +1,82 @@
...
+
+        for key, value in pn.config.__dict__.items():
+            row = nodes.row("")
+            tbody.append(row)
+
+            default_value = value
+            description = (  # TODO: Read this from the configuration registry
+                "A (typically small) value that is per default added to the diagonal "
+                "of covariance matrices in order to make inversion numerically stable."
+            )
```
marvinpfoertner, line 46: Here we have to
- [ ] use the default value instead of the value in the documentation
- [ ] read the description from the global `Configuration` object

src/probnum/_config.py
```diff
@@ -36,13 +42,38 @@ class Configuration:
     0.01
     """
 
+    def __init__(self) -> None:
+        object.__setattr__(self, "_options_registry", dict())
```
schmidtjonathan, line 46: We have to use the "default" attribute write access here, since we override `__setattr__`.

schmidtjonathan: And for a reason I haven't quite figured out, yet, using `self.__dict__`, instead of a dictionary that is initialized as an additional class attribute (i.e., here, `self._options_registry`), does not work.
If somebody knows why, I would be very grateful to learn.

schmidtjonathan:
Thanks to @marvinpfoertner for writing a really nice sphinx plugin for automatically generating the docs! I just adapted the rest :)

src/probnum/_config.py
```diff
@@ -36,39 +42,58 @@ class Configuration:
...
+
+    def __hasattr__(self, key: str) -> bool:
+        return key in self._options_registry
+
+    def __getattr__(self, key):
+        if not self.__hasattr__(key):
```
schmidtjonathan, line 52: Another thing I would like to understand, but I don't:
Using `hasattr(self, key)` here does not work. My only explanation is that it depends on the order in which python registers the methods somehow? If somebody knows, I'd appreciate a brief explanation!

marvinpfoertner:
Okay, so @schmidtjonathan and I broke Python with this thing, but it works so nobody touch it anymore! 🤣

docs/source/_ext/probnum-config-options.py
```diff
@@ -0,0 +1,80 @@
+from docutils import nodes
+from docutils.statemachine import StringList
+from sphinx.util.docutils import SphinxDirective, switch_source_input
+
+import probnum as pn
+
+
+class ProbNumConfigOptions(SphinxDirective):
```
pnkraemer, line 8: I dont understand what this class does. It does not seem crucial that I do, but would it make sense either way to document the objective and behaviour, at least with a few one-liner docstrings in the methods?

docs/source/_ext/probnum-config-options.py
```diff
@@ -0,0 +1,80 @@
...
+    def run(self):
+        table = nodes.table(
+            "",
+            classes=[
+                "longtable",
+                "colwidths-auto",
+                # "colwidths-given",
```
pnkraemer, line 15: is this on purpose?

src/probnum/_config.py
```diff
@@ -36,39 +28,61 @@ class Configuration:
...
-                    "accessed."
-                )
+            if key not in self._options_registry:
+                raise AttributeError(Configuration._NON_REGISTERED_KEY_ERR_MSG % key)
 
-            old_entries[key] = getattr(self, key)
+            old_options[key] = self._options_registry[key].value
 
-            setattr(self, key, value)
+            self._options_registry[key].value = value
```
pnkraemer, line 76: so all the options are handled inside the `_options_registry`? Is this attribute introduced somewhere? This is no meaningful review comment, I just don't understand where it is defined (and documented?).

pnkraemer (commented):
Thanks for doing this! I have two stupid questions (plus a very minor one), so it would be great if you could answer those. Other than that, the PR should be good to go.

docs/source/_ext/probnum-config-options.py
JonathanWenger: +1 for a short docstring

JonathanWenger (approved):
Great job for putting in the extra effort here to make this more maintainable. One tiny point: Could you re-order both the sidebar and table under "API Reference" so that `probnum.config` follows the alphabetical order of the subpackages (i.e. is at the top)? If you feel this is out of scope for this PR, that's fine as well.

src/probnum/_config.py
marvinpfoertner: `_options_registry` is initialized in the constructor, i.e. via

```
object.__setattr__(self, "_options_registry", dict())
```

This monstrosity is necessary, since we overwrite `__setattr__`, so simply calling

```
self._options_registry = {}
```

sadly doesn't work

docs/source/_ext/probnum-config-options.py
marvinpfoertner: I can add a docstring tomorrow, but I'm not sure how useful this is, since the docstring might be without reading the Sphinx extension tutorial and having read  the tutorial it might be trivial

docs/source/_ext/probnum-config-options.py
marvinpfoertner: Yup. In `docutils.nodes`, basically nothing is documented and it took some digging to find out how to adjust column widths automatically. In case we want to switch to statically sized columns, I added the respective command in order to save us the pain in the future

docs/source/_ext/probnum-config-options.py
JonathanWenger: Even just linking to the tutorial could be useful information for future maintainers.

pnkraemer: Thanks! I agree that even only a link to the tutorial might be useful :)

docs/source/_ext/probnum-config-options.py
pnkraemer: Do you think it makes sense to put a quick comment into the source here that explains exactly this issue? If not, that is okay as well.

src/probnum/_config.py
pnkraemer: Okay I didnt see it (maybe because it didnt look like a standard init ;)). Thanks. I will just believe you that it is necessary :D

schmidtjonathan:
@pnkraemer , @JonathanWenger , thanks for your reviews!
I integrated your feedback, do you want to have another look before I merge?

docs/source/_ext/probnum-config-options.py
schmidtjonathan: In my opinion, the keywords are sufficiently expressive, and a comment is not necessary. But I can add one, if you prefer :)

pnkraemer: I think it will do the job without (especially if I would be the only one to whom this ocmment wasnt obvious :))

pnkraemer (approved):
Thanks guys!

JonathanWenger:
> @pnkraemer , @JonathanWenger , thanks for your reviews!
> 
> I integrated your feedback, do you want to have another look before I merge?

All good! Thanks!

PR #481, opened 2021-07-14
Convenience function for perturbation-based solvers

pnkraemer:
# In a Nutshell
This PR introduces the analogon of `probsolve_ivp` for perturbation-based solvers such as `PerturbedStepSolver`.

# Note
This bases on #477 (since #477 is almost through now). I will convert this to `ready to review` once this will fit.

@nathanaelbosch @NinaEffenberger I will request your reviews soon, if that is okay with you guys :) No need to comment anything for now.


tests/test_diffeq/test_perturbsolve_ivp.py
```diff
@@ -0,0 +1,72 @@
...
+    """Methods that are not in the list raise errors."""
+    f = ivp.f
+    t0, tmax = ivp.t0, ivp.tmax
+    y0 = ivp.y0
+
+    wrong_methods = ["DOP853", "Radau", "LSODA", "BDF"]
+    for wrong_method in wrong_methods:
+        with pytest.raises(ValueError):
+            diffeq.perturbsolve_ivp(f, t0, tmax, y0, rng=rng, method=wrong_method)
+
```
NinaEffenberger, line 61: Maybe you can add a similar test `test_wrong_perturbation_style_raises_error` as you define an error message for that case, too.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+              polynomial is used for the dense output [3]_. Can be applied in
+              the complex domain.
+            * `RK23`: Explicit Runge-Kutta method of order 3(2) [4]_. The error
+              is controlled assuming accuracy of the second-order method, but
+              steps are taken using the third-order accurate formula (local
+              extrapolation is done). A cubic Hermite polynomial is used for the
+              dense output. Can be applied in the complex domain.
+
+        Other integrators are not supported currently.
+    perturb
```
NinaEffenberger, line 69: missing ':' after perturb

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+        Which perturbation style to use.
+        Currently, the following methods are supported:
+
+            * `step-lognormal`: Perturbed-step (aka random time-step numerical integration) method
+              with lognormally distributed perturbation of the step-size [1]_.
+            * `step-uniform`: Perturbed-step (aka random time-step numerical integration) method
+              with lognormally distributed perturbation of the step-size [1]_.
```
NinaEffenberger, line 76: copy-paste error: uniform instead of lognormally in line 76

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+
+        Other integrators are not supported currently.
+    noise_scale
```
NinaEffenberger, line 79: missing ':'

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+        Scale of the perturbation. Optional. Default is 10.0. The magnitude of this parameter
+        significantly impacts the width of the posterior.
```
NinaEffenberger, line 81: not sure whether 'optional' is the correct word in this context, for the method this parameter is not optional

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+    adaptive :
+        Whether to use adaptive steps or not. Default is `True`.
+    atol :
+        Absolute tolerance  of the adaptive step-size selection scheme.
+        Optional. Default is ``1e-4``.
```
NinaEffenberger, line 86: default is 1e-2

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+    rtol :
+        Relative tolerance   of the adaptive step-size selection scheme.
+        Optional. Default is ``1e-4``.
```
NinaEffenberger, line 89: see line 86

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+        Step size. If atol and rtol are not specified, this step-size is used for a fixed-step ODE solver.
+        If they are specified, this only affects the first step. Optional.
+        Default is None, in which case the first step is chosen as prescribed by :meth:`propose_firststep`.
+
+    Examples
+    --------
+    >>> from probnum.diffeq import perturbsolve_ivp
+    >>> import numpy as np
+
+    Solve a simple logistic ODE with fixed steps.
```
NinaEffenberger, line 100: Maybe you can clarify once again that a perturbed solver with uniformly perturbed steps is used. The default can't be seen in the code and I like the idea of providing a default perturbation function, but I think this should be made clear in this example.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+    >>> print(np.round(solution.states.mean, 3))
+    [[0.15 ]
+     [0.292]
+     [0.497]
+     [0.766]
+     [0.874]
+     [0.955]
+     [0.987]]
+
+    Each solution is the result of a randomly-perturbed call the an underlying Runge-Kutta solver.
```
NinaEffenberger, line 119: Each solution is the result of a randomly-perturbed call **of** an underlying Runge-Kutta solver.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+     [0.448]
+     [0.657]
+     [0.853]
+     [0.945]
+     [0.975]]
+
+    Other methods, such as `RK45` (as well as other perturbation styles) are easily accessible.
+    Let us solve the same equation, with an adaptive RK45 solver that uses lognormally perturbed steps.
+
+    >>> solution = perturbsolve_ivp(f, t0, tmax, y0, rng=rng, atol=1e-5, rtol=1e-6, method="RK45", adaptive=True)
```
NinaEffenberger, line 135: `` solution = perturbsolve_ivp(f, t0, tmax, y0, rng=rng, atol=1e-5, rtol=1e-6, method="RK45", perturb="step-lognormal", adaptive=True)``

NinaEffenberger: (you forgot to set the perturbation function :))

NinaEffenberger (changes requested):

src/probnum/diffeq/_probsolve_ivp.py
```diff
@@ -174,11 +174,11 @@ def probsolve_ivp(
     Examples
     --------
     >>> from probnum.diffeq import probsolve_ivp
-    >>> from probnum import randvars
     >>> import numpy as np
 
     Solve a simple logistic ODE with fixed steps.
 
+    >>>
```
nathanaelbosch, line 181: Was this empty line intentional? It looks ok in the docs, but I was just wondering.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+              with lognormally distributed perturbation of the step-size [1]_.
+            * `step-uniform`: Perturbed-step (aka random time-step numerical integration) method
+              with lognormally distributed perturbation of the step-size [1]_.
+
+        Other integrators are not supported currently.
+    noise_scale
+        Scale of the perturbation. Optional. Default is 10.0. The magnitude of this parameter
+        significantly impacts the width of the posterior.
+    adaptive :
+        Whether to use adaptive steps or not. Default is `True`.
```
nathanaelbosch, line 83: Interesting! I somehow thought that step-size adaptation is kind of an unsolved thing with these methods? Or was that only for the Conrad method?

src/probnum/diffeq/_perturbsolve_ivp.py
nathanaelbosch: Aren't 1e-2 tolerances too large for defaults? SciPy (and DifferentialEquations.jl too btw :wink:) uses 1e-3 for rtol and 1e-6 for atol.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+        Journal of Computational and Applied Mathematics, Vol. 6, No. 1, pp. 19-26, 1980.
+    .. [3] L. W. Shampine.
+        Some Practical Runge-Kutta Formulas.
+        Mathematics of Computation, Vol. 46, No. 173, pp. 135-150, 1986.
+    .. [4] P. Bogacki, L.F. Shampine.
+        A 3(2) Pair of Runge-Kutta Formulas.
+        Appl. Math. Lett. Vol. 2, No. 4. pp. 321-325, 1989.
+    """
+
+    if method not in METHODS:
```
nathanaelbosch, line 166: I'm not a big fan of the `SELECT_` variable names, and I think I would rather remove them in favor of `METHODS` and `PERTURBATIONS` that just contain the methods and perturbations. 
This line could then be changed to `if method not in METHODS.keys()`, or even prettier (but more implicit) `if method not in METHODS` (since it's the same).

But this is a pretty minor thing and I don't have a strong opinion against the current naming.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+        raise ValueError("Method is not supported.")
```
nathanaelbosch, line 167: Maybe adding the supplied `method`, and possibly even the list of available `METHODS`, could be helpful.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+
+    if perturb not in PERTURBS:
+        raise ValueError("Perturbation-style is not supported.")
```
nathanaelbosch, line 170: See comment above.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -0,0 +1,191 @@
...
+
+    ivp = problems.InitialValueProblem(t0=t0, tmax=tmax, y0=np.asarray(y0), f=f)
+    scipy_solver = SELECT_METHOD[method](ivp.f, ivp.t0, ivp.y0, ivp.tmax)
```
nathanaelbosch, line 173: Is there a benefit to constructing the `ivp`, just to deconstruct it again? Maybe that line could be removed.

nathanaelbosch (changes requested):
I had some very small comments, but overall this looks good :+1: Just re-request whenever you need the approval. Thanks for the PR!

src/probnum/diffeq/_perturbsolve_ivp.py
pnkraemer: I dont know where you got that from... it works :) Maybe it is not implemented elsewhere (I guess you are referring to whatever DifferentialEquations.jl does for UQ), because it is really hard to batch, or do dense output or whatever. I dont know.

src/probnum/diffeq/_perturbsolve_ivp.py
pnkraemer: Dunno why it is 1e-2. I meant 1e-4. I will change the default, not the doc :)

src/probnum/diffeq/_perturbsolve_ivp.py
pnkraemer: So you mean `SELECT_METHOD` -> `METHODS`? I think that is a fair solution.

src/probnum/diffeq/_perturbsolve_ivp.py
pnkraemer: This is because the scipy solver need (for some reason) the whole IVP at construction time. 
In #486 this will be resolved entirely.

src/probnum/diffeq/_perturbsolve_ivp.py
nathanaelbosch: DifferentialEquations.jl actually has an adaptive-step version for the Conrad solver that basically selects the diffusion from the local error estimate and some convergence rate, but that does not have a proper reference I think. Maybe I mixed the "how to calibrate" and "how to select step sizes" part, so I guess I have to check out the paper again to see if/how they do things.

src/probnum/diffeq/_perturbsolve_ivp.py
nathanaelbosch: Yes that's what I meant, together with removing the current `METHODS` and `PERTURBS` variables.

src/probnum/diffeq/_perturbsolve_ivp.py
nathanaelbosch: I don't see where the scipy solver has access to the ivp object? But I found another answer to my question: In line 186, `stepsize.propose_firststep` requires access to an `ivp` object, so that's a reason for building it.

pnkraemer: Ah, I misunderstood your question then... Glad it is resolved anyway :)

src/probnum/diffeq/_perturbsolve_ivp.py
pnkraemer: This refers to the fact that there is a default value, so it is optional for the user to pass it (which is in line with the other docs in diffeq)

src/probnum/diffeq/_probsolve_ivp.py
pnkraemer: yup

NinaEffenberger (approved):

nathanaelbosch (approved):

PR #484, opened 2021-07-14
Further bibliographic improvements

christopheroates:
# In a Nutshell
Added recent PN references in Quadrature, ODEs, PDEs and Linear Algebra, removed duplicate references, removed references to work on "numerical GPs" (as this is not PN).


docs/source/research/bibliography/ODEs.bib
```diff
@@ -328,14 +328,13 @@ @incollection{teymur18
   file = {https://arxiv.org/pdf/1805.07970.pdf}
 }
 
-@article{breen2019bulirsch,
```
philipphennig, line 331: I think this is indeed ok to remove. The authors originally contacted us to have this added, but now that we're slimming the list, I agree that it's not a crucial thing to have around.

philipphennig (commented):
@christopheroates , you're on fire! Thanks for all these edits; especially for updating outdated pre-print references with published versions. 

Can I make just one request? As per Mike Osborne's request, could you add a new section on `BayesianOptimization.bib` (separate from optimization, as they are frequently confused), and add my Entropy Search paper (which you just moved to `Optimization.bib`) there? Then I can add two or three additional papers there. 

@mmahsereci , I checked out the commit, built it (can confirm no errors), and looked at the built page. All works fine on the code side.

docs/source/research/bibliography/ODEs.bib
christopheroates: Ah, I actually removed this because it's been superseded by https://arxiv.org/abs/2106.13718 , which I've added to "General" :-)

christopheroates:
Thanks Philipp - yes, I've now added a separate section on Bayesian Optimization (I've just added Mokus et al, so please feel free to add other historical works!)

I'm actually not sure I appreciate the distinction being drawn between optimisation and Bayesian optimisation - I appreciate that the methods are qualitatively different between the two sections, but the tasks being solved are the same, aren't they?  In any case, I'm happy to go along with this for the moment :-)

mmahsereci:
Hi @christopheroates , the auto-build of the webpage failed. It seems the Bayesian optimization bibfile is missing the ending `.bib`. 

If I may put in my two cents on the other discussion, I believe the papers under "optimization" aim to optimize very a high-dimensional, perhaps stochastic functions locally/greedily, and no global minimum is expected as solution, while the ones under currently under "Bayesian optimization" expect the solution to be the global minimizer. But we also did not seem to have called the "quadrature" section "Bayesian Quadrature", so one way of distinguishing the two sections would be to call them "Local optimization" and "Global optimization".

christopheroates:
Thanks Maren, I've fixed the .bib error (sorry about that) and I like your local vs global optimisation distinction, so I've implemented that.

docs/source/research/literature.md
```diff
@@ -33,37 +31,25 @@ all:
 ---
 ```
 
-## Optimization
+## Global Optimization
 
-```{bibliography} bibliography/Optimization.bib
+```{bibliography} bibliography/Bayesian_Optimization.bib
```
mmahsereci, line 36:
```suggestion
```{bibliography} bibliography/Global_Optimization.bib
```

docs/source/research/literature.md
```diff
@@ -33,37 +31,25 @@ all:
...
-reverse problem, in some sense, has
-also found interest in machine learning: inferring a differential equation from
-(noisy) observations of trajectories that are assumed to be governed by this
-ODE. Below, this is listed under "inferring ODEs".
-
-### Solving ODEs
+## Local Optimization
 
-```{bibliography} bibliography/ODEs.bib
+```{bibliography} bibliography/Optimization.bib
```
mmahsereci, line 44:
```suggestion
```{bibliography} bibliography/Local_Optimization.bib
```

mmahsereci (commented):
No problem at all @christopheroates , this is all part of the review process to get the code to compile. I added 2 more minor comments to match the naming of the new bib files in the `literature.md` file. I also noticed that some files went unused with the changes; these are `ODE_from_path.bib` and `related.bib`. I just want to make sure this was on purpose. 

@philipphennig to comment on the content of the changes.

christopheroates:
Thanks Maren, yes - the removing of ODE_from_path.bib and related.bib was intentional, to limit the focus to PN.  (The first category was about parameter inference for ODEs, which is not PN, and the second category was too vague to be useful, in my opinion, because lots of things are "related" to PN!)

docs/source/research/literature.md
mmahsereci: @christopheroates you can either use the "Commit suggestion" button above to merge the suggested change into your fork, or edit `literature.md` manually in your fork. The proposed change simply fixes the link in the `literature.md` file to the correct bib file such the website compiles without errors. Same for the review comment below :)

docs/source/research/bibliography/ODEs.bib
philipphennig: Even better! Thanks!

docs/source/research/literature.md
philipphennig: *Global* v. *Local* Optimization is fine for me, although of course the term **Bayesian Optimization** certainly has a life of its own now. So maybe it should be mentioned in the heading of that section. 

@christopheroates : For me the key difference between BO and "Optimization" is that BO collects "external" data, which is presumed to be near-arbitrarily expensive. So it's all about sample efficiency. And in fact, Bayesian Optimizers invariably call a numerical optimization routine as an inner loop. For those numerical optimizers, computational cost is a key concern. Part of the way they deal with this is that they usually consider local optima only. 

In the deep learning community (where BO is quite popular), BayesOpt is seen as the tool to optimize hyperparameters, the outer loop, while *optimizers* are used to do the actual, high-dimensional, stochastic training. 

This hopefully also explains why I want to keep these sections separate: To avoid people getting the wrong impression that "probabilistic optimizers are Bayesian optimizers, and Bayesian optimizers are super-expensive. Thus probabilistic optimizers are not relevant for deep learning".

philipphennig:
Thanks @mmahsereci and @christopheroates!

@mmahsereci , I am fine with the content changes @christopheroates suggests. But as requested by Chris, I will add 2-3 additional citations for the BO section. So let's leave this PR open until I've sorted that out (hopefully tonight).

mmahsereci:
@philipphennig let's merge this PR first. Then I'd suggest you open your own PR with the changes you intend to do (adding 2-3 more papers) which then @christopheroates can review. Otherwise you would need to edit Chris' fork which, apart from small fixes is sub-optimal.

christopheroates:
Thanks Philipp and Maren, this all sounds sensible!  I'm afraid that I a bit lost with the discussion about forks and pull requests - could you please drop me an email if there is something that I need to do about this?  Sorry if I am causing a bottleneck for everyone!!

mmahsereci:
These contributions are fantastic @christopheroates ! Merging now..

PR #490, opened 2021-07-17
Initialization routines for ODE filters

pnkraemer:
# In a Nutshell
This PR introduces an interface for initialization routines in ODE filters.

# Detailed Description
See above. This PR also cleans up the tests for initialization routines and updates some docs. Taylor-mode AD code is cleaned up slightly.
The concept is very similar to what happens in #486, but since the content is sufficiently orthogonal, I made a new PR.


src/probnum/diffeq/_probsolve_ivp.py
```diff
@@ -275,11 +275,14 @@ def probsolve_ivp(
...
-        prior_process,
-        measmod,
+
+    rk_init = odefiltsmooth.initialize.RungeKuttaInitialization()
+    solver = odefiltsmooth.GaussianIVPFilter(
+        ivp=ivp,
+        prior_process=prior_process,
+        measurement_model=measmod,
         with_smoothing=dense_output,
+        initialization_routine=rk_init,
```
pnkraemer, line 285: In my opinion, passing the initialisation routine with these objects is so easy that the alternative constructors are rather in the way than helpful.

src/probnum/diffeq/odefiltsmooth/initialize/_initialize.py
```diff
@@ -0,0 +1,45 @@
...
+
+        Some initialization routines yield the exact initial
+        derivatives, some others only yield approximations.
+        """
+        return self._is_exact
+
+    @property
+    def requires_jax(self) -> bool:
+        """Whether the implementation of the routine relies on JAX."""
+        return self._requires_jax
```
pnkraemer, line 45: I see the potential point of criticism that these properties are not super useful.
I opted for introducing them anyway for communication reasons: in order to document more clearly in which sense the routines may not be interchangeable (`is_exact`), and in which case they have unusual requirements (`requires_jax`).

src/probnum/diffeq/odefiltsmooth/initialize/_taylor_mode.py
```diff
@@ -0,0 +1,184 @@
...
+    >>> print(improved_initrv.std)
+    [0. 0. 0. 0. 0. 0. 0. 0.]
+    """
+
+    def __init__(self):
+        super().__init__(is_exact=True, requires_jax=True)
+
+    def __call__(
+        self, ivp: problems.InitialValueProblem, prior_process: randprocs.MarkovProcess
+    ) -> randvars.RandomVariable:
```
pnkraemer, line 98: please let me know if the changed naming (and some reordering) makes it easier to understand what this code does :)

src/probnum/diffeq/odefiltsmooth/initialize/_initialize.py
```diff
@@ -0,0 +1,45 @@
+"""Interface for ODE filter initialization."""
+
+import abc
+
+from probnum import problems, randprocs, randvars
+
+
+class InitializationRoutine(abc.ABC):
```
pnkraemer, line 8: I opted for making this a class for multiple reasons:
1) the routine should only have the interface (ivp, prior) -> rv. Runge-Kutta initialisation has some additional parameters, which should not be part of the call signature IMO
2) For different ODEs, these routines might offer multiple options
3) it is easier to document some properties, e.g. whether the initialization is exact or not
4) it is a MUCH easier type in the constructor of GaussianIVPFilter

nathanaelbosch (changes requested):
The PR looks good, I just found a single typo and otherwise I have nothing to complain. Thanks a lot!

nathanaelbosch (approved):

PR #498, opened 2021-07-21
Event handling for ODE solvers

### LINKED ISSUE #313: Start stop vs t

pnkraemer:
# In a Nutshell
This PR introduces event handling functionality to PN ode solvers (on a very basic level). I recommend looking at  the newly created notebook first in order to get a feeling for the functionality and API.

# Detailed Description
This PR implements 
* a callback interface and discrete callbacks for ODE solvers.
* ODESolverState(t, rv, error_estimate, ...) which is now the object of interest to be passed around by ODE solvers. The advantage is that due to the tight coupling (t, rv, error) the code is less error-prone (I think I fixed a few errors on the way). This is also the only way of giving the modification functions in the callback a sensible interface

# Related Issues
Closes #313, because now the signature of a perform_step() function is (state, dt).


src/probnum/diffeq/_odesolver.py
```diff
@@ -79,11 +165,32 @@ def postprocess(self, odesol):
...
 
-    def method_callback(self, time, current_guess, current_error):
+    def method_callback(self, state):
         """Optional callback.
 
         Can be overwritten. Do this as soon as it is clear that the
         current guess is accepted, but before storing it. No return. For
         example: tune hyperparameters (sigma).
         """
         pass
```
pnkraemer, line 175: currently, this is still part of the ODE solver. I think it is superfluous, but since it is used for the wrapped scipy solvers, I think it cannot be removed safely.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -28,12 +28,13 @@ def perturbsolve_ivp(
     y0,
     rng,
     method="RK45",
-    perturb="step-uniform",
+    perturb="step-lognormal",
```
pnkraemer, line 31: this is a better default, because uniform perturbation only works for steps in (0, 1). lognormal perturbation always works.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -28,12 +28,13 @@ def perturbsolve_ivp(
...
     noise_scale=10.0,
     adaptive=True,
     atol=1e-6,
     rtol=1e-3,
     step=None,
+    time_stamps=None,
```
pnkraemer, line 37: I am not 100% sure about the name. If you have a better idea, let me know.

src/probnum/diffeq/callbacks/_callback.py
```diff
@@ -0,0 +1,29 @@
+"""Event handler interface."""
+
+import abc
+from typing import Callable, Union
+
+import probnum  # pylint: disable="unused-import"
```
pnkraemer, line 6: this is here for the type only...

src/probnum/diffeq/callbacks/_discrete_callback.py
```diff
@@ -0,0 +1,33 @@
...
+    A discrete event can be a time-stamp that must be included in the locations. It can
+    also be any event for which it is possible to write down a condition that evaluates
+    to `True` or `False`. If a condition evaluates to `True`, the current state can be
+    modified.
+    """
+
+    # New init because condition() types are more specific.
+    def __init__(
+        self,
+        replace: Callable[
```
pnkraemer, line 21: I went for `replace` (previously: `modify`), because (i) this is less suggestive of in-place operations, and (ii) the most convenient way of implementing this function may be with `dataclasses.replace(state, rv=something)`.

src/probnum/diffeq/odefiltsmooth/_ivpfiltsmooth.py
```diff
@@ -90,14 +94,16 @@ def __init__(
         # or from any other state.
         self._reference_coordinates = _reference_coordinates
 
-    def initialise(self):
+    def initialize(self):
```
pnkraemer, line 97: long overdue british -> american :(

src/probnum/diffeq/odefiltsmooth/_ivpfiltsmooth.py
```diff
@@ -148,27 +154,26 @@ def step(self, t, t_new, current_rv):
...
         )
 
         # Compute the measurements for the error-free component
         pred_rv_error_free, _ = self.prior_process.transition.forward_realization(
-            error_free_state, t=t, dt=dt
+            error_free_state, t=state.t, dt=dt
         )
         meas_rv_error_free, _ = self.measurement_model.forward_rv(
-            pred_rv_error_free, t=t
+            pred_rv_error_free, t=state.t + dt
```
pnkraemer, line 174: this is a bugfix, right?

src/probnum/diffeq/odefiltsmooth/_ivpfiltsmooth.py
```diff
@@ -148,27 +154,26 @@ def step(self, t, t_new, current_rv):
...
         )
-        H = self.measurement_model.linearized_model.state_trans_mat_fun(t=t)
+        H = self.measurement_model.linearized_model.state_trans_mat_fun(t=state.t + dt)
```
pnkraemer, line 176: same here?

src/probnum/diffeq/odefiltsmooth/_ivpfiltsmooth.py
```diff
@@ -188,7 +193,9 @@ def step(self, t, t_new, current_rv):
 
         # Estimate local diffusion_model and error
         local_diffusion = self.diffusion_model.estimate_locally(
-            meas_rv=meas_rv, meas_rv_assuming_zero_previous_cov=meas_rv_error_free, t=t
+            meas_rv=meas_rv,
+            meas_rv_assuming_zero_previous_cov=meas_rv_error_free,
+            t=state.t + dt,
```
pnkraemer, line 198: same here?

tests/test_diffeq/test_perturbed/test_scipy_wrapper/test_wrapped_scipy_odesolution.py
```diff
@@ -10,7 +10,7 @@
 )
 def solution_case(testsolver, scipysolver):
     testsolution = testsolver.solve(
-        diffeq.stepsize.AdaptiveSteps(0.1, atol=1e-12, rtol=1e-12)
+        diffeq.stepsize.AdaptiveSteps(0.1, atol=1e-4, rtol=1e-4)
```
pnkraemer, line 13: I found 1e-12 a little bit too strict here ;)

pnkraemer:
@nathanaelbosch this thing is ready for review now. I will assign you already, but there is no time-pressure (I suppose this and the information operators PR will have to be merged one after the other anyway).

src/probnum/diffeq/odefiltsmooth/_ivpfiltsmooth.py
```diff
@@ -231,12 +233,27 @@ def step(self, t, t_new, current_rv):
...
+                error_estimate=local_errors,
+                reference_state=reference_values,
+            )
+
+            return state
 
         # Now we can be certain that the step will be accepted. In this case
         # we update the diffusion model and continue the rest of the iteration.,
-        self.diffusion_model.update_in_place(local_diffusion, t=t)
+        self.diffusion_model.update_in_place(local_diffusion, t=state.t)
```
nathanaelbosch, line 256: Should this here be t, or t_new?

src/probnum/diffeq/_odesolver.py
```diff
@@ -1,66 +1,145 @@
...
+            if time_stopper is not None:
+                dt = time_stopper.adjust_dt_to_time_stamps(state.t, dt)
+
+            state, dt = self.perform_full_step(state, dt)
+
+            if callbacks is not None:
+                for callback in callbacks:
+                    state = callback(state)
+
+            self.num_steps += 1
```
nathanaelbosch, line 80: Unrelated to the PR, but don't we rather want to count the step attempts for benchmarks? This quantity should also be easy to get, by calling `len` on the solution.

nathanaelbosch (changes requested):
Looks good! I have some small comments here and there but I think we should be able to resolve this quickly.

src/probnum/diffeq/_odesolver.py
pnkraemer: Maybe. I am not sure what this quantity is used for at the moment.  The whole "ODE solver statistics" area is rather unexplored at the moment. We can change `self.num_steps` to a variable that counts the attempted steps, but in this case, we should rename the variable. Do you think I should do this here?

src/probnum/diffeq/odefiltsmooth/_ivpfiltsmooth.py
pnkraemer: I think it should be `t`. But I am not 100% sure (which is reinforcement that the diffusion model implementation is not at its final stage yet - it should be clear!).

src/probnum/diffeq/_odesolver.py
nathanaelbosch: I think it's fine to leave it out of this PR, it was more of a general comment to get your opinion about this quantity and quantities to track in general :)

nathanaelbosch (approved):

src/probnum/diffeq/_odesolver.py
pnkraemer: Alright. Now that I know more about what you were aiming for here: I would like to track many more quantities than that, but when and how this wish is executed is not clear yet. There are more urgent todos :)

PR #501, opened 2021-07-27
Improve SDE API 

pnkraemer:
# In a Nutshell
This PR improves the API (mostly variable naming, some more small things) and adds a duplicate() method to transitions.

# Detailed Description

* The PR splits the sde and the discrete transition objects into multiple files (which is where the large file-diff stems from).
* The idea behind the duplicate() method is that on the way to a much simpler diffusion model handling, we need to start making lots of copies of transition objects (eventually, one transition per grid point). To this end, it must be easy to (a) duplicate transitions, and (b) make slight parameter changes (which eventually will be used as `prior.transition.duplicate(diffusion=calibrated_diffusion)`).
* Discrete Gaussian transitions are slightly renamed now (since the `discrete` is part of the module path).

# Other
I intend to do the same variable renaming / api changes to discrete transitions, but probably in a future PR.
I also intend to introduce a method `duplicate_with_changed_coordinates(linop)` soon, which should make it easier to work with preconditioners / coordinate changes in state-space models.


src/probnum/filtsmooth/optim/_iterated_component.py
```diff
@@ -164,3 +164,22 @@ def __getattr__(self, attr):
...
+            except KeyError:
+                return getattr(self, key)
+
+        component = replace_key("component")
+        stopcrit = replace_key("stopcrit")
+
+        return IteratedDiscreteComponent(
+            component=component,
+            stopcrit=stopcrit,
+        )
```
pnkraemer, line 185: Some thoughts:
1) This function replicates the `dataclasse.replace` functionality (this is where the `**changes` notation stems from)
2) Maybe have a look at `markov.Transition.duplicate()` for more info
3) I am not suuuper happy with the implementation, but it does the job reliably.

src/probnum/randprocs/markov/_transition.py
```diff
@@ -425,6 +425,20 @@ def jointly_transform_base_measure_realization_list_forward(
...
+
+        # Safeguard against inheritance of the duplication functionality
+        # which is no actual duplication hence we pretend it is not implemented.
+        if type(duplicated) is not type(self):
+            raise NotImplementedError("duplication is not implemented.")
+        return duplicated
+
+    @abc.abstractmethod
+    def _duplicate(self, **changes):
+        raise NotImplementedError
```
pnkraemer, line 440: the issue here is that a `duplicate()` method should return an object of the same type, so it needs to be implemented in *each* subclass (in particular, we want to avoid inheritance). This solution with an abstract method forces that at least some duplication method can be found, however, it can only be *used* if it is implemented in this particular class (all transitions implement one).

src/probnum/randprocs/markov/continuous/_lti_sde.py
```diff
@@ -0,0 +1,223 @@
...
+    def dispersion_matrix(self):
+        return self._dispersion_matrix
+
+    @property
+    def forward_implementation(self):
+        return self._forward_implementation_string
+
+    @property
+    def backward_implementation(self):
+        return self._backward_implementation_string
```
pnkraemer, line 90: before, it was possible to overwrite `self.drift_matrix` with some potentially unexpected behaviour (i.e. the wrappers that initialise superclasses as in `nonlinear(t, x) = linear(t) @ x` might not have been changed. This is impossible now, and the only way of changing the object is via `self.duplicate(...)`.

src/probnum/filtsmooth/gaussian/approx/_extendedkalman.py
```diff
@@ -157,30 +158,35 @@ def __init__(
...
-        def forcevecfun(t):
+        def force_vector_function(t):
             return g(t, x0) - dg(t, x0) @ x0
 
-        def driftmatfun(t):
+        def drift_matrix_function(t):
             return dg(t, x0)
 
+        def dispersion_matrix_function(t):
+            return l(t, x0)
```
pnkraemer, line 174: this is what the continuous EKF does (we did this in time series).

src/probnum/randprocs/markov/continuous/_sde.py
```diff
@@ -1,38 +1,58 @@
...
     def __init__(
         self,
-        dimension: IntArgType,
-        driftfun: Callable[[FloatArgType, np.ndarray], np.ndarray],
-        dispmatfun: Callable[[FloatArgType, np.ndarray], np.ndarray],
-        jacobfun: Callable[[FloatArgType, np.ndarray], np.ndarray],
+        state_dimension: IntArgType,
+        wiener_process_dimension: IntArgType,
+        drift_function: Callable[[FloatArgType, np.ndarray], np.ndarray],
+        dispersion_function: Callable[[FloatArgType, np.ndarray], np.ndarray],
```
pnkraemer, line 24: dispersion functions are now (t, x) dependent. This is a lack of generality that was bugging me before.

src/probnum/randprocs/markov/_transition.py
schmidtjonathan: I see the problem.

Wouldn't it be sufficient to make `duplicate` an `@abc.abstractmethod`? Then everything that subclasses `Transition` has to implement it. Or am I missing something?

src/probnum/randprocs/markov/continuous/_linear_sde.py
```diff
@@ -0,0 +1,477 @@
+"""Linear SDE models as transitions."""
```
schmidtjonathan, line 1: I guess this is all just moved from the former `sde.py`, right?

src/probnum/randprocs/markov/continuous/_lti_sde.py
schmidtjonathan: Yep, I like the introduction of `@property`s. We could also consider defining a `setter` and throwing a respective warning that the user is then up to him/herself. :D But I wouldn't.
 
Super minor: I just wonder whether we should rename the `forward`-/`backward_implementation` to `..._string` as well. The name otherwise suggests some kind of functionality (resp. some kind of `implementation`).

src/probnum/randprocs/markov/continuous/_lti_sde.py
```diff
@@ -0,0 +1,223 @@
...
+        force_vector = replace_key("force_vector")
+        forward_implementation = replace_key("forward_implementation")
+        backward_implementation = replace_key("backward_implementation")
+
+        return LTISDE(
+            drift_matrix=drift_matrix,
+            dispersion_matrix=dispersion_matrix,
+            force_vector=force_vector,
+            forward_implementation=forward_implementation,
+            backward_implementation=backward_implementation,
```
schmidtjonathan, line 201: I wish we could avoid writing the same code over and over again, for every subclass of `Transition`. But from the top of my head, I cannot think of a better solution.

schmidtjonathan: Maybe something along the lines of
```python
import copy

# ...

def _duplicate(self, **changes):
  lti_kwargs = copy.deepcopy(self.__dict__)
  lti_kwargs.update(changes)
  return LTISDE(**lti_kwargs)
```

If in `changes` there is an invalid key, then I think there will be this `received unexpected argument` error.

src/probnum/randprocs/markov/_transition.py
```diff
@@ -425,6 +425,20 @@ def jointly_transform_base_measure_realization_list_forward(
             out_samples.append(curr_sample)
         return out_samples
 
+    def duplicate(self, **changes):
```
schmidtjonathan, line 428:
```suggestion
    def duplicate(self, /, **changes):
```

src/probnum/randprocs/markov/_transition.py
```diff
@@ -425,6 +425,20 @@ def jointly_transform_base_measure_realization_list_forward(
...
+        duplicated = self._duplicate(**changes)
+
+        # Safeguard against inheritance of the duplication functionality
+        # which is no actual duplication hence we pretend it is not implemented.
+        if type(duplicated) is not type(self):
+            raise NotImplementedError("duplication is not implemented.")
+        return duplicated
+
+    @abc.abstractmethod
+    def _duplicate(self, **changes):
```
schmidtjonathan, line 439:
```suggestion
    def _duplicate(self, /, **changes):
```

schmidtjonathan: I suggest requiring keyword-only here

src/probnum/randprocs/markov/_transition.py
pnkraemer: Unfortunately not, because then e.g. `LinearSDE` would be duplicated into an `SDE` if not implemented.

src/probnum/randprocs/markov/continuous/_lti_sde.py
pnkraemer: Yes, a setter could work, though I agree with you that this would be a bit pointless. 

The string: yes, that seems like a good idea to me.

src/probnum/filtsmooth/optim/_iterated_component.py
schmidtjonathan: I left some comments in the other classes (below).

src/probnum/randprocs/markov/continuous/_mfd.py
```diff
@@ -5,17 +5,17 @@
 import scipy.linalg
 
 
-def matrix_fraction_decomposition(driftmat, dispmat, dt):
+def matrix_fraction_decomposition(drift_matrix, dispersion_matrix, dt):
```
schmidtjonathan, line 8: 👍 Thanks!

src/probnum/randprocs/markov/continuous/_lti_sde.py
pnkraemer: I agree with you. Maybe there is a better solution. Unfortunately, your suggestion above doesnt work (because the elements in __dict__ are not directly related to the signature of __init__). Like you, I thought about copy.deepcopy, but wasnt able to make it work the way I wanted.

I will try to do some brainstorming for a more elegant solution. Off the top of my head, perhaps some dispatchy thing? Or maybe even make the whole thing a dataclass? Dunno...

src/probnum/randprocs/markov/integrator/_iwp.py
```diff
@@ -127,38 +127,69 @@ def __init__(
...
+    #
+    # @property
+    # def dispersion_matrix(self):
+    #     return self._dispersion_matrix
+    #
+    # @dispersion_matrix.setter
+    # def dispersion_matrix(self, dispersion_matrix):
+    #     raise TypeError(
+    #         f"{self.__class__} does not support assignment of dispersion_matrix."
+    #     )
```
schmidtjonathan, line 192: Do we want to keep this comment?

PR #505, opened 2021-07-28
LinOp arithmetic part 1

schmidtjonathan:
# In a Nutshell
Add (efficient) arithmetics for combinations of Linear Operators : Part 1

# Detailed Description
For some combinations of linear operators, there exist known arithmetics that can be implemented more efficiently.
E.g. `Kronecker @ Kronecker ~~> Kronecker` instead of yielding a `ProductLinearOperator` that builds up large call stacks.

# Related Issues
#509 ; mitigates the described problem in parts. (Subsequent PRs will consecutively solve the issue)


src/probnum/linops/_arithmetic.py
```diff
@@ -57,6 +94,295 @@ def matmul(op1: LinearOperator, op2: LinearOperator) -> LinearOperator:
...
+
+# Scaling
+def _mul_scalar_scaling(scalar: ScalarArgType, scaling: Scaling) -> Scaling:
+    if scaling.is_isotropic:
+        return Scaling(scalar * scaling.scalar, shape=scaling.shape)
+
+    return Scaling(scalar * scaling.factors, shape=scaling.shape)
+
+
+def _mul_scaling_scalar(scaling: Scaling, scalar: ScalarArgType) -> Scaling:
```
marvinpfoertner, line 109: This sort of 'reversed' method can be avoided by using a utility method similar to https://github.com/probabilistic-numerics/probnum/blob/2dadf9787357c5e3a54ed07fa466427fefc54185/src/probnum/randvars/_arithmetic.py#L103

src/probnum/linops/_arithmetic.py
```diff
@@ -57,6 +94,295 @@ def matmul(op1: LinearOperator, op2: LinearOperator) -> LinearOperator:
...
+    _matmul_fns[(op_type, ScaledLinearOperator)] = _matmul_op_scaled
+    _matmul_fns[(NegatedLinearOperator, op_type)] = _matmul_scaled_op
+    _matmul_fns[(op_type, NegatedLinearOperator)] = _matmul_op_scaled
+
+
+# Kronecker
+
+
+def _matmul_scaling_kronecker(scaling: Scaling, kronecker: Kronecker) -> Kronecker:
+    if scaling.shape[1] != kronecker.shape[0]:
```
marvinpfoertner, line 143: Can we maybe put this check in a _matmul_apply method? I imagine we will need it in any matmul routine

src/probnum/linops/_arithmetic.py
```diff
@@ -57,6 +94,295 @@ def matmul(op1: LinearOperator, op2: LinearOperator) -> LinearOperator:
...
+            f"matmul received invalid shapes {kronecker.shape} @ {scaling.shape}"
+        )
+
+    if scaling.is_isotropic:
+        return kronecker * scaling.scalar
+    return NotImplemented
+
+
+def _mul_scalar_kronecker(scalar: ScalarArgType, kronecker: Kronecker) -> Kronecker:
+    sqrt_scalar = np.sqrt(scalar)
```
marvinpfoertner, line 165: This does not work for negative scalars, we will need some sort of workaround here :/

src/probnum/linops/_arithmetic.py
```diff
@@ -65,11 +391,21 @@ def _apply(
...
         Callable[
             [LinearOperator, LinearOperator],
-            Union[LinearOperator, type(NotImplemented)],
+            Union[LinearOperator, NotImplementedType],
         ]
     ] = None,
-) -> Union[LinearOperator, type(NotImplemented)]:
-    key = (type(op1), type(op2))
+) -> Union[LinearOperator, NotImplementedType]:
+    if np.isscalar(op1):
```
marvinpfoertner, line 398: According to numpy docs, the better check here is `np.ndim(op1) == 0`.

src/probnum/linops/_arithmetic.py
```diff
@@ -65,11 +391,21 @@ def _apply(
...
+        key1 = "scalar"
```
marvinpfoertner, line 399: We should also normalize types to `numpy.number` here (`probnum.utils.as_numpy_scalar`). This makes things easier downstream. In this case, I think we don't really need the `"scalar"` key anymore.

src/probnum/linops/_kronecker.py
```diff
@@ -161,6 +161,44 @@ def _cond(self, p) -> np.inexact:
...
+            return NotImplemented
+
+        # Using (A (x) B) @ (C (x) D) = (A @ C) (x) (B @ D)
+        return Kronecker(A=_A @ _C, B=_B @ _D)
+
+    def _add_kronecker(
+        self, other: "Kronecker"
+    ) -> Union[NotImplementedType, "Kronecker"]:
+
+        if self.A == other.A:
```
marvinpfoertner, line 182: An additional check for `self.A is other.A` covers even more cases, even without the `__eq__` operator implemented

src/probnum/linops/_kronecker.py
```diff
@@ -417,3 +455,91 @@ def _cond_identical_factors(self, p) -> np.inexact:
...
+
+        super().__init__(
+            dtype=np.result_type(self.A.dtype, self.B.dtype),
+            shape=(
+                num_blocks * self.B.shape[0],
+                num_blocks * self.B.shape[1],
+            ),
+            matmul=lambda x: _kronecker_matmul(
+                self.A, self.B, x
+            ),  # TODO: is there a more efficient way?
```
marvinpfoertner, line 484: Yes, in both cases you can omit one of the `matmul`s. Let's discuss on monday

src/probnum/linops/_linear_operator.py
```diff
@@ -1079,3 +1114,98 @@ def _astype(
...
+    def _todense(self):
+        res = np.eye(self.shape[1], self.shape[1])
+        return _selection_matmul(self.indices, res)
+
+
+def _selection_matmul(indices, M):
+    return np.take(M, indices=indices, axis=-2)
+
+
+class Embedding(LinearOperator):
```
marvinpfoertner, line 1163: I think this one needs a little bit of class documentation since the name is somewhat non-obvious.
Btw. another name for a similar operation coming from the DL world would be `Scatter` which I personally find less confusing since it already indicates some sort of indexing stuff

src/probnum/linops/_scaling.py
```diff
@@ -124,6 +124,8 @@ def __init__(
 
             trace = lambda: self.shape[0] * self._scalar
         elif np.ndim(factors) == 1:
+            if np.all(factors == factors[0]):
+                self._scalar = factors[0]
```
marvinpfoertner, line 128: Personally, I would not do this check here, but rather implement a cached property which computes it on demand.
This avoids potentially unnecessary constructor overhead

src/probnum/linops/_scaling.py
```diff
@@ -269,3 +319,32 @@ def _cond_isotropic(self, p: Union[None, int, float, str]) -> np.inexact:
             )
         else:
             return np.linalg.cond(self.todense(cache=False), p=p)
+
+
+class Zero(_linear_operator.LinearOperator):
+    def __init__(self, shape, dtype=np.float64):
+
+        zero = np.zeros(shape=(), dtype=dtype)
+        matmul = lambda x: zero * x
```
marvinpfoertner, line 328: `np.zeros(x.shape, np.result_type(x, self.dtype))`?

src/probnum/randprocs/markov/discrete/_linear_gaussian.py
```diff
@@ -172,8 +172,9 @@ def _forward_rv_classic(
         new_cov = H @ crosscov + _diffusion * R
         info = {"crosscov": crosscov}
         if compute_gain:
-            if config.lazy_linalg:
-                gain = (new_cov.T.inv() @ crosscov.T).T
+            if config.matrix_free_linalg:
+                # gain = (new_cov.T.inv() @ crosscov.T).T
+                gain = crosscov @ new_cov.inv()
```
marvinpfoertner, line 177: Let's add a `.solve()` method to linops in the future

src/probnum/randvars/_normal.py
```diff
@@ -194,20 +194,20 @@ def __init__(
...
-                        or n != cov.B.shape[0]
-                        or n != cov.B.shape[1]
-                    ):
-                        raise ValueError(
-                            "Kronecker structured kernels must have factors with the same "
-                            "shape as the mean."
-                        )
-
                     compute_cov_cholesky = self._kronecker_cov_cholesky
+                    if mean.ndim == 2:
```
marvinpfoertner, line 198: What happened here?

marvinpfoertner (changes requested):
Nice! Let's discuss some details on monday.

Since we have `Zeros` now, could you also add the
```python
...
elif self._scalar == 0:
    ...
```
case in the constructor of `Scaling`?

marvinpfoertner (approved):

marvinpfoertner:
The PR can't be merged because patch coverage is slightly too low.
From my POV we can merge this anyway, since test coverage did improve significantly in the linop modules.
@JonathanWenger @pnkraemer can you merge this?

src/probnum/linops/_linear_operator.py
marvinpfoertner: Out of scope for this PR; requires decision in discussion #573

PR #515, opened 2021-08-11
Lorenz 96 system

pnkraemer:
# In a Nutshell
Introduce Lorenz96 system to the diffeq problem zoo. The previous "Lorenz" problem is now "Lorenz63". 

# Detailed Description
That was it. The Lorenz96 system is a populare test problem for (a) chaotic systems, and (b) high-dimensional problems (because its dimensionality can be varied freely).


src/probnum/problems/zoo/diffeq/_ivp_examples.py
```diff
@@ -485,3 +486,60 @@ def jac(t, y, params=params):
...
+    t0
+        Initial time. Default is 0.0
+    tmax
+        Final time. Default is 20.0
+    y0
+        *(shape=(N, ))* -- Initial value. Default is ``[1/F, ..., 1/F]``. `N` is the number of variables in the model.
+    num_variables
+        Number of variables in the model. If `y0` is specified, this argument is ignored
+        (and the number of variables is inferred from the dimension of the initial value).
+        If not, the initial value
```
nathanaelbosch, line 513: There seems to be something missing here

nathanaelbosch (changes requested):
Nice addition! I also didn't know Lorenz is actually called Lorenz63.

There is just a minor issue in one of the docstrings, afterwards this is good to go :+1:

PR #522, opened 2021-08-16
Remove globally disabled `pylint` messages

### LINKED ISSUE #469: Remove global `pylint` ignores
**Is your issue related to a problem? Please describe.**
Several subpackages (e.g. `diffeq`, `statespace`, `linops`) did already resolve all individual pylint ignores on the package level.
However, since we also have many of these ignores are also defined globally in our `pyproject.toml` file, new pylint errors emerged in the meantime.

**Describe the solution you'd like.**
- [x] Discuss which ignores should be retained long-term (in discussion #470)
- [x] Move all other ignores from `pyproject.toml` into `tox.ini`
- [x] Reformat `pyproject.toml` ignores (see example below) and add a short comment for each ignore which gives a reason why this is ignored
- [x] Refactor the script `.github/workflows/pylint_check.py` in the CI

**Sample `pyproject.toml` format**
```
[tool.pylint.messages_control]
disable = [
    # Exceptions suggested by Black:
    # https://github.com/psf/black/blob/7f75fe3669ebf0627b1b0476a6d02047e909b959/docs/compatible_configs.md#black-compatible-configurations
    "bad-continuation",
    "bad-whitespace",
    # Sometimes we want to use "mathematical notation" to name variables, e.g. `A` for matrices
    "invalid-name",
]
```

**Additional context**
I believe that we added these ignores globally in order to keep our local pylint runs from complaining too much when just calling `pylint src/`.

marvinpfoertner:
# In a Nutshell
This PR removes most globally disabled `pylint` messages.

# Detailed Description
When we introduced `pylint` for linting, we decided to disable different messages on a package level via `tox.ini` in order to be able to use `pylint` in our CI pipeline while keeping the flexibility to improve one package at a time. However, we noticed that this way we lacked a "global" run linting the package as a whole. Consequently, we decided to disable the union of all disabled messages to the global `pylint` configuration to be able to run `pylint` locally.
Unfortunately, this caused the CI to accept code that will not be accepted by our long-term `pylint` config, even if the message was removed from the per-package disable list.  
This PR

- removes all globally disabled `pylint` messages from `pyproject.toml` except the ones which we decided to keep long-term,
- disables these messages again on a per-package basis in `tox.ini`,
- reformats and documents the `pylint` configuration in `pyproject.toml`,
- adds an additional "global" `pylint` pass to `tox.ini`, and
- refactors the `pylint_check.py` script from the CI to ensure that the messages disabled in the global linting pass are exactly the union of the messages disabled in the per-module linting passes.

# Related Issues
Closes #469.


.github/workflows/pylint_check.py
```diff
@@ -1,85 +1,48 @@
...
+       it from the global linting pass in `tox.ini` such that pylint runs in the
+       strictest configuration possible.
+    2. You decided to disable a message in the global linting pass in `./tox.ini`..
+       In this case either add it to the list of messages that should be disabled in the
+       long term (`pyproject.toml`) or add it to the `GLOBAL_DISABLES` variable in this
+       python script (`./github/workflows/pylint_check.py`) if the message should only
+       be disabled in the short term.
     If you are not sure what exactly you are supposed to do, or if you think that this
-    message is wrong please feel free to ping @nathanaelbosch.
+    message is wrong please feel free to ping @marvinpfoertner or @nathanaelbosch.
```
pnkraemer, line 40: Thanks for adding this. Purely out of curiosity: how well do you think these "ping XYZ" error messages will age?

pyproject.toml
```diff
@@ -37,69 +37,44 @@ exclude = '''
...
+    "bad-continuation",
+    "bad-whitespace",
+    # We allow TODO comments in the following format: `# TODO (#[ISSUE NUMBER]): This needs to be done.`
+    "fixme",
+    # We want to use "mathematical notation" to name some of our variables, e.g. `A` for matrices
+    "invalid-name",
+    # In functions/methods which are part of the public API, we often change the type of argument variables when normalizing inputs for internal processing,
+    # e.g. we might use `pn.utils.as_numpy_scalar` to cast a `float` into a `np.double` or a `np.ndarray` into a `RandomVariable` by using `pn.randvars.asrandvar`.
+    "redefined-variable-type",
+    # Many more `disable`s are defined in `./tox.ini` on a per-module basis!
```
pnkraemer, line 54: [minor] would this be something that is removed once the last per-module is removed?

pyproject.toml
```diff
@@ -37,69 +37,44 @@ exclude = '''
...
+    "numpy",
+]
+load-plugins = [
+    "pylint.extensions.docstyle",
+    "pylint.extensions.mccabe",
+    "pylint.extensions.redefined_variable_type",
+    "pylint.extensions.overlapping_exceptions",
+    # Extensions that might be of interest in the future:
+    # "pylint.extensions.check_elif",
+    # "pylint.extensions.docparams",
```
pnkraemer, line 71: [minor] now that these lines are touched anyway, is this well-placed here? Would it be better as an issue/discussion instead?

tox.ini
```diff
@@ -41,37 +40,41 @@ commands = isort --profile black --check --diff .
 description = Code formatting with Black and isort (and possibly other tools in the future)
 basepython = python3
 deps =
-          {[testenv:black]deps}
-          {[testenv:isort]deps}
```
pnkraemer, line 45: are these indentation issues on purpose or the result of some kind of autoformatter?

tox.ini
```diff
@@ -41,37 +40,41 @@ commands = isort --profile black --check --diff .
...
+    pylint src/probnum/randvars --disable="too-many-arguments,too-many-locals,too-many-branches,too-few-public-methods,protected-access,unused-argument,no-else-return,duplicate-code,line-too-long,missing-function-docstring" --jobs=0
+    pylint src/probnum/utils --disable="no-else-return,line-too-long" --jobs=0
+    # Benchmark and Test Code Linting Pass
+    # pylint benchmarks --disable="unused-argument,attribute-defined-outside-init,missing-function-docstring" --jobs=0 # not a work in progress, but final
+    pylint benchmarks --disable="unused-argument,attribute-defined-outside-init,no-else-return,no-self-use,consider-using-from-import,line-too-long,missing-module-docstring,missing-class-docstring,missing-function-docstring" --jobs=0
+    pylint tests --disable="arguments-differ,redefined-outer-name,too-many-instance-attributes,too-many-arguments,too-many-locals,too-few-public-methods,protected-access,unnecessary-pass,unused-variable,unused-argument,unused-private-member,attribute-defined-outside-init,no-else-return,no-self-use,consider-using-from-import,duplicate-code,line-too-long,missing-module-docstring,missing-class-docstring,missing-function-docstring" --jobs=0
```
pnkraemer, line 80: holy shit :D are these all previous global ignores, just merged here? Or have you checked whether some of these are necessary? (I think moving is enough for this PR, but was curious).

pnkraemer (approved):
Thanks for taking care of this! I left a bunch of comments, all of which are mostly out of curiosity. I think the changes are fine already. :)

.github/workflows/pylint_check.py
marvinpfoertner: Good point, I guess we can take this line out and have people open an issue

pyproject.toml
marvinpfoertner: Yes. To be honest, I don't think this line is particularly well-placed here anymore. I'll remove it.

pyproject.toml
marvinpfoertner: I'll just enable the plugins and disable their messages per-package.
That way we can resolve them like any other message.

tox.ini
marvinpfoertner: They're on purpose. The file's indentation was sort of messed up (a combination of 4 and 5 space indents and tabs) so I went ahead and changed it.

tox.ini
marvinpfoertner: I manually checked all of them so this should be the minimal list for each package

.github/workflows/pylint_check.py
nathanaelbosch: I completely agree, people should open issues instead of pinging people (or ping them in the new opened issue) :+1:

pyproject.toml
nathanaelbosch: Is it already clear that we want those enabled? Depending on that I agree that you could just enable them, or we might need a discussion on what we actually want to enforce.

tox.ini
```diff
@@ -41,37 +40,41 @@ commands = isort --profile black --check --diff .
...
+    pylint src/probnum/problems --disable="too-many-arguments,too-many-locals,unused-variable,unused-argument,consider-using-from-import,duplicate-code,line-too-long,missing-module-docstring,missing-function-docstring" --jobs=0
+    pylint src/probnum/quad --disable="function-redefined,too-many-arguments,line-too-long,missing-module-docstring,missing-function-docstring" --jobs=0
+    pylint src/probnum/randprocs --disable="arguments-differ,arguments-renamed,too-many-instance-attributes,too-many-arguments,too-many-locals,protected-access,unused-argument,no-else-return,duplicate-code,line-too-long,missing-module-docstring,missing-function-docstring" --jobs=0
+    pylint src/probnum/randvars --disable="too-many-arguments,too-many-locals,too-many-branches,too-few-public-methods,protected-access,unused-argument,no-else-return,duplicate-code,line-too-long,missing-function-docstring" --jobs=0
+    pylint src/probnum/utils --disable="no-else-return,line-too-long" --jobs=0
+    # Benchmark and Test Code Linting Pass
+    # pylint benchmarks --disable="unused-argument,attribute-defined-outside-init,missing-function-docstring" --jobs=0 # not a work in progress, but final
```
nathanaelbosch, line 78: Is this comment supposed to indicate what the final pylint exceptions should be?

nathanaelbosch (approved):

pyproject.toml
marvinpfoertner: We certainly want `docparams` enabled and `check_elif` is also a good idea. I'll enable it for now and we can take it out if it causes trouble in the future. Is this fine @pnkraemer @nathanaelbosch?

tox.ini
marvinpfoertner: Yes

marvinpfoertner:
@pnkraemer @nathanaelbosch I incorporated your review comments.
If you're fine with the changes, please go ahead and merge.
I can't, since test coverage did not increase.

pyproject.toml
pnkraemer: For me this is fine. docparams will detect changes that one often forgets when refactoring (at least I do...). check_elif seems okay to me as well

PR #524, opened 2021-08-30
PEP517 compliant build and version number from file

marvinpfoertner:
# In a Nutshell
This PR makes the ProbNum build PEP517 compliant.
Moreover, we now read the `__version__` number in `__init__.py` from a `_version.py` file generated by `setuptools_scm` at build time.

# Detailed Description
See above

# Related Issues
None


JonathanWenger (approved):
Thanks for taking care of this!

PR #532, opened 2021-09-08
Scale the ODE filter error estimate with the current step size

nathanaelbosch:
# In a Nutshell
To solve stiff ODEs, the local error estimate needs to be scaled with the step size.

# Detailed Description
I found that the current version of the EK1 was not able to solve the stiff Van-der-Pol problem as formulated here:
https://diffeq.sciml.ai/stable/types/ode_types/#DiffEqProblemLibrary.ODEProblemLibrary.prob_ode_vanstiff.
With this PR, solving this problem becomes possible.

Maybe I should add a test that makes sure that this capability is not lost in the future, e.g. solving VdP on a very small time span. But I first need to find such a setting that does not take too much time to solve, while also being a problem that did not work before, but does now. I'm open for suggestions and edits of this PR :)


nathanaelbosch:
Thie problem I tried to solve here does not seem to exist, therefore I'm closing this for now.

PR #534, opened 2021-09-14
linops quickstart

mmahsereci:
# In a Nutshell
New linear operators quickstart.

# Detailed Description
content
- creation with `Matrix` from Numpy arrays and scipy sparse matrices
- matrix-free definition
- linops arithmetics
- brushes on compatibility with randvars

not included (should go in other tutorials)
- Kronecker operators
- lazy evaluation and context manager
- a very high dim example 

# Related Issues
Closes #...


docs/source/tutorials/linops/linear_operators_quickstart.ipynb
```diff
@@ -0,0 +1,622 @@
+{
```
JonathanWenger, line 1: <p>finite-dimensional [operator / vector space] instead of "finite". Just finite has a (potentially) different meaning in both contexts, namely finite-rank and finite cardinality (see e.g. https://en.wikipedia.org/wiki/Examples_of_vector_spaces#Finite_vector_spaces).</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/probabilistic-numerics/probnum/pull/534/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

JonathanWenger, line 1: <p>are built from their ...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/probabilistic-numerics/probnum/pull/534/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='12'/>

JonathanWenger, line 1: <p>Maybe make more clear the memory requirement benefit:</p><p> </p><p>"This is an implicit definition that does not require the storage of the $n^2$ entries of the matrix form of $A$...</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/probabilistic-numerics/probnum/pull/534/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

JonathanWenger (approved):
Thanks for taking the time to write the quickstart. I have only one general comment:

I slightly missed that linear operators are not only about memory efficiency (since dense matrices are not stored), but also about efficient matrix-vector products. Namely, that the implicit definition via `matvec` may be much faster than the naive O(n^2). I think one or two sentences to that effect would drive home the point in the tutorial.

docs/source/tutorials/linops/linear_operators_quickstart.ipynb
mmahsereci: <p>done</p>

mmahsereci: <p>I think we need a relative clause here, don't we? </p>

mmahsereci: <p>changed</p>

mmahsereci:
That's a good point. Is there a word for it? Can we call them "compute-efficient"?

docs/source/tutorials/linops/linear_operators_quickstart.ipynb
JonathanWenger: <p>Correct, but the sentence is in passive, which is why the past tense is used, i.e. "built". See also https://www.linguee.com/english-german/translation/houses+are+built.html</p>

JonathanWenger:
> That's a good point. Is there a word for it? Can we call them "compute-efficient"?

Sure. Or just "computed efficiently" or something like that? Really it's about memory *and* time.

docs/source/tutorials/linops/linear_operators_quickstart.ipynb
mmahsereci: ah, I thought you wanted me to skip the word "that" :bulb: . corrected.

mmahsereci:
yeah, was looking for a catchy word for the sub-section titles so that the message is not lost between the lines. I'll come  up with something.

mmahsereci:
The text is adjusted now and incorporates info about efficient matrix-vector products. Could you give me another quick thumbs-up if I can leave it at that @JonathanWenger ?

PR #544, opened 2021-10-18
Improved documentation for ODE examples

JonathanWenger:
# In a Nutshell
Documentation for the ODE examples did not always correspond to the parameter choices made in the implementation. This was corrected in this PR with some additional improvements.

# Detailed Description
- changed documentation to match the implementation for the `lotka_volterra` and `rigid_body` IVPs
- removed default arguments from docstring if they were given as default arguments in the signature to avoid similar inconsistencies in the future.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -57,7 +57,7 @@ exclude =
 
 # Autodiff backends
 jax =
-    jax<0.2.22; platform_system!="Windows"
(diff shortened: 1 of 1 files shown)


schmidtjonathan:
Thanks a lot for fixing these! I left a question but aside from that ... 👍

PR #556, opened 2021-10-29
Langevin Stein kernel and quadrature test problems

fxbriol:
# In a Nutshell
Adds Langevin Stein kernels and some quadrature test problems.

# Detailed Description
Langevin Stein kernels are characterized by the property that their kernel mean embedding can be chosen arbitrarily. This property is particularly useful for Bayesian quadrature where kernel mean embeddings are needed to compute a belief over the integrand.

### TODO
- [ ] Tests
- [ ] Test problems

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -72,7 +72,8 @@ full =
 
 # Dependencies needed to run the tests (semicolon/line-separated)
 test =
-    pytest>=4.6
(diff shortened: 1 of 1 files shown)


JonathanWenger:
@fxbriol Could you link the website with the test functions for quadrature, that you mentioned offline?

PR #559, opened 2021-11-04
Rename `lazy_linalg` config option `matrix_free`

JonathanWenger:
# In a Nutshell
This PR gives a more descriptive name (`matrix_free`) to the configuration option which uses linear operators instead of full matrices.

# Detailed Description
The term `matrix_free` is a more complete description of what functionality `LinearOperator`s offer, namely that they
- only implicitly define the matrix-vector product
- generally do not instantiate a full matrix in memory
- are memory and time-efficient
- offer lazy arithmetic

### DIFF
--- a/benchmarks/randprocs.py
+++ b/benchmarks/randprocs.py
@@ -8,11 +8,11 @@ from probnum import config, linops, randprocs, randvars
 class MarkovProcessSampling:
     """Benchmark sampling from Markov processes."""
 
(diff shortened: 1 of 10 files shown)


JonathanWenger:
Since this is a pure renaming PR I will not write tests for integrated wiener processes or the preconditioner in the integrator subpackage. These should be added in a PR focussed on those areas.

mmahsereci (approved):

Issue #557, opened 2021-11-03
`StoppingCriterion` base class with boolean arithmetic


marvinpfoertner:
**Is your feature request related to a problem? Please describe.**
With the current implementation of stopping criteria, we can only use one criterion at a time.

**Give an example use case.**
For example, in a linear solver, we might want to be able to use a maximal number of iterations as a base stopping criterion, but at the same time we might want to stop early if the residual error and the posterior contraction are low. 

**Describe the solution you'd like.**
Overriding `__and__`, `__or__` and `__invert__` in a common base class for stopping criteria enables stopping criterion arithmetic. In this case, we will be able to express the example above as

```python
stopping_criterion = MaxIterations(1000) | (ResidualNorm(1e-10) & PosteriorContraction(1e-7))
```

PR #561, opened 2021-11-04
Base class `StoppingCriterion` for stopping criteria of probabilistic numerical methods.

### LINKED ISSUE #432: A single interface for stopping criteria in probnum
I think with #420, #283, and the existing stopping criteria in `filtsmooth`, we seem to be on our best way to introduce three objects that do mostly the same. 

While for the current PRs, this might be fine, we should keep this in mind and at least find a common interface, preferably use the same class of objects for all three. All three versions seem to compare absolute and relative errors, for example.

JonathanWenger:
# In a Nutshell
Introduces a common base class for stopping criteria and moves abstract components of PNMs into a dedicated subpackage.

# Detailed Description
The common base class `StoppingCriterion` supports boolean arithmetic, allowing for seamless combination of desired stopping criteria when constructing a (probabilistic) numerical method.

For example we can now do the following:
```python
stopping_criterion = MaxIterations(1000) | (ResidualNorm(1e-10) & PosteriorContraction(1e-7))
```

# Related Issues
Closes #432 and #557


src/probnum/filtsmooth/optim/_iterated_component.py
```diff
@@ -61,7 +61,9 @@ def __init__(
     ):
         self._component = component
         self.stopcrit = (
-            _stoppingcriterion.StoppingCriterion() if stopcrit is None else stopcrit
+            _stopping_criterion.FiltSmoothStoppingCriterion()
+            if stopcrit is None
+            else stopcrit
```
mmahsereci, line 66: multi-lines of these if-else constructions are not recommended due to readability.

src/probnum/filtsmooth/optim/_state_space_optimizer.py
```diff
@@ -2,14 +2,14 @@
 
 import abc
 
-from probnum.filtsmooth.optim import _stoppingcriterion
+from probnum.filtsmooth.optim import _stopping_criterion
```
mmahsereci, line 5: same here, simply import `FiltSmoothStoppingCriterion` directly.

src/probnum/filtsmooth/optim/_stopping_criterion.py
```diff
@@ -1,8 +1,10 @@
 """Stopping criteria for iterated filtering and smoothing."""
 import numpy as np
 
+from probnum import pnm
```
mmahsereci, line 4: Is there a reason you import the whole `pnm` module? same in other places. It would be good to at least do it uniformly and have an agreed style that we stick to.

src/probnum/pnm/_stopping_criterion.py
```diff
@@ -0,0 +1,92 @@
+"""Stopping criterion for a (probabilistic) numerical method."""
+
+import abc
+
+
+class StoppingCriterion(abc.ABC):
+    r"""Stopping criterion of a (probabilistic) numerical method.
```
mmahsereci, line 7: why is probabilistic in brackets? I suppose you mean that some of the stopping criteria can also be used by non-probabilistic methods. But since ProbNum only contains PNMs, we can drop the brackets here.

src/probnum/pnm/_stopping_criterion.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Now let's modify the state such that the solver has reached a maximum number of iterations.
+
+    >>> state.iters = 1000
+    >>> stopcrit(state)
+    True
+
+    See Also
+    --------
+    LinearSolverStoppingCriterion : Stopping criterion of a probabilistic linear solver.
```
mmahsereci, line 67: Possibly add the FiltSmoothStoppingCriterion here.

mmahsereci (approved):
This looks neat, thanks a lot @JonathanWenger . I suppose the PR does not only introduce a common class for stopping criteria but also a new module `pnm` for shared base classes. I just have some minor comments on style of imports.

src/probnum/filtsmooth/optim/_state_space_optimizer.py
JonathanWenger: For some reason even without adding `StoppingCriterion` this change leads to cyclic imports.

JonathanWenger: Note that this issue is already present on `main` and is not introduced by this PR. Likely there are some cyclic dependencies within the `filtsmooth` package causing this.

JonathanWenger:
@marvinpfoertner This is now ready for review.

src/probnum/_pnmethod/_stopping_criterion.py
```diff
@@ -0,0 +1,118 @@
...
+
+    @abc.abstractmethod
+    def __call__(self, *args, **kwargs) -> bool:
+        """Check whether tracked quantities meet a desired terminal condition."""
+        raise NotImplementedError
+
+    def __and__(self, other):
+        return LambdaStoppingCriterion(
+            stopcrit=lambda *args, **kwargs: self(*args, **kwargs)
+            and other(*args, **kwargs)
```
marvinpfoertner, line 80:
```suggestion
            stopcrit=lambda *args, **kwargs: self(*args, **kwargs)
            & other(*args, **kwargs)
```

src/probnum/_pnmethod/_stopping_criterion.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    def __or__(self, other):
+        return LambdaStoppingCriterion(
+            stopcrit=lambda *args, **kwargs: self(*args, **kwargs)
+            or other(*args, **kwargs)
```
marvinpfoertner, line 86:
```suggestion
            stopcrit=lambda *args, **kwargs: self(*args, **kwargs)
            | other(*args, **kwargs)
```

src/probnum/_pnmethod/_stopping_criterion.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+
+    def __invert__(self):
+        return LambdaStoppingCriterion(
+            stopcrit=lambda *args, **kwargs: not self(*args, **kwargs)
```
marvinpfoertner, line 91:
```suggestion
            stopcrit=lambda *args, **kwargs: ~self(*args, **kwargs)
```

src/probnum/filtsmooth/optim/_state_space_optimizer.py
marvinpfoertner: Things like this should be caught by `pylint` once we enable it globally

marvinpfoertner (approved):

src/probnum/_pnmethod/_stopping_criterion.py
JonathanWenger: I don't think this will work, since 
```python
~True == -2
~False == -1
```

JonathanWenger: This fails the tests since
```python
~True == -2
~False == -1
```

marvinpfoertner: Ah damn, I made a mistake here! The other two occurences should actually be `and` and `or`, too. I somehow didn't realize that these are actually boolean operations. Sorry!

PR #564, opened 2021-11-08
Langevin Stein Kernel

JonathanWenger:
# In a Nutshell
Adds Langevin Stein kernels and some quadrature test problems.

# Detailed Description
Langevin Stein kernels are characterized by the property that their kernel mean embedding can be chosen arbitrarily. This property is particularly useful for Bayesian quadrature where kernel mean embeddings are needed to compute a belief over the integrand.

### TODO
- [ ] Tests
- [ ] Test problems


JonathanWenger:
@fxbriol Could you link the website with the test functions for quadrature, that you mentioned offline?

fxbriol:
This website: https://www.sfu.ca/~ssurjano/integration.html has the "Genz test functions". This is a set of six test functions each with different challenges for numerical integration. The advantage of these test functions is that they are defined for some arbitrary dimension d (so you can try your methods in high dimensions!) and each have a parameter controlling the difficulty of integrating them (basically this parameter reinforces the features that make them challenging to integrate). On top of this, regardless of the value of d or of this parameter, we have a closed form value of the integral when integrating against the Lebesgue measure. 

The website has some code in Matlab or R for each of these, but not Python. However, I think I have some code for them somewhere. I will try to find it.

mmahsereci:
hey, just jumping in real quick. For the test problems, let's open another PR as it is quite separate from the Stein Kernels. I know it may seem not necessary for this one PR, but let's keep the standards high in this codebase. Pinging @JonathanWenger @fxbriol .

JonathanWenger:
Closing this due to inactivity for now.

PR #565, opened 2021-11-08
[PLS Refactor] Base class `ProbabilisticLinearSolver`

JonathanWenger:
# In a Nutshell
Adds the base class `ProbabilisticLinearSolver` which allows the custom composition of probabilistic linear solvers for the solution of linear systems.

# Detailed Description
The class `ProbabilisticLinearSolver` enables customized creation of linear solvers from the components defined in `probnum.linalg.solvers`. It implements a `solve` method to be called on `LinearSystem`s and allows stepping through the solve one iteration at a time by looping through the `solve_iterator()`.

This code also adds convenience classes such as `BayesCG`, `ProbabilisticKaczmarz`, `SymMatrixBasedPLS` which already configure part of the components (e.g. a particular policy and belief update).

### Example Code
```python
import numpy as np
from probnum.problems import LinearSystem
from probnum.problems.zoo.linalg import random_spd_matrix

# Define a linear system.

rng = np.random.default_rng(42)
n = 100
A = random_spd_matrix(rng=rng, dim=n)
b = rng.standard_normal(size=(n,))
linsys = LinearSystem(A=A, b=b)

# Create a custom probabilistic linear solver from pre-defined components.

from probnum.linalg.solvers import (
    ProbabilisticLinearSolver,
    belief_updates,
    beliefs,
    information_ops,
    policies,
    stopping_criteria,
)

pls = ProbabilisticLinearSolver(
    policy=policies.ConjugateGradientPolicy(),
    information_op=information_ops.ProjectedRHSInformationOp(),
    belief_update=belief_updates.solution_based.SolutionBasedProjectedRHSBeliefUpdate(),
    stopping_criterion=stopping_criteria.MaxIterationsStoppingCriterion(100)
    | stopping_criteria.ResidualNormStoppingCriterion(atol=1e-5, rtol=1e-5),
)

# Define a prior over the solution.

from probnum import linops, randvars

prior = beliefs.LinearSystemBelief(
    x=randvars.Normal(
        mean=np.zeros((n,)),
        cov=np.eye(n),
    ),
)

# Solve the linear system using the custom solver.

belief, solver_state = pls.solve(prior=prior, problem=linsys)
np.linalg.norm(linsys.A @ belief.x.mean - linsys.b) / np.linalg.norm(linsys.b)
# 7.1886e-06

```

# Related PRs
Implements largest missing part of #283.


JonathanWenger:
Note: #579 should be reviewed first, since it is used in this PR.

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
```diff
@@ -35,9 +60,279 @@ class ProbabilisticLinearSolver:
...
+    ...     policies,
+    ...     stopping_criteria,
+    ... )
+
+    >>> pls = ProbabilisticLinearSolver(
+    ...     policy=policies.ConjugateGradientPolicy(),
+    ...     information_op=information_ops.ProjectedRHSInformationOp(),
+    ...     belief_update=belief_updates.solution_based.SolutionBasedProjectedRHSBeliefUpdate(),
+    ...     stopping_criterion=stopping_criteria.MaxIterationsStoppingCriterion(100)
+    ...     | stopping_criteria.ResidualNormStoppingCriterion(atol=1e-5, rtol=1e-5),
```
marvinpfoertner, line 96:
```suggestion
    ...     stopping_criterion=(
    ...         stopping_criteria.MaxIterationsStoppingCriterion(100)
    ...         | stopping_criteria.ResidualNormStoppingCriterion(atol=1e-5, rtol=1e-5)
    ...     ),
```

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
```diff
@@ -35,9 +60,279 @@ class ProbabilisticLinearSolver:
...
+        ------
+        solver_state
+            State of the probabilistic linear solver.
+        """
+        solver_state = LinearSolverState(problem=problem, prior=prior, rng=rng)
+
+        while True:
+
+            yield solver_state
+
```
marvinpfoertner, line 158: Since we pass the stopping criterion to the constructor and not the `solve` method, I think we should check the stopping criterion here and return if it is met. Otherwise the behavior of the iterator would seem somewhat unintuitive to me.

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
```diff
@@ -35,9 +60,279 @@ class ProbabilisticLinearSolver:
...
+        problem
+            Linear system to be solved.
+        rng
+            Random number generator.
+
+        Returns
+        -------
+        belief
+            Posterior belief :math:`(\mathsf{x}, \mathsf{A}, \mathsf{H}, \mathsf{b})`
+            over the solution :math:`x`, the system matrix :math:`A`, its (pseudo-)inverse :math:`H=A^{-1}` and the right hand side :math:`b`.
```
marvinpfoertner, line 192:
```suggestion
            over the solution :math:`x`, the system matrix :math:`A`, its (pseudo-)inverse :math:`H=A^\dagger` and the right hand side :math:`b`.
```

src/probnum/linalg/solvers/beliefs/_linear_system_belief.py
```diff
@@ -152,3 +153,12 @@ def _induced_x(self) -> randvars.RandomVariable:
...
         return self.Ainv @ self.b
+
+    def _induced_Ainv(self) -> randvars.RandomVariable:
+        r"""Induced belief about the inverse from a belief about the solution.
+
+        Computes a consistent belief about the inverse from a belief about the solution.
+        """
+        return randvars.Constant(
+            linops.Scaling(factors=0.0, shape=(self._x.shape[0], self._x.shape[0]))
+        )
```
marvinpfoertner, line 164: Isn't this a little dangerous, especially given the docstring?
How about:

```suggestion
        raise NotImplementedError()
```

marvinpfoertner (approved):
A few minor comments, but otherwise really nice 👍🏻

src/probnum/linalg/solvers/beliefs/_linear_system_belief.py
JonathanWenger: I removed the induced Ainv belief since this is not yet needed. Instead the solution-based belief update which iteratively builds up an approximation to the inverse now creates an inconsistent belief with the solution. This is okay for now since we explicitly do not guarantee this. In a future PR we can implement the rank-1 prior mean construction for A and H in Proposition S5 of https://arxiv.org/pdf/2010.09691.pdf from an initial x_0.

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
```diff
@@ -156,13 +158,17 @@ def solve_iterator(
 
             yield solver_state
 
+            # Check stopping criterion
+            if self.stopping_criterion(solver_state=solver_state):
+                break
```
marvinpfoertner, line 163: It might be useful to check the stopping criterion before the `yield`, write the result into the `state` as a `bool`, then `yield` and then `break` depending on the value of that `bool`. This way the user knows whether the solver is about to stop. However, from my POV, this is out of scope for this PR

marvinpfoertner (approved):
👍🏻

PR #568, opened 2021-11-15
Move `kernels` subpackage into `randprocs`

marvinpfoertner:
# In a Nutshell
This PR moves the `kernels` subpackages to `randprocs.kernels`

# Detailed Description
See above

# Related Issues
None


JonathanWenger (approved):
:+1:

PR #570, opened 2021-11-17
BQ stopping criteria use base class

mmahsereci:
# In a Nutshell

- The BQ stopping criteria use the `StoppingCriterion` base class now which was introduced in #561 
- Introduced structure is similar to structure is linalg module (API references not fixed yet)

# Related Issues

Addresses some of the point still open from #420


src/probnum/quad/solvers/bayesian_quadrature.py
```diff
@@ -4,19 +4,19 @@
 
 import numpy as np
 
+from probnum.quad.solvers.policies import Policy, RandomPolicy
+from probnum.quad.solvers.stopping_criteria import (
+    BQStoppingCriterion,
```
JonathanWenger, line 9: "IntegralVarianceTolerance", "MaxNevals", etc. are enough if they are grouped under probnum.quad.stopping_criteria. This way we don't need to duplicate the information contained in the package structure in the name. We will implement this change for `linalg` as well. The convention should then be to always import solver components as `stopping_criteria.MaxNevals`, for example.

src/probnum/quad/__init__.py
```diff
@@ -39,13 +40,16 @@
...
-    "StoppingCriterion",
+    "BQStoppingCriterion",
+    "IntegralVarianceToleranceStoppingCriterion",
+    "MaxNevalsStoppingCriterion",
+    "RelativeMeanChangeStoppingCriterion",
 ]
 
 # Set correct module paths. Corrects links and module paths in documentation.
 BayesianQuadrature.__module__ = "probnum.quad"
+BQStoppingCriterion.__module__ = "probnum.quad"
```
JonathanWenger, line 51: I would not move component interfaces up to `quad` but keep them in `quad.stopping_criteria`

src/probnum/quad/__init__.py
```diff
@@ -6,28 +6,29 @@
...
 )
-from .policies import Policy, RandomPolicy
-from .stop_criteria import (
-    IntegralVarianceTolerance,
-    MaxNevals,
-    RelativeMeanChange,
-    StoppingCriterion,
+from .solvers import (
+    BayesianQuadrature,
+    BQBeliefUpdate,
```
JonathanWenger, line 28: I would suggest moving components like `BQBeliefUpdate` and `BQStandardBeliefUpdate` under `quad.belief_updates`

src/probnum/quad/solvers/bayesian_quadrature.py
```diff
@@ -49,7 +49,7 @@ def __init__(
         measure: IntegrationMeasure,
         policy: Policy,
         belief_update: BQBeliefUpdate,
-        stopping_criteria: List[StoppingCriterion],
+        stopping_criteria: List[BQStoppingCriterion],
```
JonathanWenger, line 52: With the new `probnum.StoppingCriterion` this does not have to be a list anymore, but can just be `BQStoppingCriterion` and then the defaults below are defined via arithmetic.

JonathanWenger (changes requested):
Thanks for taking care of this! Great that you switched to the new `StoppingCriterion` interface so quickly. I have a few comments around naming and organization of `probnum.quad` as well as the functionality of `StoppingCriterion`. 

Finally I would recommend structuring the docs more in depth. Currently everything is visible under `probnum.quad` instead of in subfolders for components (see `linalg` as an example). This can and probably should be a separate PR, but I thought I'd mention it here since it is relevant for the subpackage structure (i.e. having a `probnum.quad.stopping_criteria` subpackage).

src/probnum/quad/__init__.py
mmahsereci: yes, but not in this PR. There are a couple of more things to address in the BQ package (see list at the end of #420 ). One of them is the directory structure, another one is the API reference structure (what the user sees). This PR simply enables the existing stopping criteria from the new base class, and so I moved the already to the correct directory.

src/probnum/quad/solvers/bayesian_quadrature.py
mmahsereci: Ah, I see, I simply mirrored the naming of the current Linalg solvers API. I'll change this then.

src/probnum/quad/solvers/bayesian_quadrature.py
mmahsereci: good point!

src/probnum/quad/__init__.py
mmahsereci: ah, sorry, I just saw your general comment now. I'd like to fix the API reference structure once the code structural changes are somewhat addressed, and the directory structure is more solid. I may ping you for this again, so it's in line with the changes in the linalg package.

mmahsereci: #574

src/probnum/quad/solvers/bayesian_quadrature.py
mmahsereci: I noted that the `bayesquad` solver used to return the type of stopping criterion that triggered in an "or" construction as info which may be quite useful for the user. With the new `LambdaStoppingCriterion` structure this is not easily possible anymore, so I removed that feature. But it might actually be quite useful in general for the user to know which of the stropping criteria the custom lambda criterion is constructed off triggered and which did not.

JonathanWenger (approved):

PR #558, opened 2021-11-03
Bugfix and additional tests for matrix-variate normal arithmetic

JonathanWenger:
# In a Nutshell
Previously attempting to do a matrix-vector product between a matrix-variate normal and a vector of shape (n,) raised an error:

```python
import numpy as np
from probnum import randvars
from probnum import linops
n = 5
X = randvars.Normal(mean=np.zeros((n, n)), cov=linops.Kronecker(np.eye(5), np.eye(5)))
b = np.ones((n,))

X @ b
# TypeError: The given shape (5,) must have 2 dimensions.
```
This PR fixes the bug and adds the corresponding tests.

# Detailed Description
Some of the normal arithmetic tests were also refactored in the process.

# Related Issues
Closes #456.

PR #572, opened 2021-11-18
1x1 Matrix-variate Normal Multiplication Bugfix

JonathanWenger:
# In a Nutshell
Fixes a bug where 1x1 random matrix multiplication with a vector would raise an error.

# Detailed Description
The case where a matrix-variate normal of `shape=(1,1)` was multiplied with a vector of `shape=(1,)` was previously not tested and actually raised an error due to the covariance having `shape=()` not `shape=(1,1)`. This PR fixes this bug.

Minimal reproducible example of bug:

```python
import numpy as np
from probnum import randvars, linops

A = randvars.Normal(np.zeros((1, 1)), linops.Kronecker(np.ones((1, 1)), np.ones((1, 1))))
A @ np.ones((1,))

# ValueError: The covariance matrix must be of shape (1, 1), but shape () was given.
```

# Related PRs
Adds a missing test case to #558


marvinpfoertner (approved):
I guess the easier fix would have been to remove

```python
 or (norm_rv.ndim == 2 and norm_rv.shape[0] == 1)
```

from https://github.com/probabilistic-numerics/probnum/blob/e5bdbe508c421d3f03ff388356421c375371b3c7/src/probnum/randvars/_arithmetic.py#L281 but this is more performant. Nice 👍🏻

PR #576, opened 2021-11-19
Invert semantics of `lazy_matrix_matrix_matmul` config option

marvinpfoertner:
# In a Nutshell
Currently, the `lazy_matrix_matrix_matmul` config option computes `Matrix` @ `Matrix` lazily if it is set to `False`, which is counterintuitive.
This PR inverts the semantics of the flag such that setting it to `True` enables lazy computation.

# Detailed Description
- inverted the semantivs of `lazy_matrix_matrix_matmul` config flag
- refactored the documentation of `matrix_free` and `lazy_matrix_matrix_matmul` to link `LinearOperator` and `Matrix` to the API docs

# Related Issues
None

### DIFF
--- a/src/probnum/_config.py
+++ b/src/probnum/_config.py
@@ -130,18 +130,23 @@ _DEFAULT_CONFIG_OPTIONS = [
         "matrix_free",
         False,
         (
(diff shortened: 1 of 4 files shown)


JonathanWenger (approved):
👍🏽

PR #579, opened 2021-11-23
Make `MaxIterations` stopping criterion stateless

JonathanWenger:
# In a Nutshell
The stopping criterion for the linear solver `MaxIterationsStoppingCriterion` is now stateless.

# Detailed Description
Previously, calling `MaxIterationsStoppingCriterion(solver_state)` modified its internal state, namely the `self.maxiter` variable. This is problematic since it causes strange bugs where two solvers may influence each other. Take the following (simplified) example.

```python
from probnum.linalg.solvers import BayesCG, stopping_criteria

# Define solvers
maxiter_stopcrit = stopping_criteria.MaxIterationsStoppingCriterion()
solver1 = BayesCG(stopping_criterion=maxiter_stopcrit)
solver2 = BayesCG(stopping_criterion=maxiter_stopcrit)

solver1.solve(problem1) # This modifies maxiter_stopcrit.maxiter to 10*problem1.A.shape[0]

solver2.solve(problem2) # This now takes at most 10*problem1.A.shape[0] iterations. This is very problematic if problem2 has a very different size!
```


mmahsereci (approved):
looks good

PR #580, opened 2021-11-24
Reorthogonalization for a stable implementation of the probabilistic linear solver

JonathanWenger:
# In a Nutshell
Add reorthogonalization to the probabilistic linear solver for better numerical stability.

# Detailed Description
Krylov subspace methods which rely on A-conjugate search directions (or orthogonal residuals) typically need reorthogonalization to attain numerical stability. This PR adds a `reorthogonalization_fn` argument to the `ConjugateGradientPolicy`.


src/probnum/utils/linalg/_orthogonalize.py
```diff
@@ -0,0 +1,165 @@
...
+        norm_fn = lambda v: euclidean_norm(v, A=inprod)
+    else:
+        inprod_fn = inprod
+        norm_fn = lambda v: np.sqrt(inprod_fn(v, v))
+
+    v_orth = v.copy()
+
+    for u in orthogonal_basis:
+        v_orth -= (inprod_fn(u, v) / inprod_fn(u, u)) * u
+
```
treid5, line 59: # The loop in 57-58 can be vectorized
# if U is the orthonormal basis, then v_orth = v - UU^TAv
basis_norm = inprod_fn(orthogonal_basis,orthogonal_basis) # Compute norm of basis
orthonormal_basis = (basis_norm)**(-1/2)*orthogonal_basis.T).T # Divide basis vectors by norm
v_orth -= orthonormal_basis @ inprod_fn(orthonormal_basis,v)

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
```diff
@@ -11,21 +14,101 @@ class ConjugateGradientPolicy(_linear_solver_policy.LinearSolverPolicy):
...
+                residual, prev_residual = self._reorthogonalized_residuals(
+                    solver_state=solver_state
+                )
+            else:
+                residual = solver_state.residual
+                prev_residual = solver_state.residuals[solver_state.step - 1]
 
+            # A-conjugacy correction (in exact arithmetic)
+            beta = (np.linalg.norm(residual) / np.linalg.norm(prev_residual)) ** 2
             action += beta * solver_state.actions[solver_state.step - 1]
```
treid5, line 66: Is line 65 correct? In CG, the next search direction is computed as s_{i+1} = r_{i+1} + beta s_{i}. Here it appears to be computed as s_{i+1} = s_i + bets s_i. This line is not changed in this pull request, so it is probably not an error.

src/probnum/utils/linalg/_inner_product.py
```diff
@@ -0,0 +1,65 @@
+"""Functions defining useful inner products."""
+
+from typing import Optional, Union
+
+import numpy as np
+
+from probnum import linops
+
+
```
treid5, line 9: This is a stylistic comment. Typically "euclidean" norm and inner product only refers to the 2-norm and the inner product that induces it.

src/probnum/linalg/solvers/_state.py
```diff
@@ -31,20 +31,23 @@ def __init__(
         prior: "probnum.linalg.solvers.beliefs.LinearSystemBelief",
         rng: Optional[np.random.Generator] = None,
     ):
+        self.rng: Optional[np.random.Generator] = rng
```
marvinpfoertner, line 34: Do we want this to be part of the state? Considering the `ad-prototype` I would suggest passing it to the `solve_iterator` and forward it to the component which requires an rng, e.g. the policy.

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
```diff
@@ -11,21 +14,101 @@ class ConjugateGradientPolicy(_linear_solver_policy.LinearSolverPolicy):
...
+        self._reorthogonalization_fn_action = reorthogonalization_fn_action
+
     def __call__(
         self, solver_state: "probnum.linalg.solvers.LinearSolverState"
     ) -> np.ndarray:
 
         action = -solver_state.residual.copy()
 
+        if self._reorthogonalization_fn_residual is not None and solver_state.step == 0:
+            solver_state.cache["reorthogonalized_residuals"] = [solver_state.residual]
```
marvinpfoertner, line 51: These "initial checks" can be avoided and made less error-prone by turning `state.cache` into a `collections.defaultdict(list)`

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
marvinpfoertner: Should be, since action is set to `-solver_state.residual.copy()` at the beginning of the method. Or do we need the reorthogonalized residual here?

src/probnum/utils/linalg/_inner_product.py
marvinpfoertner: How about we just call these functions `inprod` and `induced_norm`, since every (semi-)inner product on R^n can be represented by a positive (semi-)definite matrix `A`?

src/probnum/linalg/solvers/_state.py
```diff
@@ -31,20 +31,23 @@ def __init__(
...
-
+        # Caches
         self._actions: List[np.ndarray] = [None]
         self._observations: List[Any] = [None]
         self._residuals: List[np.ndarray] = [
             self.problem.A @ self.belief.x.mean - self.problem.b,
-            None,
         ]
-        self.rng: Optional[np.random.Generator] = rng
+        self.cache: Dict[str, Any] = {}
```
marvinpfoertner, line 47: How about we make this private and offer a setter and a getter function for each cache entry.
That way, we can ensure that the cache is only ever be written once per iteration (in the getter).
Also, we can prevent modification of previous cache entries.

marvinpfoertner (changes requested):

tests/test_utils/test_linalg/test_inner_product.py
```diff
@@ -0,0 +1,32 @@
...
+
+@pytest.fixture(scope="module")
+def vector1(n: int) -> np.ndarray:
+    rng = np.random.default_rng(567 + n)
+    return rng.standard_normal(size=(n,))
+
+
+def test_euclidean_inprod(vector0: np.ndarray, vector1: np.ndarray):
+    assert vector0 @ vector1 == pytest.approx(euclidean_inprod(v=vector0, w=vector1))
+
```
treid5, line 29: It may be useful to test that `euclidean_inprod` correctly computes the inner products of multiple vectors.

I think we could add a test like this:
`@pytest.fixture(scope = "module", params = [1, 3, 5])
def m(request) -> int:
    """ Number of simultaneous vectors """
    return request.param

@pytest.fixture(scope = "module")
def matrix0(n: int, m: int) -> np.ndarray:
    rng = np.random.default_rng()
    return rng.standard_normal(size=(n,m))

@pytest.fixture(scope = "module")
def matrix1(n: int, m: int) -> np.ndarray:
    rng = np.random.default_rng()
    return rng.standard_normal(size=(n,m))

def test_euclidean_inprod_multiple(matrix0: np.ndarray, matrix1: np.ndarray):
    m = matrix0.shape[1]
    inner_list = np.zeros(m)
    for i in range(m):
        inner_list[i] = np.inner(matrix0[i],matrix1[i])
    assert inner_list == pytest.approx(euclidean_inprod(v=matrix0,w=matrix1))
`

tests/test_utils/test_linalg/test_orthogonalize.py
```diff
@@ -0,0 +1,79 @@
...
+    ortho_vector = orthogonalization_fn(
+        v=vector, orthogonal_basis=orthogonal_basis, inprod=inprod
+    )
+    np.testing.assert_allclose(
+        orthogonal_basis @ ortho_vector,
+        np.zeros((basis_size,)),
+        atol=1e-12,
+        rtol=1e-12,
+    )
+
```
treid5, line 58: `gram_schmidt` and `modified_gram_schmidt` may fail this test even if the methods are implemented correctly. This is because the accuracy of both methods depends on the condition number of the matrix being orthogonalized. However, the accuracy of `double_gram_schmidt` does not depend on the condition number of the matrix, and if it passes, this should indicate that `gram_schmidt` and `modified_gram_schmidt` are implemented correctly

src/probnum/utils/linalg/_inner_product.py
treid5: That sounds good to me

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
treid5: It is correct on the first step because the first action is the negative residual. In CG, the action would typically be updated as 
```
beta = (np.linalg.norm(residual) / np.linalg.norm(prev_residual)) ** 2
action = -residual + beta * solver_state.actions[solver_state.step - 1]
```

Even though it is different here, it may not cause problems because the belief mean is updated as
```
cov_xy = solver_state.belief.x.cov @ action_A.T
gram = action_A @ cov_xy + self._noise_var
gram_pinv = 1.0 / gram if gram > 0.0 else 0.0
gain = cov_xy * gram_pinv
mean=solver_state.belief.x.mean + gain * proj_resid,
``` 
instead of
```
cov_xy = ORIGINAL_COVARIANCE @ action_A.T
gram = action_A @ cov_xy + self._noise_var
gram_pinv = 1.0 / gram if gram > 0.0 else 0.0
gain = cov_xy * gram_pinv
mean=solver_state.belief.x.mean + gain * proj_resid,
``` 
Using the most recent covariance acts like a reorthogonalization, so it may make both versions of the action update act the same. I am not sure about this so I will think about it more.

treid5 (commented):
I left some comments in the code. Here is a summary:

- [ ] The step that generates the next action in the conjugate gradient policy is implemented differently than in CG. I am not sure if this will cause an error
- [ ] We may want to rename the inner product and norm because "euclidean" typically only refers to the 2-norm and the inner product that induces it.
- [ ] The step that computes the orthogonalized vector in gram_schmidt can be vectorized
- [ ] I suggested a test for the inner product functions that tests whether they correctly compute multiple inner products simultaneously
- [ ] gram_schmidt and modified_gram_schmidt may fail their orthogonality test even if they are implemented correctly. However, double_gram_schmidt will probably not fail and if it passes, then gram_schmidt and modified_gram_schmidt are probably implemented correctly.

- [ ] I also suggest we add an option to `double_gram_schmidt` that chooses whether to use classical or modified Gram Schmidt. We could add the argument `classical: bool = False` and modify lines 160-165 to be this:
```
if classical:
    gs_method = lambda w: gram_schmidt(v=w, orthogonal_basis=orthogonal_basis, inprod=inprod, normalize=normalize)
else:
    gs_method = lambda w: modified_gram_schmidt(v=w, orthogonal_basis=orthogonal_basis, inprod=inprod, normalize=normalize)

v_orth = gs_method(v)
return gs_method(v_orth)
```

JonathanWenger:
> I left some comments in the code. Here is a summary: ...

Thanks for the detailed review! I'll make sure to work these in asap.

src/probnum/utils/linalg/_orthogonalize.py
JonathanWenger: I initially tried to vectorize here, but quickly ran into usability issues. If we want to vectorize here easily, the user needs to specify an inner product, that behaves such that pairwise inner products are computed and not all pairs (like `np.inner` does). So here for now I would vote for ease of use over speed to avoid premature optimization.

src/probnum/linalg/solvers/_state.py
JonathanWenger: Regarding all the feedback points for the state, I opened a new issue #622 which collects them. They are all tightly coupled with the design of the linear solver state and will likely generate a large diff unrelated to this PR. Therefore I would like to keep them separate.

src/probnum/linalg/solvers/policies/_conjugate_gradient.py
JonathanWenger: Good catch @treid5 . It does compute the update step as `-residual + beta * beta * solver_state.actions[solver_state.step - 1]` since in line 48 `action` is set to be the current residual (as to mirror gradient descent). However, now that i've added the reorthogonalization this residual currently is actually *not* reorthogonalized, while the ones used for the conjugacy correction are. I'll fix that.

JonathanWenger: Good catch @treid5 . It does compute the update step as -residual + beta * beta * solver_state.actions[solver_state.step - 1] since in line 48 action is set to be the current residual in every iteration (as to mirror gradient descent). However, now that i've added the reorthogonalization this residual currently is actually not reorthogonalized, while the ones used for the conjugacy correction are. I'll fix that.

JonathanWenger:
Thank you so much for the detailed feedback and apologies for the delay. These comments were really helpful. Here's what I changed:
- made `inner_product` more similar to `np.inner`. It handles n-d arrays now and has the same broadcasting behavior.
- introduced an option to switch the Gram-Schmidt process for the double Gram-Schmidt process
- only test double Gram-Schmidt explicitly to avoid numerical issues with the condition number of the basis matrix
- fixed a bug in CG search directions where the residual was not reorthogonalized

Any feedback regarding the state of the linear solver I collected in #622 to be implemented in a separate PR solely focussed on state design.

src/probnum/utils/linalg/_inner_product.py
```diff
@@ -1,65 +1,76 @@
...
+        Second array.
     A
         Symmetric positive (semi-)definite matrix defining the geometry.
 
     Returns
     -------
-    inprod
-        Inner product.
+    inprod :
+        *shape=(\*v.shape[:-1], \*w.shape[:-1])* -- Inner product of ``v`` and ``w``. If they are both 1-D arrays then a scalar is returned; otherwise an array is returned.
```
marvinpfoertner, line 34: Personally, I find the `np.dot`-style broadcasting extremely difficult to work with, since it practically forces you to compute all pairwise inner products. Besides, it is incompatible with the 'stack of matrices'-style broadcasting used in `__matmul__` (thus also in `pn.linops`) and in `np.linalg`. So I would propose implementing these functions here to realize the einsum expression `"...i,...i"`, even though this deviates from `np.inner`. Note that this style of broadcasting is actually more powerful than the `np.dot`-style broadcasting, since
```python
a = ... # shape: (L, M, N)
b = ... # shape: (J, K, N)
np.inner(a, b) == np.einsum("...i,...i", a[:, :, None, None, :], b[None, None, :, :, :]) # shape: (L, M, J, K)
```
Also, this way the `inner_product` implementation would compatible with its `induced_norm` broadcasting-wise.
Sorry for the strong opinion, I just really hate the broadcasting rules in `np.inner`/`np.dot`/`np.tensordot` 😀

marvinpfoertner (changes requested):
:+1:

I believe the broadcasting for `gram_schmidt` @treid5 proposed is still missing.

marvinpfoertner (approved):
👍🏻

PR #585, opened 2021-12-06
Bq test problems

fxbriol:
# In a Nutshell
This pull request includes a number of test functions to benchmark numerical integration methods. 

This includes test functions for integration against the Lebesgue measure on [0,1]^d, test functions for integration against multivariate Gaussian on R^d, and a function which transforms test functions for integration against Lebesgue measure on [0,1]^d to integrands against multivariate Gaussians on R^d


fxbriol:
@JonathanWenger @marvinpfoertner

fxbriol:
Also adding @mmahsereci since you were part of the discussion in the other pull request.

src/probnum/problems/zoo/quad/_quadproblems_uniform.py
```diff
@@ -0,0 +1,802 @@
...
+    # Check that the input points have valid values
+    assert np.all(x <= 1.0) and np.all(x >= 0.0)
+
+    # Reshape u into an (n,dim) array with identical rows
+    u = np.repeat(u.reshape([1, dim]), n, axis=0)
+
+    # Compute function values
+    f = np.exp(-np.sum(a * np.abs(x - u), axis=1))
+
+    return f.reshape((n, 1))
```
marvinpfoertner, line 60:
```suggestion
def genz_continuous(
    dim: int, a: np.ndarray = None, u: np.ndarray = None
) -> QuadratureProblem:
    """Genz 'continuous' test function on [0,1]^d.
    .. math::  f(x) = \exp(- \sum_{i=1}^d a_i |x_i - u_i|).
    Parameters
    ----------
        x
            Array of points at which to evaluate the function.
            Each row is a point and column corresponds to dimensions.
            All entries should be in [0,1].
        a
            First set of parameters affecting the difficulty of the integration problem.
        u
            Second set of parameters affecting the difficulty of the integration problem.
            All entries should be in [0,1].
    Returns
    -------
        f
            array of function evaluations at points 'x'.
    References
    ----------
        https://www.sfu.ca/~ssurjano/cont.html
    """

    # Specify default values of parameters a and u
    if a is None:
        a = np.broadcast_to(5.0, dim)

    if u is None:
        u = np.broadcast_to(0.5, dim)

    # Check that the parameters have valid values and shape
    if a.shape != (dim,):
        raise ValueError(
            f"Invalid shape {a.shape} for parameter `a`. Expected {(dim,)}."
        )

    if u.shape != (dim,):
        raise ValueError(
            f"Invalid shape {u.shape} for parameter `u`. Expected {(dim,)}."
        )

    if np.any(u < 0.0) or np.any(u > 1):
        raise ValueError(f"The parameters `u` must lie in the interval [0.0, 1.0].")

    def integrand(x: np.ndarray) -> np.ndarray:
        n = x.shape[0]

        # Check that the input points have valid values and shape
        if x.shape != (n, dim):
            raise ValueError(
                f"Invalid shape {x.shape} for input points `x`. Expected (n, {dim})."
            )

        if np.any(x < 0.0) or np.any(x > 1):
            raise ValueError(f"The input points `x` must lie in the box [0.0, 1.0]^d.")

        # Reshape u into an (n,dim) array with identical rows
        u = np.repeat(u.reshape([1, dim]), n, axis=0)

        # Compute function values
        f = np.exp(-np.sum(a * np.abs(x - u), axis=1))

        return f.reshape((n, 1))

    solution = np.prod((2.0 - np.exp(-a * u) - np.exp(a * (u - 1))) / a)

    return QuadratureProblem(
        integrand=integrand,
        lower_bd=np.broadcast_to(0.0, dim),
        upper_bd=np.broadcast_to(1.0, dim),
        output_dim=None,
        solution=solution,
    )
```

src/probnum/problems/zoo/quad/__init__.py
```diff
@@ -0,0 +1 @@
+"""Collection of test problems for numerical integration methods."""
```
marvinpfoertner, line 1:
```suggestion
"""Collection of test problems for numerical integration methods."""

from ._quadproblems_gaussian import *
from ._quadproblems_uniform import *
```

src/probnum/problems/zoo/quad/_quadproblems_uniform.py
```diff
@@ -0,0 +1,802 @@
...
+
+    (n, dim) = x.shape
+
+    # Check that the input points have valid values
+    assert np.all(x <= 1.0) and np.all(x >= 0.0)
+
+    # Compute function values
+    f = np.zeros(n)
+    for i in range(1, dim + 1):
+        f = f + ((-1.0) ** i) * np.prod(x[:, range(i)], axis=1)
```
marvinpfoertner, line 361: This should be faster:

```suggestion
    f = np.sum(
        ((-1.0) ** np.arange(1, dim + 1)) * np.cumprod(x, axis=1),
        axis=1,
    )
```

src/probnum/problems/zoo/quad/_quadproblems_gaussian.py
```diff
@@ -0,0 +1,95 @@
+import numpy as np
+from scipy import special
+
+from probnum.typing import FloatArgType
+
+
```
marvinpfoertner, line 6:
```suggestion
__all__ = [
    "sum_polynomials",
]


```

src/probnum/problems/zoo/quad/_quadproblems_uniform.py
```diff
@@ -0,0 +1,802 @@
+import itertools
+
+import numpy as np
+from scipy.stats import norm
+
+from probnum.typing import FloatArgType, IntArgType
+
+
```
marvinpfoertner, line 8:
```suggestion
__all__ = [
    "genz_continuous",
    "genz_cornerpeak",
    "genz_discontinuous",
    "genz_gaussian",
    "genz_oscillatory",
    "genz_productpeak",
    "bratley1992",
    "roos_arnold",
    "g_function",
    "morokoff_caflisch_1",
    "morokoff_caflisch_2",
]
```

These are also the names I would suggest for the problem constructors

marvinpfoertner (changes requested):
Very nice 👍🏻  I like the exhaustive docstrings!

A few general comments:
- [ ] I would suggest providing `QuadratureProblem` "constructors" instead of the individual integrand and integral functions. This makes it easier for the user to use the problems later on. I provided a proposal implementation for the `genz_continuous` problem.
- [ ] Input checking should not be done via assertions, since these are just meant for "sanity checking" and can be turned of in the interpreter configuration. User-facing functions should instead throw a meaningful exception describing the problem.
- [ ] function names should generally be `snake_case` and lowercase
- [ ] The `.DS_Store` files need to be removed. These are macOS system files which should never be checked into a git repo. You can generally ignore these files by adding a global gitignore file in your home directory

mmahsereci (commented):
hey @fxbriol , this is great! very glad that the BQ package is getting some test problems finally.

I don't want to overwelm this PR with inline comments as @marvinpfoertner has commented quite a lot already, but here are some general comments:

- Splitting PRs: It is generally easier to split PRs and make them as small as possible. For example here you could make one PR that adds a single function to get feedback on the structure of the code mostly. Then, in a second PR you could add all the other problems which is then more straightforward as you simply need to replicate the existing structure. This is generally easier and less time consuming as you do not need to re-structure all existing code in one big PR. This PR is not overly large, but you may still save some time for yourself in the future if you split PRs :) 
- Testing integrals: It's not easy to test integrals. And I am actually not sure what a robust way is here. But we should probably avoid randomness if possible. Hence, perhaps use a random number generator with  fixed seed to produce the MC samples, so we don't get flaky tests. Further, it might be neat to check if the value is covered by some confidence interval of the MC estimator rather than having a fixed relative tol of 1e-3.
- Adding shapes to documentation: Expected shapes of all in and output arrays should be added to the docstrings.
- Compatibility: Have you tried if the functions you implemented are compatible with the `bayesquad` interface? We should aim to be able to use the test functions out of the box.

fxbriol:
Dear @marvinpfoertner , @mmahsereci : thanks both for the suggestions - they are really appreciated as I am very much a beginner when it comes to this. I'll have a look at your comments early next week and do my best to address them!

F-X

fxbriol:
> Very nice 👍🏻 I like the exhaustive docstrings!
> 
> A few general comments:
> 
> * [ ]  I would suggest providing `QuadratureProblem` "constructors" instead of the individual integrand and integral functions. This makes it easier for the user to use the problems later on. I provided a proposal implementation for the `genz_continuous` problem.
> * [ ]  Input checking should not be done via assertions, since these are just meant for "sanity checking" and can be turned of in the interpreter configuration. User-facing functions should instead throw a meaningful exception describing the problem.
> * [ ]  function names should generally be `snake_case` and lowercase
> * [ ]  The `.DS_Store` files need to be removed. These are macOS system files which should never be checked into a git repo. You can generally ignore these files by adding a global gitignore file in your home directory

All done! Thanks again for those comments, very helpful.

fxbriol:
Hi @marvinpfoertner , @mmahsereci : Thanks again for your comments. I think I have now managed to implement all of them (I will certainly keep the one about keeping pull requests short in mind for the future!). I am still struggling to pass all of the checks, and was wondering if you could give me a hand with this?

Also to answer to Maren's point about Monte Carlo tests more directly: I have now added seeds. The integrands all also work with bayesquad, but I have found the bayesquad function not be very reliable which is why I do the tests with MC. I think the main reason is that we do not estimate hyper parameters, and as a result BQ is over confident and the stopping rule becomes inaccurate.

mmahsereci:
> Also to answer to Maren's point about Monte Carlo tests more directly: I have now added seeds. The integrands all also work with bayesquad, but I have found the bayesquad function not be very reliable which is why I do the tests with MC. I think the main reason is that we do not estimate hyper parameters, and as a result BQ is over confident and the stopping rule becomes inaccurate.

Sure that makes sense. What I was referring to initially was to make sure that no errors are thrown, e.g., shape errors, when the Callable you define (the test function) is used with the `bayesquad` interface. It's just a manual check if all components work/run well together. Ideally these checks are done automatically, and are called end-to-end [integration tests](https://en.wikipedia.org/wiki/Integration_testing) (the word 'integration' has a different meaning here), but we don't have them yet in probnum (at least for `quad` package), so it's good to do a quick manual check. The kernel embedding values (if the integral is correct) are tested separately already. The BQ code should not be used to check integrals as you say :)

You can also have a look [here](https://github.com/probabilistic-numerics/probnum/blob/main/tests/test_quad/util.py) and [here](https://github.com/probabilistic-numerics/probnum/blob/main/tests/test_quad/test_kernel_embeddings.py#L33) where Gauss-Hermite tensor product rule is used to test integrals. May be an alternative to MC sampling and we could re-use the existing code.

@marvinpfoertner If the `QuadratureProblem` needs some re-design in this PR, that should be OK. It was defined a bit pre-maturely before the `bayesquad` interface or any test problem in fact was defined. For instance, it might be easier to use `domain` attribute instead of bounds. But we can also change that at a later point.

marvinpfoertner:
Hi @fxbriol :) What is the status of your PR? Should we maybe set up a meeting soon to discuss how we finalize this?

marvinpfoertner (approved):
Thanks again for the PR :) Feel free to merge once the test coverage for the `_quadproblems_gaussian.py` is sufficiently high.

mmahsereci:
Closing due to inactivity. Feel free to re-open in case you want to finish the PR at some point @fxbriol .

PR #588, opened 2021-12-09
Row selection policy for the PLS

JonathanWenger:
# In a Nutshell
Expands the functionality of the random row selection / unit vector policy of the probabilistic linear solver.

# Detailed Description
Adds sampling without replacement and a row-norm dependent distribution over the rows. This policy is used for example in the [randomized Kaczmarz algorithm](https://en.wikipedia.org/wiki/Kaczmarz_method#Algorithm_2:_Randomized_Kaczmarz_algorithm).

# Related Issues and PRs
This PR depends on #580 and should be reviewed after.


src/probnum/linalg/solvers/policies/_random_unit_vector.py
```diff
@@ -9,16 +9,59 @@
...
-        action = np.zeros(n)
+        nrows = solver_state.problem.A.shape[0]
+
+        # Initial sampling probabilities
+        if not "row_sample_probs" in solver_state.cache.keys():
+            if self.probabilities == "uniform":
+                solver_state.cache["row_sample_probs"] = np.ones(nrows) / nrows
+            elif self.probabilities == "rownorm":
+                solver_state.cache["row_sample_probs"] = np.sum(
+                    solver_state.problem.A * solver_state.problem.A, axis=1
```
marvinpfoertner, line 43: Why not
```suggestion
                    solver_state.problem.A**2, axis=1
```
?

marvinpfoertner (approved):

PR #589, opened 2021-12-09
Add CODEOWNERS

marvinpfoertner:
# In a Nutshell
This PR adds a proposal for a CODEOWNERS file.


marvinpfoertner:
@JonathanWenger I propose that we merge this, unless there is any obvious problems with the current assignments. Then we can discuss the assignment in person in the next software meeting.

JonathanWenger:
Before we require reviews from code owners i propose we discuss whether secondary code owners are sensible for redundancy and fast turnarounds.

JonathanWenger (approved):

PR #581, opened 2021-11-25
[Draft] Compute Backends and Automatic Differentiation

marvinpfoertner:
# In a Nutshell
This PR introduces compute backends and automatic differentiation in ProbNum.

# Detailed Description
- Currently supported backends
  - NumPy
  - Jax
  - PyTorch

# Related Issues
None

PR #600, opened 2022-01-01
Refactor `as_numpy_scalar` to produce scalar `np.ndarray`s

marvinpfoertner:
# In a Nutshell
This PR refactors the `probnum.utils.as_numpy_scalar` to return scalar `np.ndarray`s instead of `np.number`s. This is preparation for PR #581, since JAX and PyTorch don't have array scalars.

### DIFF
--- a/tests/test_utils/test_argutils.py
+++ b/tests/test_utils/test_argutils.py
@@ -7,7 +7,7 @@ import probnum.utils as pnut
 
 
 @pytest.mark.parametrize("scalar", [1, 1.0, 1.0 + 2.0j, np.array(1.0)])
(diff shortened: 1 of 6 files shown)


JonathanWenger (approved):

tests/test_utils/test_argutils.py
```diff
@@ -10,7 +10,7 @@
 def test_as_numpy_scalar_scalar_is_good(scalar):
```
JonathanWenger, line 10:
```suggestion
def test_as_numpy_scalar_returns_scalar_array(scalar):
```

PR #606, opened 2022-01-11
Fix pylint errors for `_perturbsolve_ivp.py`

pnkraemer:
# In a Nutshell
Fix pylint errors for `_perturbsolve_ivp.py`

### DIFF
--- a/src/probnum/diffeq/_perturbsolve_ivp.py
+++ b/src/probnum/diffeq/_perturbsolve_ivp.py
@@ -63,8 +63,8 @@ def perturbsolve_ivp(
         Random number generator.
     method
         Integration method to use.
-        The following are available
-        (docs adapted from `SciPy <https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html>`_):  # pylint: disable="line-too-long"
+        The following are available (docs adapted from
+        `SciPy <https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html>`_):
 
             * `RK45` (default): Explicit Runge-Kutta method of order 5(4) [2]_.
               The error is controlled assuming accuracy of the fourth-order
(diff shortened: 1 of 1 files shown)


src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -14,45 +17,54 @@
...
         Initial value.
-    rng :
+    rng
         Random number generator.
-    method :
+    method
         Integration method to use.
-        The following are available (docs adapted from `SciPy <https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html>`_):
+        The following are available
+        (docs adapted from `SciPy <https://docs.scipy.org/doc/scipy/reference/generated/scipy.integrate.solve_ivp.html>`_):  # pylint: disable="line-too-long"
```
pnkraemer, line 67: this disable message here is a bit annoying, because it is rendered in the docs. I will try to fix it somehow.

src/probnum/diffeq/_perturbsolve_ivp.py
```diff
@@ -155,22 +196,50 @@ def perturbsolve_ivp(
...
         Some Practical Runge-Kutta Formulas.
         Mathematics of Computation, Vol. 46, No. 173, pp. 135-150, 1986.
     .. [4] P. Bogacki, L.F. Shampine.
         A 3(2) Pair of Runge-Kutta Formulas.
         Appl. Math. Lett. Vol. 2, No. 4. pp. 321-325, 1989.
-    """
+    """  # pylint: disable="line-too-long"
+    # The line-too-long-disable is for
+    # the rather long URL that points to the original SciPy docs,
+    # and which cannot be broken up easily.
```
pnkraemer, line 216: I guess practicality beats purity? If better solutions are obvious to someone reading this, please feel invited to comment.

marvinpfoertner: It should be possible to insert a line break using a backslash like so:

```python
def func():
    """https://very.very/very/very/\
very/very/long/url

    Normal indentation.
    """
```

Note the missing indentation on the second line of the docstring

pnkraemer: Thanks!

marvinpfoertner: Sorry, I just noticed that this is an r-string. It will work if you make this a regular string literal, since the backslash escapes the line break in this case

nathanaelbosch (approved):

PR #607, opened 2022-01-11
Fix pylint errors for `_probsolve_ivp.py`

pnkraemer:
# In a Nutshell
Fix pylint errors for `_probsolve_ivp.py`.

### DIFF
--- a/src/probnum/diffeq/_utils.py
+++ b/src/probnum/diffeq/_utils.py
@@ -3,7 +3,7 @@
 from probnum.diffeq import stepsize
 
 
-def construct_steprule(ivp, adaptive, step, atol, rtol):
+def construct_steprule(*, ivp, adaptive, step, atol, rtol):
     if adaptive is True:
         if atol is None or rtol is None:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


src/probnum/diffeq/_utils.py
```diff
@@ -0,0 +1,18 @@
...
+            raise ValueError(
+                "Please provide absolute and relative tolerance for adaptive steps."
+            )
+        firststep = step if step is not None else stepsize.propose_firststep(ivp)
+        steprule = stepsize.AdaptiveSteps(firststep=firststep, atol=atol, rtol=rtol)
+    else:
+        if step is None:
+            raise ValueError("Constant steps require a 'step' argument.")
+        steprule = stepsize.ConstantSteps(step)
+    return steprule
```
pnkraemer, line 18: we can also use this function for `perturbsolve_ivp`, which is why I pulled it out of `_probsolve_ivp.py`, but let's not do this yet to keep the PRs separate.

nathanaelbosch (approved):

PR #609, opened 2022-01-11
Fix pylint errors for `_odesolution.py` and `_odesolver.py`

pnkraemer:
# In a Nutshell
Fix pylint errors for `_odesolution.py` and `_odesolver.py`.


src/probnum/diffeq/_odesolver.py
```diff
@@ -158,25 +168,22 @@ def method_callback(self, state):
...
-                self._next_location = np.inf
+                next_location = np.inf
 
-        if t + dt > self._next_location:
-            dt = self._next_location - t
+        if t + dt > next_location:
+            dt = next_location - t
         return dt
+
+    return adjust
```
pnkraemer, line 189: I tried implementing this in a single function with partial(), but it did not work somehow. (At least not straightforwardly). But i think the present solution is also fine.

tests/test_diffeq/test_odesolver.py
```diff
@@ -62,26 +62,5 @@ def test_solve(self):
...
-    def test_adjust_dt_to_time_stops(self):
-        # Should interfere dt to 0.1 instead of 5.0, because 2 is in self.time_stops
-        dt = self.discrete_events.adjust_dt_to_time_stops(t=1.9, dt=5.0)
-        assert dt == pytest.approx(0.1)
-
-        # Should not interfere dt if there is no proximity to an event
-        dt = self.discrete_events.adjust_dt_to_time_stops(t=1.0, dt=0.00001)
-        assert dt == pytest.approx(0.00001)
-
-
```
pnkraemer, line 85: theses tests test the behaviour of a private method (which is suboptimal), and are redundant. Whenever I break the implementation of the time stop adjustment, other tests fail, too.

tests/test_diffeq/test_odesolver.py
```diff
@@ -1,87 +0,0 @@
-import unittest
-
```
pnkraemer, line 2: the whole test suite is quite redundant. It does not test anything that is not covered by the meaningful tests, for example those for the ODE filter.

pnkraemer: (this test suite has been a thorn in my side for a while now, because for every refactoring in the ODESolver interface, I needed to meaninglessly change the tests here.)

marvinpfoertner (approved):

PR #613, opened 2022-01-17
Filter warning for importlib

pnkraemer:
# In a Nutshell
At the moment, running the tests yields the following warning
```
../pnenv/lib/python3.8/site-packages/flatbuffers/compat.py:19
  /home/kraemer/Programmes/probnum/pnenv/lib/python3.8/site-packages/flatbuffers/compat.py:19: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp

-- Docs: https://docs.pytest.org/en/stable/warnings.html
```
This is not an issue in probnum, but in jax. (See https://github.com/google/jax/blob/main/pytest.ini).
This PR introduces a warning filter to pyproject.toml, which makes pytest ignore this specific warning.


JonathanWenger:
I'm not sure we should categorically ignore this warning, since it would prevent us from noticing if someone introduced an 'imp' dependency. Maybe we just accept that warning as coming from one of the backends and wait it out until JAX fixes it?

pnkraemer:
Yes, we can also do it this way. I am not super set on merging this PR, but the warning is annoying me a little bit to be honest.

I am wondering whether a warning in the tests would be the place to catch such an introduced dependency though...

JonathanWenger:
It seems like you can filter warnings for specific modules (see https://stackoverflow.com/a/53218641). I think this would be the way to do it by simply ignoring the deprecation warning only for JAX.

PR #614, opened 2022-01-17
Improve ODE filter initialisation

pnkraemer:
# In a Nutshell
This PR improves the ODE filter initialisation.

# Detailed Description
Now, the ODE filter actually works with Taylor-mode. There are also a bunch more initialisation routines, and improved tests thereof.


src/probnum/diffeq/odefilter/information_operators/_ode_residual.py
```diff
@@ -66,10 +66,10 @@ def jacobian(self, t: FloatLike, x: np.ndarray) -> np.ndarray:
 
     def _residual_first_order_ode(self, t: FloatLike, x: np.ndarray) -> np.ndarray:
         h0, h1 = self.projection_matrices
-        return h1 @ x - self.ode.f(t, h0 @ x)
+        return h1 @ x - np.asarray(self.ode.f(t, h0 @ x))
```
pnkraemer, line 69: required for jax-IVPs (Normal() does not take jax arrays)

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
```diff
@@ -0,0 +1,247 @@
...
+
+from ._interface import InitializationRoutine
+
+
+class _AutoDiffBase(InitializationRoutine):
+    def __init__(self):
+        super().__init__(is_exact=True, requires_jax=True)
+        self._jax, self._jnp = self._import_jax()
+
+    def _import_jax(self):
```
pnkraemer, line 15: these kinds of imports are my attempt at handling the optional JAX dependency cleanly. Once the backend-PR is through, the Jax-handling will be much cleaner anyway.

src/probnum/diffeq/odefilter/utils/_problem_utils.py
```diff
@@ -81,8 +81,11 @@ def _construct_measurement_models(
 ):
     """Construct measurement models for the IVP."""
 
-    transition_matrix = np.eye(
-        ode_information_operator.output_dim, ode_information_operator.input_dim
+    shape = ode_information_operator.output_dim, ode_information_operator.input_dim
+
+    # Make a proj2coord(0)-like matrix without relying on the prior_process
+    transition_matrix = (
+        np.eye(*shape).reshape((-1,), order="C").reshape(shape, order="F")
```
pnkraemer, line 88: the previous version was incorrect.

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
schmidtjonathan: Why don't we move the try-import-except block top-level and -- in AutoDiffBase -- only check whether jax could be imported? I think the solution you propose here is problematic.

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
```diff
@@ -0,0 +1,247 @@
...
+        except ImportError as err:
+            raise ImportError(self._import_jax_error_msg) from err
+
+    @property
+    def _import_jax_error_msg(self):
+        return (
+            "Cannot perform Jax-based initialization without the optional "
+            "dependencies jax and jaxlib. "
+            "Try installing them via `pip install jax jaxlib`."
+        )
```
schmidtjonathan, line 36: What is the argument for making this a property? I find that odd (from an OOP perspective)

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def _make_autonomous(self, *, ivp):
+        y0_autonomous = self._jnp.concatenate([ivp.y0, self._jnp.array([ivp.t0])])
+
+        def f_autonomous(y):
+            x, t = y[:-1], y[-1]
+            fx = ivp.f(t, x)
+            return self._jnp.concatenate([fx, self._jnp.array([1.0])])
+
+        return f_autonomous, y0_autonomous
```
schmidtjonathan, line 75: One might argue that this functionality should be implemented in the `IVP` class, since it is not the job of the initializer to re-formalize initial value problems

src/probnum/diffeq/odefilter/init_routines/_non_probabilistic_fit.py
```diff
@@ -0,0 +1,132 @@
+"""Runge-Kutta-based initialization routines."""
```
schmidtjonathan, line 1: Is it still Runge-Kutta based?

schmidtjonathan (commented):
I mainly focused on code quality / design, and I am sceptical in this regard, even though I get the challenges that this Jax-backend-transition phase bears.
Still, I left some points to discuss.

Additional points:
- Can you elaborate what exactly was "improve"d (functionality-wise) in this PR?
- minor comment: I do not get the change in naming ("initialization" -> "init") and I would not have done it, but that's just my opinion.

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
pnkraemer: If I remember correctly, there import error would be raised too early, i.e., when setting up the `init_routines` namespace. In this case, none of the routines could be used. 

I can check again later.

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
pnkraemer: it is just a string, making this a function feels odd. I thought about making this a class variable, but this felt weird, too. A property that is stored right next to the import function felt like the cleanest solution

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
pnkraemer: Hmmmm maybe... it is simply to simplify the calls to autodiff routines. Since this functionality can be isolated so strictly to this use case, isnt it easiest to parse if implemented here? What do you think?

pnkraemer:
Thanks for your review! 

> I mainly focused on code quality / design, and I am sceptical in this regard, even though I get the challenges that this Jax-backend-transition phase bears. Still, I left some points to discuss.


If you have a better solution than what is used here, please let me know. I am not 100% happy with that myself, bound did not find a better solution. (Hopefully, this discussion will be superfluous soon anyway.

>     * Can you elaborate what exactly was "improve"d (functionality-wise) in this PR?

I edited the description of this PR 

>     * minor comment: I do not get the change in naming ("initialization" -> "init") and I would not have done it, but that's just my opinion.

The long version was so long that it was a real pain to use. `initialization_routine` has 22 letters, and that is just for the namespace... :( 

We can have a brief offline discussion later, if you like. I think this is easier than discussing here :)

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
schmidtjonathan: Why does it have to belong to the object? Why not just define the string directly at the point where it is raised?
And if we need it multiple times and you don't want to paste it everywhere, we could define it as a global somewhere? I don't know I find that everything else feels kinda weird.

schmidtjonathan:
> If you have a better solution than what is used here, please let me know. I am not 100% happy with that myself, bound did not find a better solution. (Hopefully, this discussion will be superfluous soon anyway.

I think we would have to figure this out offline (together?) 

> I edited the description of this PR

Ok thank you very much, this makes it clearer. I wondered how the _existing_ functionality has been improved and figured that it is mainly an extension instead of an improvement.

> The long version was so long [...]

Mh, I see. Even though I am always in favor of the more expressive version -- even if it is longer -- I think that `init = initializ{ation, e, er, ...}` is clear

src/probnum/diffeq/odefilter/init_routines/_autodiff.py
pnkraemer: There are no changes made to the IVP. ``f`` and ``y0`` are wrapped into a format that is more convenient for the autodiff init-routine. It is purely internal, and at the moment, I cannot see any potential to reuse this functionality anywhere else. If such potential comes up, we can move it to a different place, but at the moment, I think it is well-placed here for this reason.
What do you think? :) 

In any case, I take from this the need to leave a little docstring in this function that explains precisely that rationale. (That is part of the most recent commit).

PR #615, opened 2022-01-19
Process noise in discrete transitions

pnkraemer:
# In a Nutshell
This PR reformulates the "noise" in discrete transitions in terms of an "additive random variable". 

# Detailed Description
More specifically, instead of transitions in the shape of
```
y ~ N(g(x), S)
```
the new formulation describes them as 
```
y = g(x) + v, v ~ N(m, C)
```
This is interesting, because it encapsulates shift, cov, and cov_cholesky in a single argument: `noise`. (The term "process-noise" does not seem appropriate for its use in observation models, for example.)  More specifically, the handling of cov and cov cholesky is outsourced entirely. This makes the transitions more maintainable and the code more readable. 

As a side quest, this PR
* Renames state_trans_* into transition_matrix, transition_fun, etc., which is cleaner
* Makes the discrete transitions kw-only


schmidtjonathan (approved):
Much cleaner, much more concise, thanks a lot! :)

PR #616, opened 2022-01-20
Additions to the global pylint configuration

marvinpfoertner:
# In a Nutshell
I added some new global pylint ignores and configuration items from the corresponding discussion #470 

# Detailed Description
- "missing-return-doc" as a temporary global ignore
- "missing-function-docstring" as a long-term ignore for the tests
- raise "max-args" and "max-locals"

# Related Issues
None


pnkraemer (approved):

mmahsereci:
We probably also need `missing-yield-doc` as a temporary global ignore. I can do a PR if you think that's meaningful @marvinpfoertner

marvinpfoertner:
> We probably also need `missing-yield-doc` as a temporary global ignore. I can do a PR if you think that's meaningful @marvinpfoertner

Damn, I forgot about that one. Yeah that would be nice, thanks :) I'll add `missing-yield-doc`to #470

Issue #571, opened 2021-11-18
Matrix Property System for `LinearOperator`s


marvinpfoertner:
**Is your feature request related to a problem? Please describe.**
Currently, a `LinearOperator` is unaware of any mathematical properties (e.g. symmetry, positive-definiteness, orthogonality) of the matrix it represents.
However, many algorithms acting on these matrices can be more efficient or in some cases only become tractable if these properties are known.

**Give an example use case.**
- Inversion of a positive-definite `LinearOperator` should be computed by Cholesky factorization
- Inversion of an orthogonal `LinearOperator` should be computed by multiplying with the transpose 

**Describe the solution you'd like.**
- add cached properties to the `LinearOperator` which use ternary logic to indicate whether the matrix has a particular property
  - return `True` if the matrix has the property
  - return `False` if the matrix does not have the property
  - return `None` if it is not known whether the matrix has the property
- minimal set of properties
  - `is_invertible`
  - `is_symmetric`
  - `is_positive_{semi}definite`
  - `is_{upper,lower}_triangular`

**Additional context**
Some of these properties are needed for #569

Issue #569, opened 2021-11-16
`LinearOperator.cholesky`


marvinpfoertner:
**Is your feature request related to a problem? Please describe.**
Currently, the implementation of `Normal` contains a lot of logic concerning the efficient computation of Cholesky factorizations of `LinearOperator`s with `Kronecker` and `SymmetricKronecker` structure.
This adds unnecessary complexity and is out of place in the implementation of a `RandomVariable`.
We should add a `cholesky` method to the `LinearOperator`, which then takes care of the specific factorizations in its subclasses.

**Give an example use case.**
```
>>> A = pn.linops.Kronecker(A=..., B=...)
>>> L = A.cholesky(lower=True)
>>> type(L)
pn.linops.Kronecker
>>> x = A.inv() @ b  # uses Cholesky internally!
```

**Describe the solution you'd like.**
- [ ] `LinearOperator.cholesky(lower: bool = True)`

PR #618, opened 2022-01-23
Cholesky decompositions of `LinearOperator`s

marvinpfoertner:
# In a Nutshell
This PR implements Cholesky decompositions for ProbNum's `linops` library.
To achieve this goal, I also had to implement symmetrization and parts of a matrix property system.

# Detailed Description
- `LinearOperator.cholesky(lower: bool) -> LinearOperator`
- `LinearOperator.symmetrize() -> LinearOperator` implementing "1/2 * (A + A^T)"
- Matrix properties
    - `is_symmetric`
    - `is_{upper,lower}_triangular`
    - `is_positive_definite`
- Fixed bugs in `Kronecker.{inv,cond}`
- `_InverseLinearOperator` uses Cholesky factorization for SPD matrices
- Fixed bug in `_InverseLinearOperator` of singular matrix
- Added new tests and test cases

# Related Issues
Implements parts of #569 and #571

### DIFF
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -557,7 +557,7 @@ class LinearOperator:
 
         The Cholesky decomposition of a symmetric positive-definite matrix :math:`A \in
         \mathbb{R}^{n \times n}` is given by :math:`A = L L^T`, where the unique
-        Cholesky factor `L \in \mathbb{R}^{n \times n}` of :math:`A` is a lower
+        Cholesky factor :math:`L \in \mathbb{R}^{n \times n}` of :math:`A` is a lower
         triangular matrix with a positive diagonal.
 
         As a side-effect, this method will set the value of :attr:`is_positive_definite`
@@ -717,17 +717,23 @@ class LinearOperator:
         norm is given by
 
         .. math::
-            sym(A) = \\frac{1}{2} (A + A^T).
+            \\operatorname{sym}(A) := \\frac{1}{2} (A + A^T).
 
         However, for efficiency reasons, it is preferrable to approximate this operator
-        in some cases. For example, a Kronecker product :math:`K = A \\ocross B` is more
+        in some cases. For example, a Kronecker product :math:`K = A \\otimes B` is more
         efficiently symmetrized by means of
 
         .. math::
-            sym(A) \\ocross sym(B)
-            = sym(K) + \\frac{1}{2} \\left(
-                sym(K) - \frac{1}{2} \\left( A \\ocross B^T + A^T \\ocross B \\right)
-            \\right).
+            :nowrap:
+
+            \\begin{equation}
+                \\operatorname{sym}(A) \\otimes \\operatorname{sym}(B)
+                = \\operatorname{sym}(K) + \\frac{1}{2} \\left(
+                    \\frac{1}{2} \\left(
+                        A \\otimes B^T + A^T \\otimes B
+                    \\right) - \\operatorname{sym}(K)
+                \\right).
+            \\end{equation}
 
         Returns
         -------
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -546,12 +546,21 @@ class LinearOperator:
 
             self.is_positive_definite = True
 
-        if self.__cholesky_cache.is_lower_triangular == lower:
+        upper = not lower
+
+        if (lower and self.__cholesky_cache.is_lower_triangular) or (
+            upper and self.__cholesky_cache.is_upper_triangular
+        ):
             return self.__cholesky_cache
+
+        assert (
+            self.__cholesky_cache.is_lower_triangular
+            or self.__cholesky_cache.is_upper_triangular
+        )
 
         return self.__cholesky_cache.T
 
-    def _cholesky(self, lower: bool = True) -> LinearOperator:
+    def _cholesky(self, lower: bool) -> LinearOperator:
         return Matrix(
             scipy.linalg.cholesky(
                 self.todense(), lower=lower, overwrite_a=False, check_finite=True
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -544,6 +544,8 @@ class LinearOperator:
             else:
                 self.__cholesky_cache.is_upper_triangular = True
 
+            self.is_positive_definite = True
+
         if self.__cholesky_cache.is_lower_triangular == lower:
             return self.__cholesky_cache
 
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -168,6 +168,10 @@ class LinearOperator:
 
         self.__cholesky_cache = None
 
+        # Property inference
+        if not self.is_square:
+            self.is_symmetric = False
+
     @property
     def shape(self) -> Tuple[int, int]:
         """Shape of the linear operator.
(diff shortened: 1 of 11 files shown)


src/probnum/linops/_linear_operator.py
```diff
@@ -296,6 +306,98 @@ def todense(self, cache: bool = True) -> np.ndarray:
...
+                f"The matrix property `{name}` does not exist."
+            ) from err
+
+        if curr_value == value:
+            return
+
+        if curr_value is not None:
+            assert isinstance(curr_value, bool)
+
+            raise ValueError(f"Can not change the value of the matrix property {name}")
```
JonathanWenger, line 391:
```suggestion
            raise ValueError(f"Can not change the value of the matrix property {name}.")
```

src/probnum/linops/_linear_operator.py
```diff
@@ -446,6 +548,93 @@ def _trace_fallback(self) -> np.number:
...
+
+        Raises
+        ------
+        numpy.linalg.LinAlgError
+            If the :class:`LinearOperator` is not symmetric, i.e. if
+            :attr:`is_symmetric` is not set to :obj:`True`.
+        numpy.linalg.LinAlgError
+            If the :class:`LinearOperator` is not positive definite.
+        """
+        if not self.is_symmetric:
```
JonathanWenger, line 594: What happens if the `is_symmetric` and `is_positive_definite` properties are `None`, i.e. it is unknown whether the matrix is spd? According to the docstring `.cholesky()` should still run and if it succeeds, set the properties accordingly. However if `self.is_symmetric is None`, this conditional should throw an exception, if i'm not mistaken.

src/probnum/linops/_linear_operator.py
```diff
@@ -446,6 +548,93 @@ def _trace_fallback(self) -> np.number:
...
+            raise np.linalg.LinAlgError(
+                "The Cholesky decomposition is only defined for symmetric matrices."
+            )
+
+        if self.is_positive_definite is False:
```
JonathanWenger, line 599: Same as above. What if `self.is_positive_definite is None`?

src/probnum/linops/_linear_operator.py
```diff
@@ -520,6 +709,59 @@ def inv(self) -> "LinearOperator":
 
         return self.__inverse()
 
+    def symmetrize(self) -> LinearOperator:
```
JonathanWenger, line 712: By introduction of this method the linear operator `Symmetrize` in `_kronecker.py` should just be removed, I think. Otherwise there are two different ways of achieving the same thing.

src/probnum/linops/_scaling.py
```diff
@@ -304,6 +314,20 @@ def _cond_isotropic(self, p: Union[None, int, float, str]) -> np.inexact:
         else:
             return np.linalg.cond(self.todense(cache=False), p=p)
 
+    def _cholesky(self, lower: bool = True) -> Scaling:
+        if self._scalar is not None:
+            if self._scalar <= 0:
+                raise np.linalg.LinAlgError(
+                    "The linear operator is not positive definite"
```
JonathanWenger, line 321:
```suggestion
                    "The linear operator is not positive definite."
```

src/probnum/linops/_scaling.py
```diff
@@ -304,6 +314,20 @@ def _cond_isotropic(self, p: Union[None, int, float, str]) -> np.inexact:
...
+                )
+
+            return Scaling(np.sqrt(self._scalar), shape=self.shape)
+
+        if np.any(self._factors <= 0):
+            raise np.linalg.LinAlgError("The linear operator is not positive definite")
```
JonathanWenger, line 327:
```suggestion
            raise np.linalg.LinAlgError("The linear operator is not positive definite.")
```

JonathanWenger (changes requested):
It's great to finally have this property system. I have a few comments which should be easily resolved. Thanks for the effort! One question to think about for the future (not this PR) what do we do with semi-definite operators? Add another property `_is_positive_semi_definite`? Often there are algorithms which also work if the linear operator is at least positive semi-definite.

src/probnum/linops/_linear_operator.py
marvinpfoertner: `Symmetrize` in `_kronecker.py` symmetrizes vectorized matrices used in Kronecker algebra, so it operates "on the arguments" of linear operators, while `linop.symmetrize` symmetrizes the `LinearOperator` itself, so I'd say they serve distinct purposes.

src/probnum/linops/_linear_operator.py
marvinpfoertner: According to the `Raises` section of the docstring, the method raises an exception if `is_symmetric` is not `True`, i.e. the method will fail if it is not known that the operator is indeed symmetric.
Similarly I state that the operator raises an exception if the matrix is not positive definite, while still setting `linop.is_positive_definite = False`.
This is intentional as it can be used as a test for positive definiteness.
So I don't really see a contradiction to the docstring here. Or am I missing something?

marvinpfoertner:
> It's great to finally have this property system. I have a few comments which should be easily resolved. Thanks for the effort! One question to think about for the future (not this PR) what do we do with semi-definite operators? Add another property `_is_positive_semi_definite`? Often there are algorithms which also work if the linear operator is at least positive semi-definite.

I agree, we should add more matrix properties over time. I also think that we will need `is_indefinite`, as well as `is_diagonal` and `is_orthogonal` (the latter for SVDs and eigendecompositions). I started a list of these properties in #571 and #569. Feel free to add `is_positive_semidefinite` there.

src/probnum/linops/_linear_operator.py
JonathanWenger: My mistake! We might want to be lenient here and only raise an exception if the properties are `False`. Otherwise this may lead to a lot of boilerplate in particular when prototyping. Your call.

src/probnum/linops/_linear_operator.py
JonathanWenger: They do, but `Symmetrize` is really only used for `SymmetricKronecker` and is arguably confusing in its naming if there is a `symmetrize()` method. My suggestion would be to fold `Symmetrize` into the `SymmetricKronecker` implementation as a (future) refactor. I'm okay with considering this as out of scope for this PR.

JonathanWenger (approved):

src/probnum/linops/_linear_operator.py
JonathanWenger: In fact, I think it would be more in line with Python philosophy to attempt the Cholesky decomposition if it is not known whether the linop is spd and then raise an exception should Cholesky fail.

marvinpfoertner: In principle, I agree with the second point. In fact this is what happens if `is_positive_definite` is `None`: We attempt the Cholesky and if it fails, we set it to `False` and raise an exception. However, after giving this some thought, this issue is much more complicated with `is_symmetric`.
Since matrices are rarely exactly symmetric, an automatic check of symmetry would be much too restrictive, unless we have some sort of "asymmetry tolerance", which is just another one of these arbitrary parameters to make things work. Personally, I came to the conclusion that it is better if the user has to call `.symmetrize` on the operator before calling `.cholesky` because it makes this whole symmetry business much more transparent.
Moreover, if `symmetrize` is too costly, the user can always choose to set `linop.symmetric = True` manually, even if the property is only fulfilled approximately.

marvinpfoertner: Note that NumPy and SciPy circumvent this issue by only operating on the lower triangle of the matrix.
```python
>>> a = np.array([
...     [1.0, 2.0],
...     [0.9, 1.0],
... ])
>>> np.linalg.cholesky(a)
array([[1.        , 0.        ],
       [0.9       , 0.43588989]])
>>> scipy.linalg.cholesky(a, lower=True)
array([[1.        , 0.        ],
       [0.9       , 0.43588989]])
```
IMHO, this is quite dangerous, since it can produce very subtle bugs.

JonathanWenger: Sounds good to me! Thanks for the detailed explanation.

marvinpfoertner: If we find a good solution, I'm also fine with removing this in the future.
However, I would propose to make this a separate PR, since this would increase complexity by quite a bit.
We can always make interfaces less restrictive

src/probnum/linops/_linear_operator.py
marvinpfoertner: I agree, let's make a separate PR here :+1

PR #626, opened 2022-02-08
Pin the pytest version and replace jaxlib dependency

pnkraemer:
# In a Nutshell
Pin the `pytest` version to <7.0, replace the `jaxlib` dependency with `jax[minimum-jaxlib]`, and update the intersphinx-mapping for scipy.

# Detailed Description
* `pytest` 7.0 does not work with pytest-cases at the moment. 
* not having a jaxlib dependency next to a jax dependency simplifies the dependency updates (e.g. for the dependabot)
* Both are fixed in the same PR, because otherwise there might be unnecessary conflicts.
* The scipy intersphinx mapping has changed. It is updated (thanks to https://gist.github.com/bskinn/0e164963428d4b51017cebdb6cda5209).


setup.cfg
```diff
@@ -54,8 +54,7 @@ exclude =
 
 # Autodiff backends
 jax =
-    jax<0.2.28; platform_system!="Windows"
-    jaxlib<0.1.77; platform_system!="Windows"
+    jax[minimum-jaxlib]<0.2.28; platform_system!="Windows"
```
marvinpfoertner, line 57:
```suggestion
    jax[minimum-jaxlib]<0.2.29; platform_system!="Windows"
```
This would directly take care of #624

marvinpfoertner (approved):

JonathanWenger:
Note that a fix has been merged for the pytest-cases issue. See https://github.com/smarie/python-pytest-cases/releases/tag/3.6.9

pnkraemer:
> Note that a fix has been merged for the pytest-cases issue. See https://github.com/smarie/python-pytest-cases/releases/tag/3.6.9

Yup, I saw it. Thanks!

Does it affect this PR in any way, though? I  could either

a) Pin the version to <= 7.0.0, (and the pytest-cases version to 3.6.9?) to include the most recent pytest-cases update
b) Do nothing, and let the dependabot take care of this eventually (this is what it is for, right?)

I personally think a) makes most sense, but I am not very experienced with this. So please let me know how to proceed :)

pnkraemer:
I just saw that the CI in #614 passes without hickups (somehow scipy seems to have fixed their doc-host again), so I could also abandon this PR and reopen a smaller one with only the jax[minimum-jaxlib] change. What do you guys think?

marvinpfoertner:
> I just saw that the CI in #614 passes without hickups (somehow scipy seems to have fixed their doc-host again), so I could also abandon this PR and reopen a smaller one with only the jax[minimum-jaxlib] change. What do you guys think?

In light of the new information, from my POV, the best course of action would be to pin `pytest<8.0.0` and `pytest-cases>=3.6.9,<4.0.0`, get rid of the SciPy related changes and apply the `jax[minimum-jaxlib] < 0.2.29` change.

Issue #627, opened 2022-02-09
Refactoring of Markov processes and Kalman filters


pnkraemer:
## Describe the issue
The current implementation of Markov processes and Kalman filters is agnostic of two rather important concepts:
* Whether the underlying state-space model is discrete or continuous
* Whether the prior or observation model is linear or nonlinear

The implementations are tied to a single ``MarkovProcess`` and ``Kalman`` class. The distinctions between the above settings are made by passing corresponding transitions and/or linearisation-wrappers to the ``MarkovProcess`` or the ``Kalman`` class.

This works fine, but has two major issues:
* **It is difficult to maintain, because unused variables are flying around.** There is zero need for variables like `t` and `dt` in discrete transitions and discrete Kalman filters, and removing them wherever applicable will make the code much more readable.
* **It is difficult to maintain, because one has to jump back and forth between many files to track down an error in the code.** The linearised information operators point to extended kalman filters which point to transitions which are then used in ODE filters etc. This can, and should, be made much clearer. 
* **It is not clear how many features actually exist.** The feature list implemented in filtsmooth is quite extensive: (non)iterated, discrete/continuous, kalman/ukf/ekf filters/smoothers and _all_ combinations thereof are implemented, but it could be made much easier to use those.

## Additional context
@schmidtjonathan did I miss anything?

## Tasks

This will touch _a lot_ of code in `filtsmooth` and `randprocs.markov`, so below I will try to compile a set of micro-PR-type tasks. I will edit freely as soon as I can come up with more. The tasks are not ordered.

- [ ] Split the filtsmooth tutorial notebooks into smaller notebooks (4 instead of 2: linear/nonlinear, continuous/discrete filtering). Among other things, this will help with future notebook-merge-conflicts.
- [ ] Split the Kalman filter implementation into a discrete and a continuous version
- [ ] Make a `MarkovSequence` type class as a discrete counterpart to a `MarkovProcess`. This can then be passed to discrete filters
- [ ] Implement ExtendedKalman, UnscentedKalman, etc., by subclassing the Kalman filters and wrapping nonlinearities into `EKFComponents`, `UKFComponents`, etc.. Make the latter private.
- [ ] Replace `_EKFComponents`, `_UKFComponents`, etc., with pure functions `linearise_ek/uk` that turn a nonlinear model into a linear model.
- [ ] Remove the inheritance of `DiscreteTransition`s from `Transition`. Rename the latter to `SDETransition`?
[...]

PR #628, opened 2022-02-09
Split filtsmooth notebooks

pnkraemer:
# In a Nutshell
Splits the "linear Gaussian filtering and smoothing" into two pieces: a discrete and a continuous state-space model.

# Detailed Description
See #627.


schmidtjonathan (approved):
Thanks ✨

PR #630, opened 2022-02-09
Scipy intersphinx mapping update

pnkraemer:
# In a Nutshell
Scipy apparently moved its intersphinx mapping (https://github.com/scipy/scipy/issues/14267). I am not sure why the doc-build still works, but it seems that the new address is https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/. 

This PR is split off from #626, to not interfere with merging the pytest version pin. If it should not be merged, then we can shoot down this one here instead of slowing the other one (no hard feelings in this case :)) .

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -144,7 +144,7 @@ nbsphinx_execute_arguments = [
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
-    "scipy": ("https://docs.scipy.org/doc/scipy/reference", None),
+    "scipy": ("https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/", None),
     "matplotlib": ("https://matplotlib.org/", None),
 }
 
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -144,7 +144,7 @@
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
-    "scipy": ("https://docs.scipy.org/doc/scipy/reference", None),
+    "scipy": ("https://docs.scipy.org/doc/scipy-1.8.0/html-scipyorg/", None),
```
rgommers, line 147: Does it work with simply `'scipy': ('https://docs.scipy.org/doc/scipy', None),`? That seemed to work for numpy, and is the preferred link.

JonathanWenger: +1 We also shouldn't pin the version of the docs here if in any way possible.

pnkraemer: Thanks for the feedback!  I changed it to `https://docs.scipy.org/doc/scipy` now. 

A few days ago pinning the 1.8 version was the only way to make it work for me, but it seems like now it is not necessary anymore. :)

marvinpfoertner (approved):

PR #636, opened 2022-02-11
Clean up ProbNum's Development Dependencies

marvinpfoertner:
# In a Nutshell
This PR cleans up ProbNum's development dependencies.

# Detailed Description
Currently, dependencies for benchmarking, linting and code formatting are specified in the tox file, which makes it impossible to install them easily into a local development environment.
Moreover, test dependencies are specified as extra dependencies for the package, which is a bit undesirable, since these extra dependencies are user-facing on PyPI.
This PR creates dedicated requirements files for these different purposes, as well as a convenient `dev-requirements.txt` file, which allows for installing the entirety of ProbNum's dev dependencies in one call.

# Related Issues
None


benchmarks/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Dependencies needed to run the benchmarks
+
+# asv>0.5.1,<=0.6
```
marvinpfoertner, line 3: ASV started creating proper releases again a few days ago.
We should just use the PyPI versions.

formatting-requirements.txt
```diff
@@ -0,0 +1,6 @@
+# Dependencies for code formatting
+
+# black>=22.1,<23
```
marvinpfoertner, line 3: Black is out of beta, so we can rely on their stability policy

formatting-requirements.txt
```diff
@@ -0,0 +1,6 @@
...
+black == 20.8b0
+# isort>=5.10,<6
```
marvinpfoertner, line 5: isort seems to stick to SemVer stability promises

marvinpfoertner:
@JonathanWenger This is now ready for review :)

benchmarks/requirements.txt
JonathanWenger: I agree. Is there a reason this is just commented out for now? I suggest we make the change in this PR.

formatting-requirements.txt
JonathanWenger: I would extend comment here and reference a corresponding issue which references that the entire library should be blacked once more with an ignore for the diff.

JonathanWenger (approved):

benchmarks/requirements.txt
marvinpfoertner: I thought there might be an issue with the benchmark runner, but I agree, we should probably make this change now

JonathanWenger: The reason this was pinned was, if I remember correctly, that running the benchmarks quickly in dev mode was not available for the release. This made it hard to include the benchmarks into the CI.

marvinpfoertner: Ok, I added the PyPI version

formatting-requirements.txt
marvinpfoertner: I added a comment, but we don't need to open an issue, since dependabot will remind us about this update :)

PR #638, opened 2022-02-11
Update the `isort` config and bump `black` version

marvinpfoertner:
# In a Nutshell
`black` released a stable version recently, so we can rely on their stability policy and constrain only the major version number.
Also, I made some updates to the `isort` config, related to the upcoming compute backend PR.

# Detailed Description
By default, `isort` puts `as` imports on separate lines, e.g.
```python
from probnum import randprocs as _randprocs
from probnum import randvars as _randvars
```
Setting `combine_as_imports = true` results in
```python
from probnum import (
    randprocs as rprocs,
    randvars as rvs,
)
```
which is much more readable.

Moreover, by default, isort first lists all `import ...`-style imports and then all `from ... import`-style imports
```python
import functools
import os
import pathlib
from functools import partial
```
which is confusing. By setting `force_sort_within_sections = true`, the from imports are sorted as expected:
```python
import functools
from functools import partial
import os
import pathlib
```

# Related Issues
None


.git-blame-ignore-revs
```diff
@@ -1,2 +1,5 @@
 # Migrate code style to black
 636049eeaa3f515ff39a72d546363bbfd1f140dc
+
+# Update isort config and bump black version
```
JonathanWenger:
```suggestion
# Update isort config and bump black version to >=22.1
```

JonathanWenger (changes requested):
Make sure to also update the `.pre-commit-config.yaml` to the right black and isort version. We should also make an announcement to that effect so that people can reinstall their pre-commit hooks.

marvinpfoertner:
@JonathanWenger I added the correct `isort` and `black` versions to `.pre-commit-config.yaml`

JonathanWenger (approved):

PR #639, opened 2022-02-11
Add `Function` class and refactor `RandomProcess.{input,output}_{dim => shape}`

marvinpfoertner:
# In a Nutshell
This PR rounds off our `RandomProcess` implementation by introducing a `Function` class to "annotate" a callable with input and output shape information.

# Detailed Description
- [x] Add `Function` class with `input_shape` and `output_shape` attributes
- [x] Refactor `RandomProcess.mean` to be a `Function`
- [x] Refactor `RandomProcess.cov` to be a `Kernel`
- [x] Refactor `RandomProcess.{input,output}_{dim => shape}`. This makes handling of corner cases (scalar in- and outputs) much easier and saves a lot of code.
- [x] Refactor `Kernel` to use `input_shape` instead of `input_dim`
- [x] Make tests run again
- [x] Update tests
- [x] Update docstrings

# Related Issues
None


JonathanWenger:
I added some tests and docstrings. If you could double check the docs in particular (with regard to shapes etc.) that would be great.

src/probnum/randprocs/_gaussian_process.py
```diff
@@ -63,45 +61,61 @@ class GaussianProcess(_random_process.RandomProcess[_InputType, _OutputType]):
 
     def __init__(
         self,
-        mean: Callable[[_InputType], _OutputType],
+        mean: _function.Function,
```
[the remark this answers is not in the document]
marvinpfoertner: I agree. The motivation here was that I did not want to assume general `Function`s to have scalar outputs by default. But since `Kernel`s have `shape=()` by default it seemed reasonable to have a `MeanFunction` doing the same thing

marvinpfoertner:
> @marvinpfoertner I added some tests and docstrings. If you could double check the docs in particular (with regard to shapes etc.) that would be great.

Thanks for your help @JonathanWenger! :) I'll probably finalize this tomorrow at the latest.

src/probnum/__init__.py
```diff
@@ -34,18 +34,23 @@
...
 )
+from ._function import Function, LambdaFunction
 from ._version import version as __version__
 from .randvars import asrandvar
 
 # Public classes and functions. Order is reflected in documentation.
 __all__ = [
     "asrandvar",
+    "Function",
+    "LambdaFunction",
```
marvinpfoertner, line 45:
```suggestion
```
I'm not actually sure that we want those here, since `from probnum import *` would import `Function` into the current namespace, but I guess that's the user's fault then 😁

JonathanWenger: I guess it's the same with PN method components. Eventually we could move this to `backend` maybe? For now if we want it to be documented than it should be imported into the `probnum` namespace if the module is there.

PR #644, opened 2022-02-15
Split nonlinear filtering notebooks

pnkraemer:
# In a Nutshell
Splits the nonlinear filtering notebooks into discrete and continuous versions. And, while at it, splits the particle filter notebooks into two as well.


schmidtjonathan (approved):
Thanks :)

PR #645, opened 2022-02-16
Split discrete and continuous Kalman filter implementations

pnkraemer:
# In a Nutshell
Splits the current `Kalman` implementation into a discrete and a continuous part. The `DiscreteKalman` does Gaussian filtering and smoothing for discrete-discrete state-space models. The `ContinuousKalman` does the same for continuous-discrete state-space models. The goal of this PR is to make another step towards easier-to-maintain filtering and smoothing code.
(This PR also does some first mini-steps in simplifying the filter() implementation in the discrete object.)

# Detailed Description
The filter() and smooth() functions in the Kalman filter currently carry around a lot of dead weights for discrete models. Most notably, a `dt` variable, which does not make sense. (Also, the `t` variable is pointless, but this can only be refactored as a part of a bigger effort; detailed in #627.)
Soon, when there are suitable priors/problem classes for discrete vs continuous state-space models (see also #646), and when approximate Kalman filters are refactored, the gains obtained through the separation introduced here will be even more evident. 

Some further changes:
* Iterated filtsmooth in Kalman was not used anymore, so I deleted it. (It is superseded by `filtsmooth.gaussian.optim`)
* discrete-discrete iterated Kalman smoothing did not have a test case, which I added.


src/probnum/filtsmooth/gaussian/_kalman.py
```diff
@@ -21,28 +21,8 @@
...
-    """
-
-    def iterated_filtsmooth(
-        self,
-        regression_problem: problems.TimeSeriesRegressionProblem,
-        init_posterior: Optional[_kalmanposterior.SmoothingPosterior] = None,
-        stopcrit: Optional[FiltSmoothStoppingCriterion] = None,
-    ):
+class _KalmanBase(_bayesfiltsmooth.BayesFiltSmooth):
+    def iterated_filtsmooth(self, *args, **kwargs):
```
pnkraemer, line 25: This wrapper around the state generator works with args and kwargs for the following reason:

Soon, the regression problems might change types (discrete vs continuous), and the `filter()` and `iterated_filtsmooth_generator()` signature adapts accordingly. But this function here (and the `filtsmooth()` style methods below) remains the same. args and kwargs lead to a smaller, more meaningful diff in the future.

src/probnum/filtsmooth/gaussian/_kalman.py
```diff
@@ -165,30 +42,20 @@ def filtsmooth(
...
         filter_posterior, info_dicts = filter_result
         smooth_posterior = self.smooth(filter_posterior)
         return smooth_posterior, info_dicts
 
-    def filter(
-        self,
-        regression_problem: problems.TimeSeriesRegressionProblem,
-        _previous_posterior: Optional[_timeseriesposterior.TimeSeriesPosterior] = None,
-    ):
+    def filter(self, *args, **kwargs):
```
pnkraemer, line 50: This wrapper around the state generator works with args and kwargs for the following reason:

Soon, the regression problems might change types (discrete vs continuous), and the filter() and iterated_filtsmooth_generator() signature adapts accordingly. But this function here (and the filtsmooth() style methods below) remains the same. args and kwargs lead to a smaller, more meaningful diff in the future.

PR #646, opened 2022-02-16
Introduce MarkovSequence

pnkraemer:
# In a Nutshell
This PR introduces a `MarkovSequence` as a discrete-"time" counterpart to the `MarkovProcess`.

# Detailed Description
* For context, see #627. 
* A parallel refactoring for Kalman filter implementations (continuous vs discrete) happens in #645, but each PR is standalone.
* For the moment, functionality has not been touched much, because it would've led to a domino effect and suddenly, every function in `filtsmooth` would have to be changed. Instead, I decided to nudge the implementations towards the correct way by raising an error if the incorrect process type is used, but the functionality remains the same. Functionality changes will be introduced gradually in future PRs, as part of #627.
* I also deleted the redundant tests for generate_artificial_measurements, because if this function has even the slightest dysfunction, many tests will break anyway. I found myself refactoring this test code frequently, usually because of some minor changes in the `Transition` API (which should not have been the case).


src/probnum/randprocs/markov/_markov.py
```diff
@@ -69,6 +46,23 @@ def mean(self) -> _function.Function:
     def cov(self) -> "MarkovProcess.Kernel":
         return MarkovProcess.Kernel(self.__call__)
 
+    class Kernel(kernels.Kernel):
```
pnkraemer, line 49: the codecov complaints from here on below are not the fault of this PR, because it simply moves the objects around. So I would propose to ignore them.

src/probnum/randprocs/markov/_markov.py
```diff
@@ -107,27 +101,67 @@ def _sample_at_input(
...
-    ) -> np.ndarray:
-        raise NotImplementedError
+        *,
+        initarg: np.ndarray,
+        initrv: randvars.RandomVariable,
+        transition: continuous.SDE,
+    ):
+        if not isinstance(transition, continuous.SDE):
+            msg = "The transition is not continuous. Did you mean 'MarkovSequence'?"
+            raise TypeError(msg)
```
pnkraemer, line 138: I do not see the point of covering this exception with a test, so I would propose to ignore codecov here as well.

src/probnum/randprocs/markov/_markov.py
```diff
@@ -107,27 +101,67 @@ def _sample_at_input(
...
+    def __init__(
+        self,
+        *,
+        initarg: np.ndarray,
+        initrv: randvars.RandomVariable,
+        transition: continuous.SDE,
+    ):
+        if not isinstance(transition, discrete.NonlinearGaussian):
+            msg = "The transition is not discrete. Did you mean 'MarkovProcess'?"
+            raise TypeError(msg)
```
pnkraemer, line 160: see above.

pnkraemer:
The only failing part of the CI is the patch-coverage. I explained in the diff why I think it should be ignored here.

src/probnum/randprocs/markov/_markov.py
marvinpfoertner: I agree. See the comment above to resolve the codecov message :)

src/probnum/randprocs/markov/_markov.py
```diff
@@ -124,3 +100,68 @@ def _evaluate(self, x0: np.ndarray, x1: Optional[np.ndarray]) -> np.ndarray:
...
+    """
+
+    def __init__(
+        self,
+        *,
+        initarg: np.ndarray,
+        initrv: randvars.RandomVariable,
+        transition: continuous.SDE,
+    ):
+        if not isinstance(transition, continuous.SDE):
```
marvinpfoertner, line 136:
```suggestion
        if not isinstance(transition, continuous.SDE):  # pragma: no cover
```
should ignore the whole if clause in coverage

src/probnum/randprocs/markov/_markov.py
```diff
@@ -124,3 +100,68 @@ def _evaluate(self, x0: np.ndarray, x1: Optional[np.ndarray]) -> np.ndarray:
...
+    """Discrete-time Markov processes."""
+
+    def __init__(
+        self,
+        *,
+        initarg: np.ndarray,
+        initrv: randvars.RandomVariable,
+        transition: continuous.SDE,
+    ):
+        if not isinstance(transition, discrete.NonlinearGaussian):
```
marvinpfoertner, line 158:
```suggestion
        if not isinstance(transition, discrete.NonlinearGaussian):  # pragma: no cover
```

marvinpfoertner (approved):
Nice. I like the separation between the two types of random process 👍🏻

PR #650, opened 2022-02-17
Remove `input_shape` broadcasting capabilities from `Kernel`

marvinpfoertner:
# In a Nutshell
The previous refactor of the `Kernel` introduced the capability of broadcasting the dimensions corresponding to the `input_shape` of the `Kernel`. This turned out to be difficult to implement and document. Since we don't need the feature, we simplify the code and documentation by removing it.

# Detailed Description
- remove input shape broadcasting from `Kernel` and `Function`
- require the shape of `Kernel` and `Function` inputs to match `input_shape` along the last dimensions
- update docs
- update tests
- new tests corresponding to `Kernel`s with `input_shape=()`

# Related Issues
None


src/probnum/randprocs/kernels/_kernel.py
```diff
@@ -343,138 +312,97 @@ def _evaluate(
...
-                )
-                raise ValueError(err_msg) from ve
-        else:
-            try:
-                np.broadcast_to(
-                    x0,
-                    shape=x0.shape[: x0.ndim - self._input_ndim] + self._input_shape,
-                )
+        err_msg = (
+            "The shape of the input array `{argname}` must match the `input_shape` "
```
JonathanWenger, line 367: Are there some f-strings missing? The expressions `{argname}` and `{shape}` are not actually replaced by the corresponding variables.

src/probnum/randprocs/kernels/_kernel.py
```diff
@@ -495,8 +423,8 @@ class IsotropicMixin(abc.ABC):  # pylint: disable=too-few-public-methods
     def _squared_euclidean_distances(
         self, x0: np.ndarray, x1: Optional[np.ndarray]
     ) -> np.ndarray:
-        """Implementation of the squared Euclidean distance, which supports kernel
-        broadcasting semantics."""
+        """Implementation of the squared Euclidean distance, which supports scalar
```
JonathanWenger, line 426: Why scalar inputs? Doesnt this compute pairwise squared euclidean distances between _vectors_? Same for the related functions below.

JonathanWenger (changes requested):

src/probnum/randprocs/kernels/_kernel.py
marvinpfoertner: These are replaced using str.format below

src/probnum/randprocs/kernels/_kernel.py
marvinpfoertner: Conceptually yes, but in order to unify implementation, we also need to be able to handle the base-cases `input_shape=(1,)` and `input_shape=()`. The latter is technically a scalar.

JonathanWenger (approved):

PR #652, opened 2022-02-25
`Kernel.output_shape` semantics and further `RandomProcess` refactoring

marvinpfoertner:
# In a Nutshell
A few months ago, I redefined the shape of the outputs of `Kernel.__call__` to `cov.output_shape + batch_shape`.
However, this is very counterintuitive when in the context of random processes, since we would expect the outputs of the `mean`, `var`, and `std` functions to be `batch_shape + randproc.output_shape`, i.e. "the other way around".
My original change was made with multi-output GPs in mind, but even here it turns out that the `batch_shape + cov.output_shape`-order is easier to handle.

# Detailed 
- rename `Kernel.{shape => output_shape}`
- change output shapes of `Kernel.{__call__, matrix}` to `batch_shape + kernel.output_shape`
- adapt random processes to these changes
- fixed a bug in `Matern` 3/2 kernel
- refactor `RandomProcess`
  - add `input_ndim` and `output_ndim` as public arguments (also add these to `Function` and `Kernel`)
  - add default implementation of `_sample_at_input`
  - remove unnecessary overrides of `push_forward` in derived classes
  - add `mean` and `cov` arguments to the `RandomProcess` constructor and provide default implementations
  - improve the docstrings

# Related Issues
None

### DIFF
--- a/src/probnum/_function.py
+++ b/src/probnum/_function.py
@@ -56,6 +56,11 @@ class Function(abc.ABC):
         For scalar-valued function, this is an empty tuple."""
         return self._output_shape
 
+    @property
+    def output_ndim(self) -> int:
+        """Syntactic sugar for ``len(output_shape)``."""
+        return self._output_ndim
+
     def __call__(self, x: ArrayLike) -> np.ndarray:
         """Evaluate the function at a given input.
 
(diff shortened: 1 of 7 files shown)


JonathanWenger (approved):
Nice work! I didn't find anything that needs to be changed.

PR #654, opened 2022-03-01
Kernel arithmetic

JonathanWenger:
# In a Nutshell
This PR adds basic arithmetic for `Kernel`s. Kernel arithmetic allows user-friendly definition of custom kernels from existing ones.

# Detailed Description
This PR adds a `SumKernel`, `ScaledKernel` and `ProductKernel` as well as addition and multiplication of kernels. This allows the definition of kernels in the following fashion

```python
import numpy as np
from probnum.randprocs.kernels import ExpQuad, WhiteNoise

# Data
x = np.linspace(-1, 1, 5)

# Kernel via arithmetic
k = ExpQuad(input_shape=()) + 0.01 * WhiteNoise(input_shape=())

# Kernel matrix
k.matrix(x)
"""
array([[1.01      , 0.8824969 , 0.60653066, 0.32465247, 0.13533528],
       [0.8824969 , 1.01      , 0.8824969 , 0.60653066, 0.32465247],
       [0.60653066, 0.8824969 , 1.01      , 0.8824969 , 0.60653066],
       [0.32465247, 0.60653066, 0.8824969 , 1.01      , 0.8824969 ],
       [0.13533528, 0.32465247, 0.60653066, 0.8824969 , 1.01      ]])
"""
```

# Other
This PR deliberately does not introduce an operator registry analogously to the one for `LinearOperator`s yet. Algebraic reductions as for `LinearOperator`s are most likely way less common for kernels.

### DIFF
--- a/src/probnum/randprocs/kernels/_kernel.py
+++ b/src/probnum/randprocs/kernels/_kernel.py
@@ -1,12 +1,16 @@
 """Kernel / covariance function."""
 
+from __future__ import annotations
+
 import abc
-from typing import Optional
+from typing import Optional, Union
 
 import numpy as np
 
 from probnum import utils as _pn_utils
-from probnum.typing import ArrayLike, ShapeLike, ShapeType
+from probnum.typing import ArrayLike, ScalarLike, ShapeLike, ShapeType
+
+BinaryOperandType = Union["Kernel", ScalarLike]
 
 
 class Kernel(abc.ABC):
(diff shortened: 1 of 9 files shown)


src/probnum/randprocs/kernels/_kernel.py
```diff
@@ -342,6 +357,7 @@ def _evaluate(
         k_x0_x1 :
             See "Returns" section in the docstring of :meth:`__call__`.
         """
+        raise NotImplementedError
```
marvinpfoertner, line 360: I don't think we technically need this here, since the `ABC` can not be instantiated without implementing this method

marvinpfoertner (approved):
Looks good to me :) 👍🏻

PR #658, opened 2022-03-10
`ImmediateStop` stopping criterion or `quad`

mmahsereci:
# In a Nutshell
This PR adds the `ImmediateStop` stopping criterion for the `quad` package. This is in preparation for changes in #620 . 

In essence, currently there is no good way to not run an acquisition loop when calling `integrate` in case the integration is supposed to happen on a fixed given dataset. The wrapper `bayesquad_from_data` currently uses a workaround that is prone to bugs. The latter will be refactored in a follow up PR as it is a bit larger.

# Detailed Description
n/a

# Related Issues
Closes #...

### DIFF
--- a/src/probnum/quad/__init__.py
+++ b/src/probnum/quad/__init__.py
@@ -9,6 +9,7 @@ choosing points to evaluate the integrand based on said model.
 from probnum.quad.solvers.policies import Policy, RandomPolicy
 from probnum.quad.solvers.stopping_criteria import (
     BQStoppingCriterion,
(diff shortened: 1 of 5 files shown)


mmahsereci:
@tskarvone FYI this the second PR factored out of #620 . I hope reviewing is more straightforward like this.

tskarvone:
Everything looks fine to me. It's good to have the current ad hoc workaround replaced with something more intuitive.

tskarvone (approved):

PR #662, opened 2022-03-18
Improved type hints in documentation

JonathanWenger:
# In a Nutshell
Type aliases are not properly shown in the documentation and usually expanded completely. This can make the documentation very hard to read and the typing information not very useful. 

Screenshot from 2022-03-18 22-23-39

This PR uses Sphinx native functionality to display type hints instead.

# Detailed Description
- Added `probnum.typing` to the API Documentation
- Removed plugin `sphinx-autodoc-typehints`
- Added Sphinx extension `autodoc` 
- Removed colons in some commonly used classes

If a type alias still is expanded completely in the documentation the following should be added to the file:

```
from __future__ import annotations
```

# Other

Note that now the new standard for documenting `Parameters` is to **not** use a colon (:). 
```
Parameters
---------------
arg :
    Argument, where the docs may show no type hint.
betterarg
    Argument which will be documented with a type hint.
```
Otherwise `autodoc` may interpret the type of the parameter as empty.


JonathanWenger:
Note that currently in some classes type hints for `Parameters` are not shown. This seems to be a bug in `automodapi`. As this might influence the entire way we currently generate the documentation (i.e. switch to a different plugin for api generation) I would prefer to consider this as out of scope for this PR.

docs/source/conf.py
```diff
@@ -54,22 +50,38 @@
...
+    "IntLike": "typing.IntLike",
+    "FloatLike": "typing.FloatLike",
+    "ShapeLike": "typing.ShapeLike",
+    "DTypeLike": "typing.DTypeLike",
+    "ArrayIndicesLike": "typing.ArrayIndicesLike",
+    "ScalarLike": "typing.ScalarLike",
+    "ArrayLike": "typing.ArrayLike",
+    "LinearOperatorLike": "typing.LinearOperatorLike",
+    "NotImplementedType": "typing.NotImplementedType",
+}
```
marvinpfoertner, line 70:
```suggestion
autodoc_type_aliases = {type_alias: f"typing.{type_alias}" for type_alias in pn.typing.__all__}
```

and then adding the type aliases to `pn.typing.__all__` might scale a little bit better

src/probnum/randvars/_constant.py
```diff
@@ -1,4 +1,5 @@
 """(Almost surely) constant random variables."""
```
marvinpfoertner, line 1: I did not notice this before, but I'm not sure if we want exactly this comment here. This reads more like "I'm pretty sure that this module contains constant random variables, but we can never know"

marvinpfoertner (approved):
Minor stuff, otherwise perfect 👍🏻

docs/source/conf.py
JonathanWenger: oh yeah this is much better. :+1:

src/probnum/randvars/_constant.py
JonathanWenger: Changed it to a better description.

src/probnum/randprocs/_random_process.py
```diff
@@ -41,7 +41,7 @@ class RandomProcess(Generic[InputType, OutputType], abc.ABC):
     --------
     ~probnum.randvars.RandomVariable : Random variables.
     ~probnum.randprocs.GaussianProcess : Gaussian processes.
-    ~probnum.randprocs.markov.MarkovProcess : Random processes with the Markov property.
+    ~probnum.randprocs.markov.MarkovProcess : Random processes with the Maqrkov property.
```
marvinpfoertner, line 44:
```suggestion
    ~probnum.randprocs.markov.MarkovProcess : Random processes with the Markov property.
```

marvinpfoertner (approved):
Just a typo left :)

PR #583, opened 2021-11-29
Matern embedding

tskarvone:
This pull request implements the following:

1. Product Matérn kernels, which are products of _d_ one-dimensional Matérn kernels.
2. Explicit expression not based on the Bessel function for the Matérn 3.5 kernel.
3. Kernel means and variances for product Matérn kernels and one-dimensional Matérn kernels.

PR #665, opened 2022-03-21
Fixing parameters expansion in `ProductMatern`

mmahsereci:
# In a Nutshell
The parameter expansion of `ProductMatern` now catches cases where `lengthscales` and `nus` are given by `numpy.ndarray`s of shape `()`. This was actually a bug since it broke some kernel conversion in `quad`. I added a test to catch the latter.

# Detailed Description
Bug in `quad` was introduced in #583 .

# Related Issues
Closes #...


JonathanWenger (approved):
👍🏽

PR #679, opened 2022-03-25
Pin versions of doc dependencies

JonathanWenger:
# In a Nutshell
Unpinned versions of the `pydata-sphinx-theme` and the dependency `Jinja` of `nbsphinx` caused the doc build to fail. This PR pins the appropriate versions and thus fixes the doc build.


marvinpfoertner (approved):
👍🏻

PR #685, opened 2022-03-27
Migrate `setup.cfg` to `pyproject.toml` and refactor `pyproject.toml`

marvinpfoertner:
# In a Nutshell
This PR migrates `setuptools` configuration from `setup.cfg` to `pyproject.toml` and refactors `pyproject.toml`

# Detailed Description
- migrate `setuptools` configuration from `setup.cfg` to `pyproject.toml`
- update `black` configuration
- update `isort` configuration
- remove `requirements.txt`

# Related Issues
None


marvinpfoertner:
FYI: This PR is probably best reviewed commit by commit

pyproject.toml
```diff
@@ -1,14 +1,93 @@
...
 build-backend = "setuptools.build_meta"
 
+# see https://setuptools.pypa.io/en/stable/userguide/pyproject_config.html#setuptools-specific-configuration
+
+[tool.setuptools]
+platforms = [
+    "any",
+]
+zip-safe = false
+packages = ["probnum"]
```
marvinpfoertner, line 87: I specified `probnum` here manually instead of collecting it magically

marvinpfoertner: This means that any additional packages under `src` that we want installed must also be listed here in the future, which gives a very nice overview of the package contents.

pyproject.toml
```diff
@@ -104,12 +171,31 @@ max-locals = 20
...
+# see https://pycqa.github.io/isort/docs/configuration/options.html
+
 [tool.isort]
-multi_line_output = 3
-include_trailing_comma = true
-force_grid_wrap = 0
-use_parentheses = true
-line_length = 88
+# see https://pycqa.github.io/isort/docs/configuration/profiles.html#black
+profile = "black"
```
marvinpfoertner, line 201: This combines all deleted options, but it will be kept up-to-date with `black`s behavior

setup.py
```diff
@@ -5,5 +5,28 @@
...
+] + extras_require["jax"]
+extras_require["pls_calib"] = [
+    "GPy",
+    # GPy can't be imported without matplotlib.
+    # This is and should not be a ProbNum dependency.
+    "matplotlib",
+]
+extras_require["full"] = (
+    extras_require["jax"] + extras_require["zoo"] + extras_require["pls_calib"]
+)
```
marvinpfoertner, line 26: This is unfortunate, but I still think it is better to add this here and get rid of `setup.cfg` now.
We need `setup.py` anyway and it is easier to migrate this to `pyproject.toml` once this is resolved than repeating this PR then.

pyproject.toml
```diff
@@ -1,14 +1,95 @@
...
+
+# see https://peps.python.org/pep-0621/
+
+[project]
+name = "probnum"
+description = "Probabilistic Numerics in Python."
+readme = "README.md"
+requires-python = ">=3.8,<3.11"
+license = {file = "LICENSE.txt"}
+authors = [
```
JonathanWenger, line 13: I realize having specific names and e-mail addresses here is probably part of the setuptools specification, but I am not a huge fan, since it will cause PRs when maintainers change and may not stay up-to-date. (I realize similar issues exist with the docs). I somewhat prefer the option to just specify this on GitHub and let the contributors list show who creates this package.

setup.py
JonathanWenger: Agreed.

JonathanWenger (approved):
Looks good to me! I like the consolidation of config files. The only change I'd suggest is to not explicitly specify authors and maintainers.

pyproject.toml
marvinpfoertner: In this case we should create a joint mail address for all ProbNum maintainers soon, since this is the only piece of contact information listed on PyPI. It also seems to be quite common in other open source project (see NumPy, SciPy, PyTorch)

PR #696, opened 2022-04-01
Fixed the `line-too-long` message for all modules under `probnum.linalg` subpackage.

peterw2333:
# In a Nutshell
Fixed the `line-too-long` message for all modules under `probnum.linalg` subpackage.

# Detailed Description
Splitted lines longer than 100 characters into multiple lines, each shorter than 100 characters, as required by pylint. Stopped ignoring `line-too-long` issue for `probnum.linalg` subpackage by changing configuration in `tox.ini`.
Solving each message in different pull requests as requested in #671.

# Related Issues
Finishes first to-do in #671.


peterw2333:
@JonathanWenger Could you please approve running the checks again? I have made a new commit in response to the test results stating that maximum line length should be 88, instead of the default of 100 in pylint (as I have wrongly assumed).

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
```diff
@@ -231,7 +236,8 @@ class BayesCG(ProbabilisticLinearSolver):
 
     def __init__(
         self,
-        stopping_criterion: stopping_criteria.LinearSolverStoppingCriterion = stopping_criteria.MaxIterationsStoppingCriterion()
+        stopping_criterion: stopping_criteria.LinearSolverStoppingCriterion =
+stopping_criteria.MaxIterationsStoppingCriterion()
```
JonathanWenger, line 240: This does not seem compatible with the style determined by our code formatter `black`. The `InvocationError` means there is a piece of code which is not correctly formatted. I suggest you run `tox -re format`. This should reformat this line correctly and satisfy `black`.

JonathanWenger (changes requested):

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
peterw2333: Done! Reformatted code using black and made a new commit for the changes.

peterw2333: Now that the formatting problem is fixed, but tests are raising errors :(
It looks like the problems are all with installling dependencies, which may be caused by installing the dependencies with `pip` rather than `pip3` as in https://stackoverflow.com/questions/57235210/i-have-problem-while-installing-requirement-txt. 
But installing dependencies with `pip` seems to be working fine in previous commits, so I am not sure whether this is the problem.

JonathanWenger: > Also, if we reformat the code as black suggested, i.e. put stopping_criterion: stopping_criteria.LinearSolverStoppingCriterion = stopping_criteria.MaxIterationsStoppingCriterion() in a single line, pylint would be raising line-too-long error for not breaking the long line. It looks like black formatting and pylint formatting are having a conflict here.

Here we'll just have to make a compromise and put in a pylint ignore. Our class names for the linear solvers are just too long to satisfy both black and pylint. So an ignore at that point is totally fine.

Thanks for taking care of this so diligently!

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
```diff
@@ -63,7 +65,8 @@ class ProbabilisticLinearSolver(
     See Also
     --------
     ~probnum.linalg.problinsolve : Solve linear systems in a Bayesian framework.
-    ~probnum.linalg.bayescg : Solve linear systems with prior information on the solution.
+    ~probnum.linalg.bayescg : Solve linear systems with prior information on the
+    solution.
```
JonathanWenger, line 69:
```suggestion
    ~probnum.linalg.bayescg : Solve linear systems with prior information on the solution.
```
A line break here causes the built docs to display another link instead of the full sentence as a description of `probnum.linalg.bayescg`.
Screenshot from 2022-04-08 18-32-29

src/probnum/linalg/solvers/information_ops/_linear_solver_information_op.py
```diff
@@ -9,12 +9,14 @@
...
-    For a given action, the information operator collects information about the linear system to be solved.
+    For a given action, the information operator collects information about the
+    linear system to be solved.
 
     See Also
     --------
     MatVecInformationOp: Collect information via matrix-vector multiplication.
-    ProjectedResidualInformationOp: Collect information via a projection of the current residual.
+    ProjectedResidualInformationOp: Collect information via a projection of the
+    current residual.
```
JonathanWenger, line 19:
```suggestion
    ProjectedResidualInformationOp: Collect information via a projection of the current residual.
```
Similar as above: current residual will be put into a new line in the compiled docs which is undesirable.

src/probnum/linalg/solvers/stopping_criteria/_linear_solver_stopping_criterion.py
```diff
@@ -7,12 +7,15 @@
...
+    Checks whether quantities tracked by the
+    :class:`~probnum.linalg.solvers.LinearSolverState` meet a desired
+    terminal condition.
 
     See Also
     --------
     ResidualNormStoppingCriterion : Stop based on the norm of the residual.
-    PosteriorContractionStoppingCriterion : Stop based on the uncertainty about the quantity of interest.
+    PosteriorContractionStoppingCriterion : Stop based on the uncertainty about
+    the quantity of interest.
```
JonathanWenger, line 18: Same as above.

src/probnum/linalg/solvers/_probabilistic_linear_solver.py
peterw2333: Thank you for replying! I think good formatting in any project plays an important role in enhancing readability and ease of understanding, and it is important for a contributor to maintain it, so I am a bit perfectionist on this issue.

> Here we'll just have to make a compromise and put in a pylint ignore.

That's what I think as well, for class names to be meaningful they must be long. So we have to make a compromise and put a pylint ignore at the start of that file.

JonathanWenger (approved):
LGTM! Thanks for taking care of this!

PR #699, opened 2022-04-10
Fixed `no-else-return` message in `probnum.linalg` module

peterw2333:
# In a Nutshell
Fixed `no-else-return` message in `probnum.linalg` module

# Detailed Description
Removes unnecessary `else` after `return`, and changing unneccessary `elif` after `return` to `if`.
[pylint no-else-return(R1705)](https://pylint.pycqa.org/en/v2.13.5/messages/refactor/no-else-return.html)

# Related Issues
Finishes objective 2 in #671


JonathanWenger (approved):
Looks good, thanks!

PR #704, opened 2022-04-18
Fixed `else-if-used` pylint message in `probnum.problems` subpackage

luffingluffy:
# In a Nutshell
Fixed the `else-if-used` message in `probnum.problems` subpackage.

# Detailed Description
Changed nested `else-if` statements to `elif` to reduce nesting as required by pylint. Stopped ignoring else-if-used issue for `probnum.problems` subpackage by changing configuration in `tox.ini`.

# Related Issues
Finishes 2nd to-do in #672


src/probnum/problems/zoo/diffeq/_ivp_examples.py
```diff
@@ -519,11 +519,10 @@ def lorenz96(t0=0.0, tmax=30.0, y0=None, num_variables=5, params=(8.0,)):
         if num_variables < 4:
             raise ValueError("The number of variables must be at least 4.")
         y0 = np.ones(num_variables) * constant_forcing
-    else:
-        if len(y0) < 4:
-            raise ValueError(
-                "The number of variables (i.e. the length of the initial vector) must be at least 4."
-            )
+    elif (len(y0) < 4):
```
pnkraemer, line 522: Would be great if you could remove the parentheses in the elif , because this makes the elif statement more similar to the rest of the code
```suggestion
    elif len(y0) < 4:
```

pnkraemer: It turns out that black and pylint also dont like the parentheses :)

luffingluffy: Ah, my bad! Just came off a Java project. Thanks for the catch though, I'll remember to install black next time!

pnkraemer (approved):

PR #705, opened 2022-04-19
Fixed `line-too-long` pylint message in `probnum` and `probnum.problems` subpackage

luffingluffy:
# In a Nutshell
Fixed the `line-too-long` message in `probnum` and `probnum.problems` subpackage.

# Detailed Description
Reformat long lines into multiple shorter lines as required by pylint. Stopped ignoring `line-too-long` issue in the global pass and in `probnum.problems` subpackage by changing configuration in tox.ini.

# Related Issues
Finishes 1st to-do in #672


src/probnum/problems/zoo/linalg/_suitesparse_matrix.py
```diff
@@ -56,9 +56,11 @@ def suitesparse_matrix(
...
     except ImportError as err:
         raise ImportError(
-            "Cannot query SuiteSparse Matrix collection without optional dependency ",
-            "`requests`. Install ProbNum with optional dependencies for the problem zoo via",
-            " `pip install probnum[zoo]` or install requests directly: `pip install requests`.",
+            "Cannot query SuiteSparse Matrix collection without",
+            "optional dependency `requests`. Install ProbNum with",
+            "optional dependencies for the problem zoo via",
+            "`pip install probnum[zoo]` or install requests",
+            "directly: `pip install requests`.",
```
pnkraemer, line 63: Whitespaces are missing behind 'without', 'with', 'via', and 'requests'. Without the extra whitespaces, the error message will be displayed incorrectly. 

```suggestion
            "Cannot query SuiteSparse Matrix collection without ",
            "optional dependency `requests`. Install ProbNum with ",
            "optional dependencies for the problem zoo via ",
            "`pip install probnum[zoo]` or install requests ",
            "directly: `pip install requests`.",
```

src/probnum/problems/zoo/linalg/_suitesparse_matrix.py
```diff
@@ -198,7 +200,11 @@ def _download(
             import requests  # pylint: disable=import-outside-toplevel
         except ImportError as err:
             raise ImportError(
-                "Cannot query SuiteSparse Matrix collection without optional dependency `requests`. Install ProbNum with optional dependencies for the problem zoo via `pip install probnum[zoo]` or install requests directly: `pip install requests`."
+                "Cannot query SuiteSparse Matrix collection without",
+                "optional dependency `requests`. Install ProbNum with",
+                "optional dependencies for the problem zoo via",
+                "`pip install probnum[zoo]` or install requests",
+                "directly: `pip install requests`.",
```
pnkraemer, line 207: same here

tox.ini
```diff
@@ -64,13 +64,13 @@ deps =
 ignore_errors = true
 commands =
     # Global Linting Pass
-    pylint src/probnum --disable="no-member,abstract-method,arguments-differ,arguments-renamed,redefined-builtin,redefined-outer-name,too-many-instance-attributes,too-many-arguments,too-many-locals,too-many-lines,too-many-statements,too-many-branches,too-complex,too-few-public-methods,protected-access,unnecessary-pass,unused-variable,unused-argument,attribute-defined-outside-init,no-else-return,no-else-raise,no-self-use,else-if-used,consider-using-from-import,duplicate-code,line-too-long,missing-module-docstring,missing-class-docstring,missing-function-docstring,missing-param-doc,missing-type-doc,missing-raises-doc,useless-param-doc,useless-type-doc,missing-return-type-doc" --jobs=0
```
pnkraemer, line 67: Could you please add this change ("Removing line-too-long from the global linting pass") to the PR description as well? This way, it is easier to trace things back if issues come up. :)

pnkraemer (changes requested):
Thanks! There were two instances of "incorrect" linebreaks, but once those are corrected, this PR can be merged. :)

tox.ini
luffingluffy: Yep, okay

src/probnum/problems/zoo/linalg/_suitesparse_matrix.py
luffingluffy: Hm, I initially added whitespaces after those words, however I did my own testing with `print()` and noticed that whitespaces were added where necessary as seen in the attached image. I might be just misunderstanding how Python string manipulation work though.. 

image

pnkraemer: Thanks for clarifying this. I did not know about the whitespace-print thing!

Technically, however, this is an exception not a print. For exceptions, this return-style produces odd outputs. Note how the "a", "b", "c" below is not concatenated:

Screenshot from 2022-04-23 09-56-50

However, this seems to be a little out of scope for this PR, which should only be about line-too-long statements. The linebreak issues I am talking about have been there before, so it is not your responsibility to fix them :) Let's handle this in a different pull request.

pnkraemer (approved):
Thanks a lot!

src/probnum/problems/zoo/linalg/_suitesparse_matrix.py
luffingluffy: Ah gotcha, I guess string concatenations are handled somewhat different for Exceptions in Python. Also, can I clarify whether the commas between the strings are meant to be displayed as well? I just followed the original code as I assumed that was the expected behaviour however, after seeing your image, it struck me as odd. Perhaps the following was the intended outcome? 
image

luffingluffy:
@pnkraemer I think you are merging changes to my fork instead. Did I make an error in choosing the merge direction for this?

pnkraemer:
No, everything is fine, I am just updating the fork the merge this PR into main soon. :)

src/probnum/problems/zoo/linalg/_suitesparse_matrix.py
pnkraemer: yes, something like this would be desirable in my opinion

PR #707, opened 2022-04-27
Fix flaky `quad` test.

### LINKED ISSUE #661: `test_kernel_mean_matern_lebesgue_measure` is a bit flaky.
**Describe the bug**
The test `test_kernel_mean_matern_lebesgue_measure` occasionally fails. It should be investigated why. 

Examples:
- https://github.com/probabilistic-numerics/probnum/runs/5539496853?check_suite_focus=true#step:5:24035
- https://github.com/probabilistic-numerics/probnum/runs/5500893583?check_suite_focus=true#step:5:23997

**Expected behavior**
Non-flaky test.

**Additional context**
The test tests an integral value which is notorious hard to test. In case the test uses an rng (I haven't checked yet), we should aim to avoid it. If it does not, I am quite unsure why the test fails occasionally.

mmahsereci:
# In a Nutshell
This PR will hopefully fix the flaky kernel embedding test as linked in the issue.

# Detailed Description
- make test use the provided `rng` fixture

# Related Issues
Closes #661

### DIFF
--- a/tests/test_quad/test_kernel_embeddings.py
+++ b/tests/test_quad/test_kernel_embeddings.py
@@ -113,9 +113,9 @@ def test_kernel_var_lebesgue_measure(kernel_embedding):
 @pytest.mark.parametrize("input_dim", [1, 2, 3])
 @pytest.mark.parametrize("num_data", [1, 2, 5])
 @pytest.mark.parametrize("measure_name", ["lebesgue"])
-def test_kernel_mean_matern_lebesgue_measure(matern_kernel, measure, num_data):
+def test_kernel_mean_matern_lebesgue_measure(matern_kernel, measure, num_data, rng):
     kernel_embedding = KernelEmbedding(matern_kernel, measure)
-    points = kernel_embedding.measure.sample(num_data)
+    points = kernel_embedding.measure.sample(rng=rng, n_sample=num_data)
     # Numerical integration using scipy's quad
     num_kernel_means = np.zeros((num_data,))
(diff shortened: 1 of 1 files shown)


tskarvone:
Just that I understand correctly the effect of this change: Previously we did not specify an rng with the effect that a random seed was used but now that we do specify it, a deterministic seed specified in the rng fixture is used instead.

It is odd that both Monte Carlo and Gaussian quadratures from `utils.py` are both used in the tests. Let us merge this now and next week I will see if we can use Gaussian quadrature in all the tests, which in my opinion would be a more robust solution than using deterministic Monte Carlo.

tskarvone (approved):

mmahsereci:
> Just that I understand correctly the effect of this change: Previously we did not specify an rng with the effect that a random seed was used but now that we do specify it, a deterministic seed specified in the rng fixture is used instead.

yes.

> It is odd that both Monte Carlo and Gaussian quadratures from `utils.py` are both used in the tests. Let us merge this now and next week I will see if we can use Gaussian quadrature in all the tests, which in my opinion would be a more robust solution than using deterministic Monte Carlo.

The points that are sampled using the seed are the points where the kernel mean is evaluated. By sampling it from the measure it can be ensured that the points lie in some meaningful range. It seems though that partially we use `scipy.integrate.quad` for 1D integrals and sometimes Gaussian quadrature. Unifying how we tests integrals would be nice indeed, especially as scipy's `quad` function has a somewhat arbitrary run-time.

PR #709, opened 2022-05-05
Scaling parameter for kernels

tskarvone:
The PR implements a scaling parameter `sigma_sq` for all covariance kernels. Currently this parameter is only implemented for the white noise kernel. The parameter scales all kernel evaluations as `sigma_sq * unscaled_kernel(x0, x1)`. 

A future PR will implement maximum likelihood estimation of the scaling parameter in `quad`.

### DIFF
--- a/src/probnum/quad/kernel_embeddings/_expquad_gauss.py
+++ b/src/probnum/quad/kernel_embeddings/_expquad_gauss.py
@@ -20,7 +20,7 @@ def _kernel_mean_expquad_gauss(
         \begin{equation}
             k_P(x)
             =
-            \det( I + \Sigma / l^2)^{-1/2}
+            \sigma^2 \det( I + \Sigma / l^2)^{-1/2}
             \exp\bigg(-\frac{1}{2}(x-\mu)^\maths{T} (l^2 I + \Sigma)^{-1}
                         (x-\mu) \bigg),
         \end{equation}
@@ -61,7 +61,7 @@ def _kernel_mean_expquad_gauss(
         exp_factor = np.exp(-0.5 * ((x - measure.mean) * chol_inv_x.T).sum(axis=1))
         det_factor = kernel.lengthscale**input_dim / np.diag(chol[0]).prod()
 
-    return det_factor * exp_factor
+    return kernel.sigma_sq * det_factor * exp_factor
 
 
 def _kernel_variance_expquad_gauss(kernel: ExpQuad, measure: GaussianMeasure) -> float:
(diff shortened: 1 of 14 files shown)


JonathanWenger:
How is this PR different from the existing functionality implemented in #654 ?

tskarvone:
> How is this PR different from the existing functionality implemented in #654? Consider the following, which works on `main`:

Thanks! Although Maren me there should be some sort of scaled kernel there already, I appear to have failed to find it. 

However, at least when it comes to estimating and updating kernel parameters I feel that it is much more natural to have the scale parameter an attribute of the kernel itself rather than only of a `ScaledKernel`. Otherwise the kernel needs either to be converted to a `ScaledKernel` somewhere in the code or the user needs to be required to provide a `ScaledKernel`. Would it make sense to include both (a) scale parameter as an attribute of every kernel and (b) `ScaledKernel` for complete kernel arithmetic?

We could alternatively ignore this PR for now and I could move on to implementing to implementing parameter estimation and updating for `quad`. That should give me a clearer idea on what is a good way to handle the scale parameter.

JonathanWenger:
> However, at least when it comes to estimating and updating kernel parameters I feel that it is much more natural to have the scale parameter an attribute of the kernel itself rather than only of a `ScaledKernel`. Otherwise the kernel needs either to be converted to a `ScaledKernel` somewhere in the code or the user needs to be required to provide a `ScaledKernel`. 

A quick solution would be simply to check whether the input is either a ScaledKernel or not and create a dictionary of the hyperparameters pointing to the kernel's parameters. A more long-term solution would be to equip every kernel with such a hyperparameter property that has all adjustable hyperparameters of the kernels it is a composite of. Just adding the `output_scale` to every kernel I don't think is a solution, since once you want to support kernels constructed from other kernels (e.g. `SumKernel`s) you are faced with the same issue.

> Would it make sense to include both (a) scale parameter as an attribute of every kernel and (b) `ScaledKernel` for complete kernel arithmetic?

I am not a fan of doing both since it creates two ways to do the same thing. In particular to a user it would then not be possible to specify a kernel for BQ as ``2.0 * kernels.ExpQuad(input_shape=())`` but only as  ``kernels.ExpQuad(input_shape=(), output_scale=2.0)``, whereas in other parts of the library the first one is acceptable. Creating two ways to do the same thing will also lead to more complicated code down the line since we always have to check for both options. 

> We could alternatively ignore this PR for now and I could move on to implementing to implementing parameter estimation and updating for `quad`. That should give me a clearer idea on what is a good way to handle the scale parameter.

:+1: I am fine with that. Whatever helps your progress.

tskarvone:
I had a very helpful discussion with @marvinpfoertner regarding this. The best approach will be to use `ScaledKernel`, so I will remove this PR.

mmahsereci:
> I had a very helpful discussion with @marvinpfoertner regarding this. The best approach will be to use `ScaledKernel`, so I will remove this PR.

Could one of you add the discussion here such that this PR can be referenced in the future in case the topic comes up again? @tskarvone @marvinpfoertner Thanks! :+1:

PR #725, opened 2022-08-31
Algebraic Operations on `Functions`

marvinpfoertner:
# In a Nutshell
This PR implements addition and scalar multiplication on `Function`s.

This is a proposal for how to redesign all algebraic structures in ProbNum.

# Detailed Description
- `Function + Function`
- `ScalarLike * Function`
- leverages `functools.singledispatchmethod` to register the correct result for combinations of different function types

# Related Issues
None


tests/test_functions/test_algebra.py
```diff
@@ -0,0 +1,100 @@
...
+    xs = rng.uniform(-1.0, 1.0, batch_shape + op0.input_shape)
+
+    np.testing.assert_array_equal(
+        fn_sub(xs),
+        op0(xs) - op1(xs),
+    )
+
+
+@pytest.mark.parametrize("scalar", [1.0, 3, 1000.0])
+def test_rmul_scalar_evaluation(
```
JonathanWenger, line 86: Maybe add a `mul` test as well here, just to make sure we're also covering the case `fn * scalar`?

JonathanWenger (approved):
Looks great! Just a few clarification questions, but otherwise good to go. Thanks :pray:!

tests/test_functions/test_algebra.py
marvinpfoertner: Good point, we did actually not support that before!
I added the `__mul__` operator and the test 👍🏻

src/probnum/functions/_zero.py
```diff
@@ -15,3 +15,11 @@ def _evaluate(self, x: np.ndarray) -> np.ndarray:
             x,
             shape=x.shape[: x.ndim - self._input_ndim] + self._output_shape,
         )
+
+    @functools.singledispatchmethod
+    def __add__(self, other):
+        return super().__add__(other)
+
+    @functools.singledispatchmethod
```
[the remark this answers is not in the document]
marvinpfoertner: Yes, in a sense. The decorator adds a dispatcher to `Function.__sub__`, but if we don't add this here once more, calls to `@Zero.__sub__.register` will actually be handled by the base class, which screws up the dispatching.

PR #735, opened 2022-11-07
Refactor LinearOperator for method overriding

timweiland:
# In a Nutshell
_matmul, _apply, _transpose etc. can now be implemented by overriding the corresponding methods in a subclass instead of passing the implementations in the constructor. The previous approach is still possible via the new LambdaLinearOperator.

# Detailed Description
* Refactor LinearOperator such that method implementations are provided via subclass overrides
* Introduce LambdaLinearOperator subclass which provides support for the old approach of passing method implementations via the constructor
* Replace all previous uses of LinearOperator with LambdaLinearOperator to maintain compatibility

### DIFF
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 from typing import Callable, Optional, Tuple, Union
+import abc
 
 import numpy as np
 import scipy.linalg
@@ -19,7 +20,7 @@ BinaryOperandType = Union[
 # pylint: disable="too-many-lines"
 
 
-class LinearOperator:
+class LinearOperator(abc.ABC):
     r"""Composite base class for finite-dimensional linear operators.
 
     This class provides a way to define finite-dimensional linear operators without
@@ -32,19 +33,21 @@ class LinearOperator:
     :class:`LinearOperator`, that defers linear operations to the original operators and
     combines the results.
 
-    To construct a concrete :class:`LinearOperator`, either pass appropriate callables
-    to the constructor of this class, or subclass it.
+    To construct a concrete :class:`LinearOperator`, subclass it and implement the
+    appropriate methods. The following methods may be implemented by a subclass:
+    ``_matmul`` (necessary), ``_rmatmul``, ``_apply``, ``_todense``, ``_transpose``,
+    ``_inverse``, ``_rank``, ``_eigvals``, ``_cond``, ``_det``, ``_logabsdet``,
+    ``_trace``.
+    Check the corresponding docstrings for more information.
 
-    A subclass must implement either one of the methods ``_matvec`` and ``_matmat``, and
-    the attributes/properties ``shape`` (pair of integers) and ``dtype`` (may be
-    ``None``). It may call the ``__init__`` on this class to have these attributes
-    validated. Implementing ``_matvec`` automatically implements ``_matmat`` (using a
-    naive algorithm) and vice-versa.
+    A subclass must implement the method ``_matmul`` and the attributes/properties 
+    ``shape`` (pair of integers) and ``dtype`` (may be ``None``).
+    It may call the ``__init__`` on this class to have these attributes
+    validated.
 
-    Optionally, a subclass may implement ``_rmatvec`` or ``_adjoint`` to implement the
-    Hermitian adjoint (conjugate transpose). As with ``_matvec`` and ``_matmat``,
-    implementing either ``_rmatvec`` or ``_adjoint`` implements the other automatically.
-    Implementing ``_adjoint`` is preferable; ``_rmatvec`` is mostly there for backwards
+    Optionally, a subclass may implement ``_rmatmul`` or ``_transpose`` to implement the
+    Hermitian adjoint (conjugate transpose).
+    Implementing ``_transpose`` is preferable; ``_rmatmul`` is mostly there for backwards
     compatibility.
 
     Parameters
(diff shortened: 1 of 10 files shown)


src/probnum/linops/_linear_operator.py
```diff
@@ -649,12 +762,25 @@ def __neg__(self) -> "LinearOperator":
...
+            a LinearOperator.
+        """
+        raise NotImplementedError("_transpose is not implemented.")
+
     @property
     def T(self) -> "LinearOperator":
-        if self.__transpose is None:
-            return self._transpose_fallback()
+        if self._is_implemented("_transpose"):
+            return self._transpose()
```
marvinpfoertner, line 782:
```suggestion
        try:
            self._transpose_cache = self._transpose()
        except NotImplementedError:
```

src/probnum/linops/_linear_operator.py
```diff
@@ -649,12 +762,25 @@ def __neg__(self) -> "LinearOperator":
...
 
-        return self.__transpose()
+        return self._transpose_fallback()
```
marvinpfoertner, line 783:
```suggestion
            self._transpose_cache = TransposeLinearOperator(self)

        return self._transpose_cache
```

src/probnum/linops/_linear_operator.py
```diff
@@ -697,20 +823,41 @@ def transpose(self, *axes: Union[int, Tuple[int]]) -> "LinearOperator":
...
     def _transpose_fallback(self) -> "LinearOperator":
-        if self.__rmatmul is not None:
+        if self._is_implemented("_rmatmul"):
             return TransposedLinearOperator(
                 self,
-                matmul=lambda x: self.__rmatmul(x[..., np.newaxis])[..., :],
+                matmul=lambda x: self._rmatmul(x[..., np.newaxis])[..., :],
             )
 
         return TransposedLinearOperator(self)
```
marvinpfoertner, line 832: I added the transpose fallback in the suggestions for the `_transpose` method, so we can simply remove this

src/probnum/linops/_linear_operator.py
```diff
@@ -992,7 +1190,163 @@ def _apply_to_matrix_stack(
...
+        logabsdet: Optional[Callable[[], np.flexible]] = None,
+        trace: Optional[Callable[[], np.number]] = None,
+    ):
+        self._matmul_attr = matmul  # (self @ x)
+        self._rmatmul_attr = rmatmul  # (x @ self)
+        self._apply_attr = apply  # __call__
+
+        self._todense_attr = todense
+
+        self._transpose_attr = transpose
```
marvinpfoertner, line 1290:
```suggestion
        self._transpose_attr = (
            transpose
            if transpose is not None
            else TransposeLinearOperator(self, matmul=rmatmul)
        )
```

src/probnum/linops/_linear_operator.py
```diff
@@ -992,7 +1190,163 @@ def _apply_to_matrix_stack(
...
+        super().__init__(shape, dtype)
+    
+    def _is_implemented(self, method_name: str) -> bool:
+        return getattr(self, method_name + "_attr") is not None
+        
+    def _call_if_implemented(self, method_name: str):
+        method = getattr(self, method_name + "_attr")
+        if method is not None:
+            return method
+        raise NotImplementedError()
```
marvinpfoertner, line 1310:
```suggestion
    def _call_if_implemented(self, method: Optional[callable]) -> callable:
        if method is not None:
            return method
        raise NotImplementedError()
```

marvinpfoertner (changes requested):
Thanks for taking care of this! 🙏🏻 

Some general points:
- Instead of just saying "You may implement this method in a subclass.", we should probably state that overriding the default implementation only makes sense if we know a more efficient implementation (in terms of memory or computation time) than the fallback.
- Can you get rid of the double underscores in the cache variables? We don't really need those anymore

JonathanWenger:
Great job on your first PR @timweiland. To fix the fromatting and isort issues, once you've installed tox you can run
```bash
tox -e format
```
To automatically keep code formatted you can also install the formatting libraries directly via
```
cd probnum
pip install -r formatting-requirements
```
You can then run `black` and `isort` on save in VSCode for example. To avoid comitting files which are incorrectly formatted in the future you can also alternatively install the provided precommit hooks to check for correct formatting prior to comitting (see https://probnum.readthedocs.io/en/latest/development/pull_request.html#pre-commit-hooks)

timweiland:
Thanks @JonathanWenger! I've fixed all of the problems now except for one which I'm unsure about:
Pylint complains that in this code, self.T is not callable (E1102): `y = (self.T)(x, axis=-1)`
But clearly, self.T returns a LinearOperator which is callable: `def T(self) -> "LinearOperator":`
Do you know how to fix this? [It seems that it might have something to do with the @property decorator.](https://stackoverflow.com/questions/53598374/pylint-giving-not-callable-error-for-object-property-that-is-callable)

timweiland:
> Thanks @JonathanWenger! I've fixed all of the problems now except for one which I'm unsure about: Pylint complains that in this code, self.T is not callable (E1102): `y = (self.T)(x, axis=-1)` But clearly, self.T returns a LinearOperator which is callable: `def T(self) -> "LinearOperator":` Do you know how to fix this? [It seems that it might have something to do with the @property decorator.](https://stackoverflow.com/questions/53598374/pylint-giving-not-callable-error-for-object-property-that-is-callable)

I talked to @marvinpfoertner about this and we agreed to simply make pylint ignore this line.

marvinpfoertner (approved):
Thanks again for doing this! :) Congrats on you first ProbNum PR! 🎉

PR #737, opened 2022-11-08
Caching of the LU decomposition computed by `_InverseLinearOperator`

marvinpfoertner:
# In a Nutshell
Currently, executing
```python
A = pn.linops.Matrix(...)
x1 = A.inv() @ b1
x2 = A.inv() @ b2
```
runs `scipy.linalg.lu_factor` on `A.dense` twice.
In principle, it would be possible to cache the return value of `A.inv()`, but that would introduce cyclic references,
which might cause memory overhead before being collected by the garbage collector.
An elegant solution is to cache all expensive quantities in the linear operator `A` itself, and then create "throwaway" objects like `_InverseLinearOperator` and `TransposeLinearOperator`, which are essentially views of `A`, accessing these quantities through the base linear operator.

# Detailed Description
- caching of LU decomposition in `LinearOperator`s
- bugfix in `_InverseLinearOperator.__rmatmul__`
- add test case for `_InverseLinearOperator`
- Replace custom implementation of `LinearOperator.broadcast_matmul` with `np.vectorize`-based implementation 

# Related Issues
None

### DIFF
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -1377,8 +1377,8 @@ class _InverseLinearOperator(LambdaLinearOperator):
         ----------
         x
             :code:`shape=(N,K)` --
-            The right-hand sides :math:`X` of the linear systems, where :attr:`shape`
-            :code:`=(N,N)`.
+            The right-hand sides :math:`X` of the linear systems, where
+            :code:`A.shape == (N, N)`.
         trans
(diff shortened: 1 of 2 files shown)


marvinpfoertner:
@JonathanWenger This PR should be reviewed after #735, since it branches off of this branch.

JonathanWenger (approved):
Nice! Elegant solution of the circular dependency issue.

PR #742, opened 2022-11-21
adding some info to `quad` docs.

mmahsereci:
# In a Nutshell
So far the `quad` docs do not mention anywhere which kernel embeddings are available for Bayesian quadarture. This PR adds this info to the docs. 

# Detailed Description
Further, made the helper function that checks the kernel-measure pair into a staticmethod as it made sense.

# Related Issues
Closes #...

PR #743, opened 2022-11-21
Consistent file structure for `quad`

mmahsereci:
# In a Nutshell
No changes in functionality. No changes in docs. Simply making file structure consistent among `quad` modules.

# Detailed Description
~~Should be reviewed after #742 is merged as it branches off of it.~~

# Related Issues
Closes #...


src/probnum/quad/integration_measures/_integration_measure.py
```diff
@@ -0,0 +1,77 @@
+"""Base class of an integration measures for Bayesian quadrature."""
```
JonathanWenger, line 1:
```suggestion
"""Base class of an integration measure for Bayesian quadrature."""
```

JonathanWenger (approved):
:+1:

PR #744, opened 2022-11-21
Make `quad` policies stateless

mmahsereci:
# In a Nutshell
The BQ policies were not stateless. This PR makes them stateless by passing the `rng` to the call function instead it being an attribute of the policy instances. The PR also adds some long overdue shape tests for the policies.

# Detailed Description
~~should be reviewed after #743 is merged since it branches off of it.~~

# Related Issues
Closes #...


src/probnum/quad/solvers/_bayesian_quadrature.py
```diff
@@ -278,6 +269,7 @@ def integrate(
         fun: Optional[Callable],
         nodes: Optional[np.ndarray],
         fun_evals: Optional[np.ndarray],
+        rng: Union[IntLike, np.random.Generator] = np.random.default_rng(),
```
JonathanWenger, line 272:
```suggestion
        rng: Optional[np.random.Generator] = None,
```

I would suggest to make the random number generator optional. It's always better to be explicit than implicit so that when a method that actually uses random numbers is used, the user is made aware through an exception if they do not provide their own RNG instance. Otherwise the global random state might be changed without the user being aware of it. With the arbitrary backend in #581 , rng states will always have to be passed anyways, since frameworks like JAX do not have global random state.

src/probnum/quad/solvers/_bayesian_quadrature.py
```diff
@@ -316,6 +310,10 @@ def integrate(
             If dimension of ``nodes`` or ``fun_evals`` is incorrect, or if their
             shapes do not match.
         """
+        # Get the rng
+        if isinstance(rng, get_args(IntLike)):
+            rng = np.random.default_rng(int(rng))
+
```
JonathanWenger, line 316:
```suggestion
```
See above, why I think this can be dangerous and should be avoided.

tests/test_quad/test_bayesian_quadrature.py
```diff
@@ -81,6 +83,30 @@ def test_bq_from_problem_defaults(bq_no_policy, bq):
...
+        (1000, None, None, MaxNevals),
+        (None, 1e-5, None, IntegralVarianceTolerance),
+        (None, None, 1e-5, RelativeMeanChange),
+        (None, 1e-5, 1e-5, LambdaStoppingCriterion),
+        (1000, None, 1e-5, LambdaStoppingCriterion),
+        (1000, 1e-5, None, LambdaStoppingCriterion),
+        (1000, 1e-5, 1e-5, LambdaStoppingCriterion),
+    ],
+)
+def test_bq_from_problem_stopping_condition_assignment(max_evals, var_tol, rel_tol, t):
```
JonathanWenger, line 99:
```suggestion
def test_bq_from_problem_stopping_criterion_assignment(max_evals, var_tol, rel_tol, t):
```

tests/test_quad/test_bayesian_quadrature.py
```diff
@@ -120,3 +146,10 @@ def test_integrate_wrong_input(bq, bq_no_policy, data):
         bq.integrate(fun=fun, nodes=wrong_nodes, fun_evals=fun_evals)
     with pytest.raises(ValueError):
         bq_no_policy.integrate(fun=None, nodes=wrong_nodes, fun_evals=fun_evals)
+
+
+@pytest.mark.parametrize("rng", [np.random.default_rng(42), 42])
+def test_integrate_runs_with_integer_rng(bq, data, rng):
+    # make sure integrate runs with both a rn generator and a seed.
```
JonathanWenger, line 153: Again, I think we should avoid this pattern.

JonathanWenger (changes requested):
All looks good, except for the implicit use of the global random state. I think we should always force the user to provide their own random number generator to avoid unexpected side effects or non-determinism. With the changes coming in #581 this will become mandatory anyways.

mmahsereci:
> All looks good, except for the implicit use of the global random state. I think we should always force the user to provide their own random number generator to avoid unexpected side effects or non-determinism. With the changes coming in #581 this will become mandatory anyways.

Sure, I'll remove the default. What about wrappers like `bayesquad` that are not really part of the core library? I assume the default rng there needs to go as well @JonathanWenger ? And thanks for reviewing so promptly!

JonathanWenger:
Yes, there the user should also provide their own RNG. It should still be optional and raise an error if at some point the compiled BQ method does need random numbers. Same as for the 'inegrate' method.

src/probnum/quad/solvers/_bayesian_quadrature.py
mmahsereci: done

src/probnum/quad/solvers/_bayesian_quadrature.py
mmahsereci: removed since obsolete.

tests/test_quad/test_bayesian_quadrature.py
mmahsereci: done

tests/test_quad/test_bayesian_quadrature.py
mmahsereci: removed since obsolete.

JonathanWenger (approved):

PR #745, opened 2022-11-23
Initial designs for `quad`

mmahsereci:
# In a Nutshell
This PR adds initial designs to the `quad` package. Further, `bayesquad` and related methods now expose an options dictionary for optional parameters similar to some scipy methods. This is because the api became increasingly messy when new options are added.

# Detailed Description
~~should be reviewed after #744 is merged.~~

# Related Issues
Closes #...


JonathanWenger (approved):
Very nice!

PR #748, opened 2022-12-05
Make `quad` integration measures stateless 

mmahsereci:
# In a Nutshell
This PR removes the default rng from the `quad` integration measures. Further, some long overdue shape tests for the sample method of the measures are added.

# Detailed Description
na

# Related Issues
Closes #...

### DIFF
--- a/src/probnum/quad/_utils.py
+++ b/src/probnum/quad/_utils.py
@@ -53,12 +53,11 @@ def as_domain(
     """
     if input_dim is not None and input_dim < 1:
         raise ValueError(
-            f"If given, input dimension must be positive. Current value "
-            f"is ({input_dim})."
+            f"Input dimension must be positive. Current value is ({input_dim})."
         )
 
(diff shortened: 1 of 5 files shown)


tskarvone (approved):
Looks good to me. As a user, I personally find it a bit annoying to have to provide an rng myself, but there is probably a good reason for this change.

mmahsereci:
> Looks good to me. As a user, I personally find it a bit annoying to have to provide an rng myself, but there is probably a good reason for this change.

It's mostly in preparation for #581 . See https://github.com/probabilistic-numerics/probnum/pull/744#discussion_r1031119249

mmahsereci:
The tox docs build seems to have failed when merging into `main`. I am not reverting though as the RTD build seems fine. Don't have time too look into it right away unfortunately. pinging @JonathanWenger for info.

JonathanWenger:
@mmahsereci Should be fixed now in #754 .

PR #749, opened 2022-12-06
Introducing `quad` acquisition functions

mmahsereci:
# In a Nutshell
This PR introduces acquisition functions to the `quad` package. Since there are no gradients yet, the corresponding policy samples random candidates and returns the one with the highest acquisition value. It will be straightforward to add a policy that uses a gradient-based optimizer once gradients are available. 

# Detailed Description
- The `BQState` got augmented with the cholesky of the gram matrix.
- The PR introduces a base class for acquisition functions as well as uncertainty sampling specifically for now.
- The PR also adds some tests for default values of the options dictionary.

~~Should be reviewed after #748 is merged as it branches off of it.~~

# Related Issues
Closes #...


src/probnum/quad/solvers/policies/_max_aquisition_policy.py
```diff
@@ -0,0 +1,71 @@
...
+class RandomMaxAcquisitionPolicy(Policy):
+    """Policy that maximizes an acquisition function by sampling random candidate nodes.
+
+    The candidate nodes are random draws from the integration measure. The node with the
+    largest acquisition value is chosen.
+
+    Parameters
+    ----------
+    batch_size
+        Size of batch of nodes when calling the policy once (must be equal to 1).
```
tskarvone, line 27: Just to confirm I'm thinking this correctly: If we returned `n = batch_size > 1` nodes by taking from the random sample those `n` nodes at which the acquisition function takes its `n` largest values, we would likely end up with a bunch of nodes very close to one another, which is not very useful.

src/probnum/quad/solvers/acquisition_functions/_predictive_variance.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def __call__(
+        self,
+        x: np.ndarray,
+        bq_state: BQState,
+    ) -> Tuple[np.ndarray, Optional[np.ndarray]]:
+        predictive_variance = bq_state.kernel(x, x)
+        if bq_state.fun_evals.shape != (0,):
+            kXx = bq_state.kernel.matrix(bq_state.nodes, x)
+            bq_weights = BQStandardBeliefUpdate.gram_cho_solve(
```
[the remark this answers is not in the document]
mmahsereci: That's a good point. Changed to `regression_weights` (in case we add noisy observations at some point).

src/probnum/quad/solvers/policies/_max_aquisition_policy.py
mmahsereci: Yes, exactly :)

mmahsereci: There are ways around it e.g., by penalizing areas around points (e.g. this https://arxiv.org/pdf/1812.01553v1.pdf)

tskarvone (approved):

PR #754, opened 2022-12-15
Updated tox configuration with new keyword for externals

JonathanWenger:
# In a Nutshell
The documentation build is failing with the following output:

```bash
docs: failed with make is not allowed, use allowlist_externals to allow it
.pkg: _exit> python /opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/pyproject_api/_backend.py True setuptools.build_meta
  docs: FAIL code 1 (103.39 seconds)
  evaluation failed :( (103.66 seconds)
```

This seems to be caused by a change in the tox configuration for external tools like `make`.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,7 +1,7 @@
 # Dependencies for a full ProbNum development environment
 -e .[full]
 
-tox>=3,<4
+tox>=3.18,<4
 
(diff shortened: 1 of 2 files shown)


mmahsereci (approved):
awesome, thanks!

PR #757, opened 2022-12-20
`MaxAcquisitionPolicy` for `quad`

mmahsereci:
# In a Nutshell
This PR introduces a policy (`MaxAcquisitionPolicy`) for the `quad` package which optimizes an acquisition function with an optimizer. 

# Detailed Description
- Scipy's 'Nelder-Mead'  is used for now as optimization method since gradients are not available. This can be swapped out once gradients become available.
- policy is added to the existing tests for policies. 
- unified the use of `n_` and `num_` to denote 'number of ...' in some places.

# Related Issues
Closes #...

### DIFF
--- a/src/probnum/quad/solvers/policies/_max_aquisition_policy.py
+++ b/src/probnum/quad/solvers/policies/_max_aquisition_policy.py
@@ -88,6 +88,13 @@ class MaxAcquisitionPolicy(Policy):
     ------
     ValueError
         If ``batch_size`` is not 1, or if ``n_restarts`` is too small.
+
+    Notes
+    -----
+    The policy uses SciPy's 'Nelder-Mead' optimizer when gradients are unavailable.
+    This is the current standard setting since ``probnum`` does not provide gradients
+    yet.
+
     """
 
     def __init__(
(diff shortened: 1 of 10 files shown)


src/probnum/quad/solvers/policies/_max_aquisition_policy.py
```diff
@@ -69,3 +70,87 @@ def __call__(
...
+        domain = bq_state.measure.domain
+        bounds = list(zip(domain[0], domain[1]))
+
+        if self.acquisition_func.has_gradients:
+            # Todo (#581): Gradients should not be available yet.
+            raise NotImplementedError
+
+        f = lambda x: -self.acquisition_func(x[None, :], bq_state)[0][0]
+        initial_points = measure.sample(n_sample=self.n_restarts, rng=rng)
+        x_min, f_min, success = None, np.inf, False
```
tskarvone, line 140: Since `f` is non-positive, should we have `f_min = -np.Inf` instead (with a corresponding change on line 145)?

mmahsereci: Not sure I understand. $f(x)$ needs to be maximized but standard optimizers always minimize. Hence, in line138 the sign of $f(x)$ is flipped. The initial value of `f_min` is set to plus infinity and is replaced whenever a smaller value is found after a single run of the optimizer. Accordingly, the current best guess `x_min` is updated with the value corresponding to `f_min`. Maybe I am not seeing something? 

The code should be equivalent to running the optimizer on $-f(x)$ `initial_points` times and then returning the minimizer with the smallest function value. 

There are no assumptions on the values $f(x)$ inside this method.

tskarvone: Sorry - it seems I got confused.

tskarvone (approved):

PR #758, opened 2022-12-22
Some more acquisition functions for `quad`

mmahsereci:
# In a Nutshell
This PR introduces the mutual information acquisition function and the integral variance acquisition function to the `quad` package.

# Detailed Description
- Both acquisition functions are available via the `RandomMaxAcquisitionPolicy` or `MaxAcquisitionPolicy` (for the latter see #757 ), hence there are 4 new policies in total.
- New `predict_integrand` method in the belief updater as such a prediction is needed for more than one acquisition function.
- The configurations for `RandomMaxAcquisitionPolicy` or `MaxAcquisitionPolicy` are renamed such that they are applicable to all acquisition functions, e.g., `us_rand_n_candidates` is renamed to `n_candidates`.
- Acquisition functions are added to existing tests. 

~~Should be reviewed after #757 is merged as it branches off of it (and currently has both diffs displaying).~~

# Related Issues
Closes #...


src/probnum/quad/solvers/_bayesian_quadrature.py
```diff
@@ -215,17 +225,19 @@ def from_problem(
...
         elif policy == "vdc":
             policy = VanDerCorputPolicy(batch_size, measure)
-        elif policy == "us_rand":
+        # all random max acquisition policies
+        elif policy in ["us_rand", "mi_rand", "ivr_rand"]:
             policy = RandomMaxAcquisitionPolicy(
                 batch_size=1,
-                acquisition_func=WeightedPredictiveVariance(),
-                n_candidates=us_rand_n_candidates,
+                acquisition_func=acquisition_dict[policy[:-5]],
```
tskarvone, line 232: This seems somewhat dangerous to me (and it took me a while to understand what is going on here).

src/probnum/quad/solvers/acquisition_functions/_integral_variance_reduction.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class IntegralVarianceReduction(AcquisitionFunction):
+    r"""The normalized reduction of the integral variance.
+
+    The acquisition function is
+
+    .. math::
+        a(x) &= \mathfrak{v}^{-1}(\mathfrak{v} - \mathfrak{v}(x))\\
+             &= \frac{(\int k(x', x)p(x')\mathrm{d}x')^2}{\mathfrak{v} v(x)}\\
```
tskarvone, line 24: I feel that `k(x', x)` here is not good notation for the posterior kernel as this is very easily confused with the prior kernel if one only reads the equation.

mmahsereci: Sure, we can change it. Any suggestion? Perhaps $k_{post}$, $k_\mathcal{D}$ or $k_N$?

src/probnum/quad/solvers/_bayesian_quadrature.py
mmahsereci: I added  a comment such that the string slice is a bit more clear. 

The added assert makes sure that the if-statement is implemented correctly in case someone ever adds a new random policy (exception raise is not needed as this is not expected to be not true). 

Is it more readable like this?

src/probnum/quad/solvers/acquisition_functions/_integral_variance_reduction.py
tskarvone: All of those appear good to me. I've also seen \bar{k} somewhere.

src/probnum/quad/solvers/_bayesian_quadrature.py
tskarvone: Thanks, that is hopefully more readable.

tskarvone (approved):

PR #762, opened 2023-01-06
CI status badge now displays correct status.

JonathanWenger:
# In a Nutshell
Due to a breaking change (https://github.com/badges/shields/issues/8671) in the behavior of status badges, the CI status was not displayed anymore. This PR corrects the link according to the recommendations in https://github.com/badges/shields/issues/8671.


marvinpfoertner (approved):
🙏🏻

PR #763, opened 2023-01-06
Fix bug in `asrandvar` introduced by breaking changes in ScIPy 1.10

JonathanWenger:
# In a Nutshell
SciPy 1.10 introduced a `Covariance` object (see https://scipy.github.io/devdocs/release.1.10.0.html#scipy-stats-improvements), which is being used for `multivariate_normal` random variables. This caused a bug in the `asrandvar` function which is handled in this PR.


marvinpfoertner (approved):
🙏🏻

PR #764, opened 2023-01-06
Fix matmul implementation of Zero linop

### LINKED ISSUE #761: Zero linop matmul bug
**Describe the bug**
`pn.linops.Zero` does not implement the `_matmul` semantic of a `LinearOperator` correctly. Currently, it is implemented via
`matmul = lambda x: np.zeros(x.shape, np.result_type(x, self.dtype))`

**To Reproduce**
```python
z = pn.linops.Zero((2, 1))
x = np.array([[42.]])
z @ x
```
will throw an AssertionError inside of `__matmul__` of `LinearOperator`, because the result has shape (1, 1).

**Expected behavior**
The result should have shape (2, 1). I think the correct implementation should take the shape of x and replace the value at index -2 with self.shape[0].

**System:**
 - OS: Arch Linux
 - Python 3.10.6, Probnum: latest commit

**Additional context**
If this is actually a bug and I'm not misunderstanding something, I can make a quick PR to get this resolved.

timweiland:
# In a Nutshell
matmul result now has the correct shape

# Detailed Description
The Zero linear operator had a bug where in the matmul implementation, it simply returned a zero vector of the same shape of the input. But this does not always match the shape we expect from the matmul output.

# Related Issues
Closes #761

### DIFF
--- a/tests/test_linops/test_linops_cases/linear_operator_cases.py
+++ b/tests/test_linops/test_linops_cases/linear_operator_cases.py
@@ -69,7 +69,6 @@ def case_identity(n: int) -> Tuple[pn.linops.LinearOperator, np.ndarray]:
     return pn.linops.Identity(shape=n), np.eye(n)
 
 
-@pytest_cases.case(tags=("symmetric"))
 @pytest.mark.parametrize("shape", [(3, 3), (3, 4), (6, 5)])
 def case_zero(shape: tuple) -> Tuple[pn.linops.LinearOperator, np.ndarray]:
     return pn.linops.Zero(shape=shape), np.zeros(shape)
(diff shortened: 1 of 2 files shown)


marvinpfoertner (approved):
🙏🏻

tests/test_linops/test_linops_cases/linear_operator_cases.py
```diff
@@ -69,6 +69,12 @@ def case_identity(n: int) -> Tuple[pn.linops.LinearOperator, np.ndarray]:
     return pn.linops.Identity(shape=n), np.eye(n)
 
 
+@pytest_cases.case(tags=("symmetric"))
```
marvinpfoertner, line 72: Here, the operator is not symmetric in all cases

marvinpfoertner (approved):

tests/test_linops/test_linops_cases/linear_operator_cases.py
timweiland: Oops, sorry! I removed that tag just now

PR #765, opened 2023-01-07
Make Matrix transpose keep some properties

timweiland:
# In a Nutshell
Transposing a Matrix now maintains symmetry and positive definiteness - if the original Matrix had them. It also flips triangular properties.

# Detailed Description
- Extend previous implementation of _transpose for Matrix
- If original Matrix is lower/upper triangular, transposed Matrix is upper/lower triangular
- If original Matrix is symmetric/positive definite, transposed Matrix is symmetric/positive definite

I am not sure if this is the right way to do it. I know that _TransposedLinearOperator already keeps these properties, but this only comes into play when the user doesn't provide a custom _transpose implementation. I think it might make sense to - more generally - transfer these properties in LinearOperator's transpose before returning the transposed linop object. But I was a bit hesitant to do that since it might interfere with this rule that you're not allowed to set LinearOperator properties that already have a value.

### DIFF
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -766,14 +766,18 @@ class LinearOperator(abc.ABC):
     @property
     def T(self) -> "LinearOperator":
         try:
-            return self._transpose()
+            transposed = self._transpose()
         except NotImplementedError:
-            pass
+            # This does not need caching, since the `TransposeLinearOperator` only accesses
+            # quantities (particularly `todense`), which are cached inside the original
+            # `LinearOperator`.
+            transposed = TransposedLinearOperator(self)
 
-        # This does not need caching, since the `TransposeLinearOperator` only accesses
-        # quantities (particularly `todense`), which are cached inside the original
-        # `LinearOperator`.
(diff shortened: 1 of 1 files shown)


marvinpfoertner (commented):
I agree with your assessment that this should be done in `LinearOperator.transpose`.
And I don't think that this interferes with the "not changing properties that are already set"-policy.
There are two cases:
1. the corresponding attribute of `self._transpose` (say `is_lower_triangular`) is not already set. In this case, you can just set it without causing any problems.
2. the corresponding attribute of `self._transpose` is already set to the value that you want to set it to.
The case where you want to set the attribute to a different value can't really happen, since this would mean that there is an error in the code.

IMHO we can just take your implementation from `Matrix._transpose` and move it to `LinearOperator.transpose`.

Thoughts @timweiland @JonathanWenger?

timweiland:
Good point :+1: 
Just pushed a new commit which moves this logic to `LinearOperator.T` (which gets called by `LinearOperator.transpose`).

marvinpfoertner (approved):

JonathanWenger (approved):

PR #771, opened 2023-01-30
Efficient implementation of arbitrary half-integer Matérn Kernels

marvinpfoertner:
# In a Nutshell
This PR updates the implementation of the Matérn kernel to efficiently support half-integer Matérn kernels of arbitrary order.
It also enables the use of per-dimension lengthscales for the `Matern` and `ExpQuad` kernels.

# Detailed Description
- efficient implementation of arbitrary half-integer Matern kernels
- per-dimension lengthscales for `Matern` and `ExpQuad`
- `input_size` property for all `Kernel`s

# Related Issues
None


src/probnum/randprocs/kernels/_matern.py
```diff
@@ -65,53 +90,135 @@ class Matern(Kernel, IsotropicMixin):
...
+
+    @property
+    def nu(self) -> ScalarType:
+        r"""Smoothness parameter :math:`\nu`."""
+        return self._nu
+
+    @functools.cached_property
+    def p(self) -> Optional[int]:
+        r"""Order :math:`p` of the polynomial part of the kernel if :math:`\nu` is a
+        half integer :math:`\nu = p + \frac{1}{2}`. Otherwise this is set to `None`."""
```
JonathanWenger, line 130:
```suggestion
        r"""Order :math:`p` of the polynomial part of the kernel, if :math:`\nu` is a
        half integer, i.e. :math:`\nu = p + \frac{1}{2}`. In this case :math:`p` corresponds
        to the differentiability of the sample paths of a GP with the given kernel.
        Otherwise this is set to `None`."""
```

JonathanWenger (approved):
Very nicely done!

PR #772, opened 2023-02-05
`LinearOperator.solve` method

marvinpfoertner:
# In a Nutshell
This PR adds the method `solve` to the `LinearOperator` interface.

# Detailed Description
- `LinearOperator.solve`
- `_InverseLinearOperator` uses `LinearOperator.solve` as `_matmul`
- remove redundant `rmatmul` argument in `LambdaLinearOperator`

# Related Issues
None

### DIFF
--- a/src/probnum/linops/_arithmetic_fallbacks.py
+++ b/src/probnum/linops/_arithmetic_fallbacks.py
@@ -36,7 +36,6 @@ class ScaledLinearOperator(LambdaLinearOperator):
             self._linop.shape,
             dtype=dtype,
             matmul=lambda x: self._scalar * (self._linop @ x),
-            rmatmul=lambda x: self._scalar * (x @ self._linop),
             todense=lambda: self._scalar * self._linop.todense(cache=False),
             transpose=lambda: self._scalar * self._linop.T,
             inverse=self._inv,
@@ -53,6 +52,9 @@ class ScaledLinearOperator(LambdaLinearOperator):
                 self.is_positive_definite = True
             else:
                 self.is_positive_definite = False
+
+    def _solve(self, B: np.ndarray) -> np.ndarray:
+        return self.inv() @ B
 
     def _inv(self) -> "ScaledLinearOperator":
         if self._scalar == 0:
@@ -98,9 +100,6 @@ class SumLinearOperator(LambdaLinearOperator):
             ),
             matmul=lambda x: functools.reduce(
                 operator.add, (summand @ x for summand in self._summands)
-            ),
-            rmatmul=lambda x: functools.reduce(
-                operator.add, (x @ summand for summand in self._summands)
             ),
             todense=lambda: functools.reduce(
                 operator.add,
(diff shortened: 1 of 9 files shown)


src/probnum/linops/_arithmetic_fallbacks.py
```diff
@@ -227,6 +223,9 @@ def __repr__(self):
             res += f"\t{s}, \n"
         return res + "]"
 
+    def _solve(self, B: np.ndarray) -> np.ndarray:
+        return functools.reduce(lambda vec, op: op @ vec, self._factors, B)
```
JonathanWenger, line 227: I may be misunderstanding, but doesn't this line compute $(B F_1) F_2$, where the linear operator is given by $L = F_1 F_2$? Shouldn't it rather be something like 
```suggestion
        return functools.reduce(lambda rhs, op: op.solve(rhs), self._factors, B)
```
If yes, also add a test for the product linear operator.

src/probnum/linops/_linear_operator.py
```diff
@@ -172,6 +174,97 @@ def __call__(self, x: np.ndarray, axis: Optional[int] = None) -> np.ndarray:
         else:
             raise ValueError("The operand must be at least one dimensional.")
 
+    def solve(self, b: ArrayLike) -> np.ndarray:
+        """Solves linear systems :code:`A @ x = b`, where :code:`b` is either a vector
+        or a (stack of) matrices.
+
+        This method behaves exactly like :code:`A.inv() @ b`.
```
JonathanWenger, line 181: Do we want to guarantee this? This is generally not the case for numpy arrays and will also not be the case anymore once we allow the user to choose a linear solver. I suggest not promising this in the docstring.
```suggestion
```

JonathanWenger (changes requested):
Mostly looks good. Only the implementation of the solve method for the `ProductLinearOperator` needs a second look I think.

src/probnum/linops/_arithmetic_fallbacks.py
marvinpfoertner: Oops, that was kind of a last minute addition. I wonder why codecov did not catch the missing coverage?

src/probnum/linops/_linear_operator.py
marvinpfoertner: I changed the docstring to communicate what I actually meant with this line in the first place.

JonathanWenger (approved):
👍👍

marvinpfoertner:
GitHub Actions seems to have network issues when communicating with its apt mirror. I already restarted the `run_notebooks` action multiple times in this PR and in #776.
If it's alright with you, I will bypass branch protections and merge anyway. The same also applies to #776.

JonathanWenger:
> GitHub Actions seems to have network issues when communicating with its apt mirror. I already restarted the `run_notebooks` action multiple times in this PR and in #776. If it's alright with you, I will bypass branch protections and merge anyway. The same also applies to #776.

Others are reporting similar failures. See e.g. https://github.com/actions/runner-images/issues/675

PR #773, opened 2023-02-05
Fallback implementations of `LinearOperator` methods

marvinpfoertner:
# In a Nutshell
The fallback implementations of methods in `LinearOperator` are now implemented in `_{method}` instead of in `{method}` after `_{method}` raised a `NotImplementedError`.

# Detailed Description
See above

# Related Issues
None


JonathanWenger (approved):
Nice!

PR #776, opened 2023-02-06
`{Identity,Symmetric}Kronecker` and `Symmetrize` implementation via subclassing

marvinpfoertner:
# In a Nutshell
This PR refactors the implementation of the linear operators in`_kronecker.py` to subclass directly from `LinearOperator` instead of from `LambdaLinearOperator`. This reduces complexity by a fair bit.

# Detailed Description
See above

# Related Issues
None


JonathanWenger (approved):
LGTM!

PR #781, opened 2023-02-15
Revert "Update ipython requirement from <8.10.0 to <8.11.0"

JonathanWenger:
Reverts probabilistic-numerics/probnum#780


JonathanWenger:
Not reverting this commit after all, since the notebook failures seem to come from a dependency that is not fixed, not `ipython` necessarily.

PR #785, opened 2023-02-16
Split `Kernel.input_shape` into `Kernel.input_shape_{0,1}`

marvinpfoertner:
# In a Nutshell
This PR splits the `input_shape` of a `Kernel` into separate input shapes for the two arguments. With this change, we can model cross-covariance functions of random processes with different domains.

# Detailed Description
- split `Kernel.input_shape` into `Kernel.input_shape_{0,1}`
- retain `Kernel.input_shape` as a convenience argument if `input_shape_0 == input_shape_1`
- the same is done for `input_{ndim_size}`

# Related Issues
None

### DIFF
--- a/src/probnum/randprocs/kernels/_kernel.py
+++ b/src/probnum/randprocs/kernels/_kernel.py
@@ -42,11 +42,19 @@ class Kernel(abc.ABC):
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


JonathanWenger (approved):
👍

PR #789, opened 2023-02-17
Rename `Kernel` to `CovarianceFunction`

marvinpfoertner:
# In a Nutshell
ProbNum's `Kernel` class does not actually represent kernels in the sense of a symmetric, positive-definite kernel (which is the canonical notion of a kernel in machine learning).
Rather, it implements (cross-)covariance functions of random processes, the important difference being that cross-covariance functions are in general not positive-definite and not even symmetric.
Hence, this PR renames the `Kernel` class to `CovarianceFunction`.

# Detailed Description
- renamed `Kernel` to `CovarianceFunction`
- renamed the module `randprocs.kernels` to `randprocs.covfuncs`
- added a deprecated dummy module `randprocs.kernels` and an alias `randprocs.kernels.Kernel` for backwards-compatibility
- renamings in tests and benchmarks for `CovarianceFunction`s
- update the documentation with the new naming scheme

# Related Issues
None


JonathanWenger (approved):
~~Consider the name `covfunctions` for consistency with `probnum.functions`.~~
EDIT: this would not be consistent with the convention of shortening all module names with two or more words.

Otherwise all looks good to me.

PR #793, opened 2023-02-23
Add KeOps support to CovarianceFunction

timweiland:
# In a Nutshell
Enable efficient matrix-free implementations of covariance linear operators via KeOps

# Detailed Description
The overall idea is inspired by @JonathanWenger's KeOps integration in IterGP.
- Add new method _keops_lazy_tensor that can be implemented in a subclass to return the LazyTensor equivalent of _evaluate_matrix
- Add new class CovarianceLinearOperator that represents the pairwise covariances and uses KeOps by default (if possible)
- Add KeOps implementations for Matern and ExpQuad

One problem I came across is that by making KeOps optional (implemented via import guards), we cannot use KeOps typing in function signatures because otherwise everything breaks for users who do not have KeOps installed. Any ideas on this?

EDIT: And I guess the CI tests are also problematic because they don't have KeOps installed (?), so the KeOps implementations will not be tested.

### DIFF
--- a/src/probnum/randprocs/covfuncs/_covariance_function.py
+++ b/src/probnum/randprocs/covfuncs/_covariance_function.py
@@ -11,7 +11,7 @@ import numpy as np
 
 try:
     from pykeops.numpy import LazyTensor
-except ImportError:
+except ImportError:  # pragma: no cover
     pass
 
 from probnum import linops, utils as _pn_utils
--- a/src/probnum/randprocs/covfuncs/_covariance_function.py
+++ b/src/probnum/randprocs/covfuncs/_covariance_function.py
@@ -10,7 +10,7 @@ from typing import Optional, Union
 import numpy as np
 
 try:
-    from pykeops.numpy import LazyTensor, Pm, Vi, Vj
+    from pykeops.numpy import LazyTensor
 except ImportError:
     pass
 
@@ -416,8 +416,6 @@ class CovarianceFunction(abc.ABC):
         self,
         x0: ArrayLike,
         x1: Optional[ArrayLike] = None,
-        *,
-        use_keops=True,
     ) -> linops.LinearOperator:
         r""":class:`~probnum.linops.LinearOperator` representing the pairwise
         covariances of evaluations of :math:`f_0` and :math:`f_1` at the given input
@@ -451,9 +449,6 @@ class CovarianceFunction(abc.ABC):
             Can also be set to :data:`None`, in which case the function will behave as
             if ``x1 == x0`` (potentially using a more efficient implementation for this
             particular case).
-        use_keops
-            Optional keyword flag to choose whether to use a KeOps-based implementation
-            or the standard implementation.
 
         Returns
         -------
@@ -482,7 +477,7 @@ class CovarianceFunction(abc.ABC):
         if x1 is not None:
             x1 = self._preprocess_linop_input(x1, argnum=1)
 
-        k_linop_x0_x1 = self._evaluate_linop(x0, x1, use_keops)
+        k_linop_x0_x1 = self._evaluate_linop(x0, x1)
 
         assert isinstance(k_linop_x0_x1, linops.LinearOperator)
         assert k_linop_x0_x1.shape == (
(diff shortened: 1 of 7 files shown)


JonathanWenger:
You should be able to add typehints using [postponed evaluation](https://peps.python.org/pep-0563/), I believe. 

```python
def some_fn() -> "pykeops.numpy.LazyTensor":
    pass
```

If the documentation does not properly parse the type hints, we might have to do a `probnum[full]` install for the docs (see below).

To test the KeOps functionality in the CI, try to add KeOps as an optional dependency in the setup.py, that is then added to the `probnum[full]` installation, which the CI uses:

```python
extras_require["matrixfree"] = [
    "pykeops>=2.1.1<3.0"
]
...
extras_require["full"] = (
    extras_require["jax"] + extras_require["zoo"] + extras_require["pls_calib"] + extras_require["matrixfree"]
)
```

src/probnum/randprocs/covfuncs/_covariance_function.py
```diff
@@ -438,6 +451,9 @@ def linop(
             Can also be set to :data:`None`, in which case the function will behave as
             if ``x1 == x0`` (potentially using a more efficient implementation for this
             particular case).
+        use_keops
+            Optional keyword flag to choose whether to use a KeOps-based implementation
+            or the standard implementation.
```
marvinpfoertner, line 456: Personally, I would remove this argument and just stick to the default behavior outlined above.
If we really need to force a standard implementation at some point, then a setting in `probnum.config` allows for more fine-grained control.
But I can't really think of a situation in which one would want to use a NumPy-based implementation if KeOps is available to be honest.
Thoughts @timweiland @JonathanWenger?

src/probnum/randprocs/covfuncs/_covariance_function.py
```diff
@@ -9,9 +9,16 @@
 
 import numpy as np
 
+try:
+    from pykeops.numpy import LazyTensor, Pm, Vi, Vj
+except ImportError:
+    pass
+
```
marvinpfoertner, line 16: It might be a good idea to move these imports into `_keops_lazy_tensor`, not sure though

marvinpfoertner (commented):
Looks very good! Thanks for taking care of this! 🙏🏻 
Let's discuss the `use_keops` argument below.

marvinpfoertner:
> To test the KeOps functionality in the CI, try to add KeOps as an optional dependency in the setup.py, that is then added to the `probnum[full]` installation, which the CI uses:
> 
> ```python
> extras_require["matrixfree"] = [
>     "pykeops>=2.1.1<3.0"
> ]

@JonathanWenger I agree this is the way to go, but we might want to choose something more specific than `matrixfree` here. How about `covfuncs-matrixfree`?

src/probnum/randprocs/covfuncs/_covariance_function.py
JonathanWenger: Agreed. One could check the `matrix_free` config option here. But currently the default is set to `False` which is not the behavior we would want here.

JonathanWenger:
> > To test the KeOps functionality in the CI, try to add KeOps as an optional dependency in the setup.py, that is then added to the `probnum[full]` installation, which the CI uses:
> > ```python
> > extras_require["matrixfree"] = [
> >     "pykeops>=2.1.1<3.0"
> > ]
> > ```
> 
> @JonathanWenger I agree this is the way to go, but we might want to choose something more specific than `matrixfree` here. How about `covfuncs-matrixfree`?

I'm fine with specific but maybe just `probnum[keops]` then? `covfuncs-matrixfree` is a handful and I expect we will package the available dependencies differently (and more user-focused) anyway.

src/probnum/randprocs/covfuncs/_covariance_function.py
timweiland: > But I can't really think of a situation in which one would want to use a NumPy-based implementation if KeOps is available to be honest.

I agree :+1:
 Just wasn't sure about it, and it made testing a little easier.

timweiland:
Thanks to both of you for the quick reviews and the helpful suggestions! I just implemented them.

I don't feel great about the way the KeOps test works now. I spent quite some time looking up a way to make the CovarianceLinearOperator think that KeOps isn't installed even if it is (to test the standard implementations), but couldn't find a way to make it work with pytest. I tried mocking the import function and setting `sys.modules['pykeops'] = None`, but both of those approaches don't work because probnum (and thus also KeOps) is already imported by the time the `CovarianceLinearOperator`s are created.

My hotfix was to just manually set `linop._use_keops = False`, but that doesn't feel right...

marvinpfoertner (approved):
🙏🏻

marvinpfoertner:
> My hotfix was to just manually set `linop._use_keops = False`, but that doesn't feel right...

Looks good to me for now, we can always change this later if we find a better solution :)

PR #794, opened 2023-03-03
Add `BlockDiagonalMatrix`

timweiland:
# In a Nutshell
Add a custom linear operator `BlockDiagonalMatrix` for block diagonal structures.

# Detailed Description
- Add `BlockDiagonalMatrix` along with custom implementations of `det`, `trace`, `eigvals`, etc...
- Add two test cases for it
- Fix bug in eigvals test. Previously this checked the arrays returned by `linop` and `matrix` for equality. But the eigenvalues may have a different order. For the comparison to be valid, we need to sort the arrays before comparing them.

### DIFF
--- a/src/probnum/linops/_block.py
+++ b/src/probnum/linops/_block.py
@@ -1,0 +1,133 @@
+"""Operators that utilize block structure."""
+from __future__ import annotations
+
+import functools
+
+import numpy as np
+from scipy.linalg import block_diag
+
+from probnum.typing import DTypeLike, LinearOperatorLike
+
+from . import _linear_operator, _utils
+
+
+class BlockDiagonalMatrix(_linear_operator.LinearOperator):
+    r"""Forms a block diagonal matrix from the input linear operators.
+
+    Given linear operators A, B, ..., Z, this represents the block linear
+    operator
+    .. math::
+    \\begin{bmatrix}
+    A & & & & \\\\
+    & B & & & \\\\
+    & & \\ddots & & \\\\
+    & & & \\ddots & \\\\
+    & & & & Z
+    \\end{bmatrix}
+
+    Parameters
+    ----------
+    *blocks :
+        The (ordered!) input linear operators.
+    """
+
+    def __init__(self, *blocks: LinearOperatorLike):
+        blocks = tuple(_utils.aslinop(block) for block in blocks)
+        assert len(blocks) > 0
+
+        dtype = functools.reduce(np.promote_types, (block.dtype for block in blocks))
+        shape_0 = sum(block.shape[0] for block in blocks)
+        shape_1 = sum(block.shape[1] for block in blocks)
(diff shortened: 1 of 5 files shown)


src/probnum/linops/_block.py
```diff
@@ -0,0 +1,134 @@
...
+        self.split_indices = np.array(
+            tuple(block.shape[1] for block in blocks)
+        ).cumsum()[:-1]
+
+        super().__init__((shape_0, shape_1), dtype)
+
+        self.is_symmetric = all(block.is_symmetric for block in blocks)
+        self.is_positive_definite = all(block.is_positive_definite for block in blocks)
+        self.is_upper_triangular = all(block.is_upper_triangular for block in blocks)
+        self.is_lower_triangular = all(block.is_lower_triangular for block in blocks)
```
marvinpfoertner, line 51: These assignments should adhere to the ternary logic used in the matrix properties.

For instance, if all the blocks are symmetric, then `self.is_symmetric` should be set to `True` and if any of the blocks is known non-symmetric (`block.is_symmetric is False`), then `self.is_symmetric` should be set to `False`. But if none of the blocks are known non-symmetric and there are some blocks for which `block.is_symmetric is None`, then `self.is_symmetric` should be set to `None`, since we don't know anything about its symmetry properties in this case.

The same holds for the other properties.

src/probnum/linops/_block.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+    def _transpose(self) -> BlockDiagonalMatrix:
+        return BlockDiagonalMatrix(*[block.T for block in self.blocks])
+
+    def _solve(self, B: np.ndarray) -> np.ndarray:
+        if all(block.is_square for block in self.blocks):
+            return np.concatenate(
+                tuple(
+                    cur_block.inv() @ cur_x
```
marvinpfoertner, line 79:
```suggestion
                    cur_block.solve(cur_x)
```

marvinpfoertner (approved):
Nice! 🙏🏻 

Since block-diagonal matrices with non-square blocks are super weird, I would be in favor of requiring that the blocks are square in the constructor.
But I don't have a strong opinion about this, so feel free to decide what to do with this :)
Thoughts @JonathanWenger?

src/probnum/linops/_block.py
timweiland: Good catch, thanks!

timweiland:
Thanks for the reviews!
Upon further thought and after discussion with @marvinpfoertner, I decided not to require all blocks to be square.
For example, `BlockDiagonalMatrix` is useful for implementing independent multi-output kernels. When we consider k(X, X), everything is fine and all blocks are square, but k(x, X) (i.e. the cross-covariances between the input and the training data) will have non-square blocks in general. So there is a good use case for it, I think.

PR #800, opened 2023-03-08
Bugfix in `CovarianceLinearOperator` transpose

timweiland:
# In a Nutshell
Bugfix in transpose of `CovarianceLinearOperator`

# Detailed Description
The previous `_transpose` implementation only worked for the case that both inputs are equal. The new implementation works for the general case where x0 and x1 may not be equal.

The new implementation now requires passing a callback `keops_lazy_tensor_fn` (as opposed to directly passing a `LazyTensor`) which the `CovarianceLinearOperator` can then pass on in `_transpose`.

### DIFF
--- a/src/probnum/randprocs/covfuncs/_covariance_linear_operator.py
+++ b/src/probnum/randprocs/covfuncs/_covariance_linear_operator.py
@@ -1,11 +1,12 @@
 """LinearOperator that represents pairwise covariances of evaluations."""
 
-from typing import Callable, Optional, Tuple
+from typing import Callable, Optional
 import warnings
 
 import numpy as np
 
 from probnum import linops
+from probnum.typing import ShapeType
 
 _USE_KEOPS = True
 try:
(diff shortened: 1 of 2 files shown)


src/probnum/randprocs/covfuncs/_covariance_linear_operator.py
```diff
@@ -81,5 +84,9 @@ def _matmul(self, x: np.ndarray) -> np.ndarray:
 
     def _transpose(self) -> linops.LinearOperator:
         return CovarianceLinearOperator(
-            self._x0, self._x1, self._evaluate_dense_matrix, self.keops_lazy_tensor
+            self._x1,
+            self._x0,
+            (self.shape[1], self.shape[0]),
+            self._evaluate_dense_matrix,
+            self._keops_lazy_tensor_fn,
         )
```
marvinpfoertner, line 92: If I'm not mistaken, then this implementation is only correct if `self` is a actually a covariance function. If `self` is a cross-covariance function, e.g. $\mathcal{L} k$ for some linear operator $\mathcal{L}$, then it might not even be well-defined to swap the inputs like this.

src/probnum/randprocs/covfuncs/_covariance_linear_operator.py
```diff
@@ -50,21 +50,24 @@ def __init__(
         self,
         x0: np.ndarray,
         x1: Optional[np.ndarray],
-        shape,
-        evaluate_dense_matrix: callable,
-        keops_lazy_tensor: Optional["LazyTensor"] = None,
+        shape: Tuple[int],
```
marvinpfoertner, line 53:
```suggestion
        shape: ShapeType,
```
where `ShapeType` is defined in `probnum.typing`.

src/probnum/randprocs/covfuncs/_covariance_linear_operator.py
timweiland: Hmm, you're right. Unfortunately, as far as I know, KeOps doesn't allow you to simply get the transpose of a `LazyTensor`. I guess the correct way to do it would then be to add something like a `trans` flag which, if set to `True`, calls `(x.T @ self.keops_lazy_tensor).T` instead of `self.keops_lazy_tensor @ x` in `_matmul`? What do you think?

JonathanWenger: Regarding the transpose of a 'keops.LazyTensor': Does 'mylazytensor.t()' not do what you want? See also: http://www.kernel-operations.io/keops/python/api/common/GenericLazyTensor.html#pykeops.common.lazy_tensor.GenericLazyTensor.t

timweiland: Well, this is awkward. You're absolutely right. I don't know why I didn't find that...
Thanks! Then I'll use this.

PR #801, opened 2023-03-08
Use triangular solves in default `_solve`

timweiland:
# In a Nutshell
Make use of triangular property in `_solve`

# Detailed Description
- Make the default `_solve` implementation of `LinearOperator` use triangular solves if the linear operator is known to be lower/upper triangular
- Add test cases for triangular matrices


marvinpfoertner (approved):
🙏🏻

PR #807, opened 2023-03-21
Python 3.11

marvinpfoertner:
# In a Nutshell
This PR bumps the Python version to 3.11.

# Detailed Description
- run unit tests for Python 3.11

# Related Issues
None


marvinpfoertner:
@JonathanWenger The tests fail because GPy is not compatible with Python 3.11 (see https://github.com/SheffieldML/GPy/issues/998)

JonathanWenger:
> @JonathanWenger The tests fail because GPy is not compatible with Python 3.11 (see [SheffieldML/GPy#998](https://github.com/SheffieldML/GPy/issues/998)). How should we address this?

As discussed offline, we'll remove the GPy dependency as part of this PR. Since removing it was planned anyway.

PR #818, opened 2023-05-04
Override `linop` instead of `_evaluate_linop`

timweiland:
# In a Nutshell
Override `linop` instead of `_evaluate_linop` in arithmetic fallback covariance functions

# Detailed Description
In arithmetic fallback covariance functions, override `linop` instead of `_evaluate_linop`, because the latter has the problem that the input preprocessing destroys attributes of np.array subclasses. Keeping the option of np.array subclassing can be very useful for certain linop implementations, e.g. tensor product covariance functions, where you might want to subclass np.array to represent grid structure in the inputs.


JonathanWenger (approved):
LGTM, but maybe @marvinpfoertner should take a quick look as well. I think he may have written this oroginally.

src/probnum/randprocs/covfuncs/_arithmetic_fallbacks.py
```diff
@@ -55,8 +55,8 @@ def __init__(self, covfunc: CovarianceFunction, scalar: ScalarLike):
     def _evaluate(self, x0: np.ndarray, x1: Optional[np.ndarray] = None) -> np.ndarray:
         return self._scalar * self._covfunc(x0, x1)
 
-    def _evaluate_linop(
-        self, x0: np.ndarray, x1: Optional[np.ndarray]
+    def linop(
+        self, x0: utils.ArrayLike, x1: Optional[utils.ArrayLike] = None
```
marvinpfoertner, line 59: Could you add `ArrayLike` to `from probnum.typing import [...]` above?

```suggestion
        self, x0: ArrayLike, x1: Optional[ArrayLike] = None
```

src/probnum/randprocs/covfuncs/_arithmetic_fallbacks.py
```diff
@@ -104,8 +103,8 @@ def _evaluate(self, x0: np.ndarray, x1: Optional[np.ndarray]) -> np.ndarray:
             operator.add, (summand(x0, x1) for summand in self._summands)
         )
 
-    def _evaluate_linop(
-        self, x0: np.ndarray, x1: Optional[np.ndarray]
+    def linop(
+        self, x0: utils.ArrayLike, x1: Optional[utils.ArrayLike] = None
```
marvinpfoertner, line 107:
```suggestion
        self, x0: ArrayLike, x1: Optional[ArrayLike] = None
```

marvinpfoertner (approved):
LGTM! 🙏🏻

src/probnum/randprocs/covfuncs/_arithmetic_fallbacks.py
timweiland: Good catch!

marvinpfoertner (approved):

PR #820, opened 2023-05-08
Linop diagonal

timweiland:
# In a Nutshell
Add `LinearOperator.diagonal()`.

# Detailed Description
... which computes the diagonal of the linear operator. Default implementation uses `np.diagonal`, and subclasses may use more efficient implementations, e.g. for Kronecker linear operators.

# "Future work"
- Figure out a good way to compute the diagonal of a kronecker product when the second factor is not square
- Implement diagonal of a covariance linear operator (shouldn't be too hard since it gets both inputs as arguments, but then you would possibly have to reshape and call the covariance function so that it only computes the diagonal)

### DIFF
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -1615,7 +1615,7 @@ class Matrix(LambdaLinearOperator):
             matmul = LinearOperator.broadcast_matmat(lambda x: self.A @ x)
             todense = self.A.toarray
             trace = lambda: self.A.diagonal().sum()
-            diagonal = lambda: self.A.diagonal()
+            diagonal = self.A.diagonal
         else:
             self.A = np.asarray(A)
             self.A.setflags(write=False)
--- a/src/probnum/linops/_linear_operator.py
+++ b/src/probnum/linops/_linear_operator.py
@@ -109,6 +109,7 @@ class LinearOperator(abc.ABC):  # pylint: disable=too-many-instance-attributes
         self._det_cache = None
         self._logabsdet_cache = None
         self._trace_cache = None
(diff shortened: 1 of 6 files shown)


src/probnum/linops/_linear_operator.py
```diff
@@ -777,6 +778,22 @@ def trace(self) -> np.number:
 
         return self._trace_cache
 
+    def _diagonal(self) -> np.ndarray:
+        """Diagonal of the linear operator.
+
+        You may implement this method in a subclass.
+        """
+        return np.diagonal(self.todense(cache=False))
```
marvinpfoertner, line 786: How about we modify the default implementation of `trace` (a loop that iteratively multiplies with canonical unit vectors) to compute the diagonal instead and then we can implement `trace` as `np.sum(self.diagonal)` by default? This would mean O(n) memory complexity in both cases and we can leverage the diagonal cache in both cases.

marvinpfoertner (changes requested):

src/probnum/linops/_linear_operator.py
timweiland: It's O(n^2) memory complexity then, right? Because the default implementation of `diagonal` computes the dense representation of the linop. This might be a problem for large matrix-free linops where no memory-efficient implementation of `diagonal` is available. (To be fair, I can't think of an example right now - anytime I think of large linops I think of covariance matrices where computation of the diagonal is indeed possible with O(n) memory.)
But from an implementation POV, I agree that it would be more elegant.

Just wanted to raise that point, lmk what you think

marvinpfoertner (approved):
🙏🏻

PR #821, opened 2023-05-09
Fix readthedocs build

timweiland:
# In a Nutshell
RTD builds have been broken for the past couple of days. This is because urllib3 was released which is not compatible with the docker image we're using for the build. See https://github.com/readthedocs/readthedocs.org/issues/10290

With this PR, RTD should use a more recent docker image which hopefully fixes the problem.

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -5,6 +5,12 @@
 # RTD configuration file version
 version: 2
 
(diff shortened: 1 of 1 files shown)


JonathanWenger (approved):
Thanks for taking care of this 🙏🏼

PR #842, opened 2023-10-09
Reduce frequency of dependabot updates to monthly

JonathanWenger:
# In a Nutshell
Previously dependabot checked dependencies weekly. We depend on libraries that frequently change their version (e.g. JAX), but we don't need to upgrade the dependencies as frequently.


marvinpfoertner (approved):
🙏🏻

PR #864, opened 2025-06-10
Enable ProbNum Import with NumPy 2.0 (Tested via linpde-gp)

### LINKED ISSUE #862: Bug with numpy when importing ProbNum package
**Describe the bug**
When I run "Import probnum as pn" and it runs the __init__.py file, I get an attribute error:
 `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.
I downloaded probnum in a new/fresh conda environment, with no other packages installed prior to installing probnum (apart from the ones that are automatically installed in the process of installing the probnum package).

**To Reproduce**
Create a new conda environment with the most recent python version. Install pip, then run "pip install probnum". Open a python script and run "import probnum as pn" with no other lines of code. 

**Expected behavior**
I would expect no attribute error to appear when importing the package.

**System:**
 - OS: MacOS 14.4.1
 - Numpy Version: 2.1.0, ProbNum Version: 0.1.16, Python Version: 3.12.4, Pip Version: 24.2

bydeng01:
# In a Nutshell
This PR updates probnum to be compatible with NumPy 2.0 and above.

# Detailed Description
- Fixes AttributeError caused by the removal of np.float_ in NumPy 2.0.
- Replaces deprecated NumPy aliases with explicit types (np.float64, np.int64, etc.).
- Tested against the linpde-gp test suite with NumPy 2.0 — all tests pass (https://github.com/bydeng01/linpde-gp/tree/main/experiments).

# Related Issues
Closes #862   https://github.com/probabilistic-numerics/probnum/issues/862

### DIFF
--- a/docs/source/development/implementing_a_probnum_method.ipynb
+++ b/docs/source/development/implementing_a_probnum_method.ipynb
@@ -728,7 +728,7 @@
     "# %load -s function_evaluation quadopt_example/observation_operators\n",
     "def function_evaluation(\n",
     "    fun: Callable[[FloatLike], FloatLike], action: FloatLike\n",
-    ") -> np.float_:\n",
+    ") -> np.float64:\n",
     "    \"\"\"Observe a (noisy) function evaluation of the quadratic objective.\n",
     "\n",
     "    Parameters\n",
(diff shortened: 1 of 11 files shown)


marvinpfoertner (approved):
LGTM! Thanks for taking care of this! :)

